PR #7, opened 2019-04-18
Image generation improvements

cfergeau:
Once again, a mix of various changes in this pull request. The main goal was to be able to generate the image without any input from the user, which is mostly done after this series. The pull secret is passed through an environment variable. This also changes the name of the generated VMs  from 'test1' to 'crc', and even makes it configurable.

PR #8, opened 2019-04-24
Add generation of crc-bundle-info.json file

cfergeau:
This PR is just a single commit, but it's on top of the changes already proposed in the PR #7 
This should get clearer once #7 is merged.
The related issue ( https://github.com/code-ready/crc/issues/34 ) has some open comment in https://github.com/code-ready/crc/issues/34#issuecomment-485845805 :

What I'm generating so far:
```
{
  "type": "snc",
  "buildInfo": {
    "buildTime": "2019-04-23T14:55:32+00:00",
    "openshiftInstallerVersion": "./openshift-install v0.16.1\nbuilt from commit e3fceacc975953f56cb09931e6be015a36eb6075",                                                                                       
    "sncVersion": "git9662"
  },
  "clusterInfo": {
    "clusterName": "crc",
    "baseDomain": "tt.testing",
    "sshPrivateKeyFile": "master_privatekey",
    "kubeConfig": "kubeconfig"
  },
  "storage": {
    "diskImages": [
      {
        "name": "crc.qcow2",
        "format": "qcow2"
      }
    ]
  }
}
```
The "diskImages" bit should probably be made simpler for now, it's not clear at all to me right now how we would describe multi node setups in this file. Not sure it would be convenient at all.


crc-bundle-info.json.sample
```diff
@@ -0,0 +1,39 @@
...
+    "openshiftInstallerVersion": "./openshift-install v0.16.1\nbuilt from commit e3fceacc975953f56cb09931e6be015a36eb6075",
+    # Output of 'git describe' or 'git rev-parse' of the 'snc' script
+    # repository used when generating the bundle
+    "sncVersion": "git9662"
+  },
+  "clusterInfo": {
+    # Name of the openshift cluster stored in the bundle
+    "clusterName": "crc",
+    # Base domain name used for the openshift cluster
+    "baseDomain": "tt.testing",
```
praveenkumar: We also need the hostname which can be `${VM_PREFIX}-master-0` in our case.

cfergeau: This has been added in the latest iterations.

createdisk.sh
```diff
@@ -29,10 +29,22 @@ function create_disk_image {
     rm -fr $tarballDirectory/${VM_PREFIX}-master-0 $tarballDirectory/${VM_PREFIX}-base
 }
 
+function update_json_description {
+    cat $1/crc-bundle-info.json \
+        | ${JQ} ".clusterInfo.masterHostname = \"${VM_PREFIX}-master-0\"" \
```
praveenkumar: @cfergeau you added it here but your sample file still doesn't have a field for `masterHostname` under `clusterInfo`

createdisk.sh
```diff
@@ -29,10 +29,22 @@ function create_disk_image {
...
+        | ${JQ} '.clusterInfo.sshPrivateKeyFile = "master_privatekey"' \
```
praveenkumar: @cfergeau shouldn't it would be now `id_rsa_crc` ?

createdisk.sh
cfergeau: Ah indeed, thanks for catching this!

createdisk.sh
cfergeau: I kept the initial name, but I definitely prefer id_rsa_crc :)

PR #25, opened 2019-05-17
Misc cleanups in crc_libvirt.sh template

cfergeau:
Some cleanups I did a while ago, which I'm pushing now since crc_libvirt.sh needed a bugfix.

### DIFF
--- a/crc_libvirt.template
+++ b/crc_libvirt.template
@@ -18,15 +18,6 @@ prerequisites()
     # Enable IP forwarding
     sudo sysctl net.ipv4.ip_forward=1
 
-    # Configure libvirt to accept TCP connections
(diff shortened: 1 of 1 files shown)


crc_libvirt.template
```diff
@@ -132,61 +121,37 @@ EOF
...
     </graphics>
     <video>
-      <model type='cirrus' vram='16384' heads='1' primary='yes'/>
-      <address type='pci' domain='0x0000' bus='0x00' slot='0x02' function='0x0'/>
+      <model type='cirrus'/>
     </video>
-    <memballoon model='virtio'>
-      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x0'/>
-    </memballoon>
+    <memballoon model='virtio'/>
```
praveenkumar: I don't think we need anything from https://github.com/code-ready/snc/blob/df5451776a9414faa6395b99f2ce5c3a09e7598d/crc_libvirt.template#L136-L151 also, you can remove that part same what I did recently for libvirt driver side. 

https://github.com/code-ready/machine-driver-libvirt/blob/master/constants.go#L37-L50

PR #36, opened 2019-06-04
Disable kubelet service and pull dnsmasq image

### LINKED ISSUE #34: Disable the kubelet service before creating the disk image
We should disable the kubelet service since on Mac/Windows we are going to run the dns server before starting the cluster.

### LINKED ISSUE #35: Pull dnsmasq container image before creating the disk image
As of now we are going to use `quay.io/crcont/dnsmasq:latest` as part of internal dns resolution so it would be great to have it in as part of disk image then pulling it from the registry everytime a user start the cluster.

anjannath:
Fixes #34 
Fixes #35

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -140,6 +140,10 @@ if [ -z $random_string ]; then
 fi
 VM_PREFIX=${CRC_VM_NAME}-${random_string}
 
+# Disable kubelet service and pull dnsmasq image from quay.io/crcon/dnsmasq
+ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo systemctl disable kubelet
+ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo podman pull quay.io/crcont/dnsmasq:latest
+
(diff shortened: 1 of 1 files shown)


anjannath:
This is untested..

cfergeau:
Changes look good, though I would split them in 2 separate commits, as this would provide an opportunity to explain for each change *why* it's desirable to do this, so that we remember the rationale in a few years from now if needed.
Take this comment with a grain of salt though, as this is *not* how commits are usually handled in the project :)

createdisk.sh
```diff
@@ -140,6 +140,10 @@ if [ -z $random_string ]; then
 fi
 VM_PREFIX=${CRC_VM_NAME}-${random_string}
 
+# Disable kubelet service and pull dnsmasq image from quay.io/crcon/dnsmasq
+ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo systemctl disable kubelet
+ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo podman pull quay.io/crcont/dnsmasq:latest
+
```

PR #40, opened 2019-06-06
Enable lazy_refcounts on the qcow2 image we ship

### LINKED ISSUE #38: Workaround hyperkit qcow2 bug
This is a workaround for an old issue that the project has been having, see https://github.com/code-ready/osp4/issues/9
Some qcow2 images created by qemu-img causes a freeze in hyperkit qcow2 implementation, see https://github.com/moby/hyperkit/issues/221
Enabling lazy refcounts in the qcow2 image seems to avoid this issue, and this is a performance improvement which should not have many adverse effects (slower startup in case of improper VM shutdown), see https://lists.gnu.org/archive/html/qemu-devel/2012-06/msg03827.html

cfergeau:
This works around a bug in hyperkit qcow2 implementation on macos which
would cause the whole VM to freeze when it tries to dynamically grow the
qcow2 image
https://github.com/moby/hyperkit/issues/221

Lazy refcounts are a qcow2 performance improvement which should not have
many adverse effects (slower startup in case of improper VM shutdown),
see https://lists.gnu.org/archive/html/qemu-devel/2012-06/msg03827.html

This fixes #38.
https://github.com/code-ready/snc/issues/38

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -44,7 +44,7 @@ function create_qemu_image {
 
     # Resize the image from the default 1+15GB to 1+20GB
     # This should also take care of making the image sparse
-    ${QEMU_IMG} create -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 21G
+    ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 21G
     ${VIRT_RESIZE} --expand /dev/sda3 $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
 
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -44,7 +44,7 @@ function create_qemu_image {
 
     # Resize the image from the default 1+15GB to 1+20GB
     # This should also take care of making the image sparse
-    ${QEMU_IMG} create -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 21G
+    ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 21G
```

PR #42, opened 2019-06-12
Wait until VM shutdown gracefully.

praveenkumar:


createdisk.sh
```diff
@@ -146,7 +146,11 @@ ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo podma
 
 # Shutdown the instance
 sudo virsh shutdown ${VM_PREFIX}-master-0
-
+# Wait till instance shutdown gracefully
+until sudo virsh domstate ${VM_PREFIX}-master-0 | grep shut; do
```
cfergeau: Initially I wanted to suggest using `virsh event` rather than doing this polling, but on second thought this would be racy as the VM could stop before `virsh event` starts. `virsh domstate` can be localized though (on my system `shut off` is `fermé`) so you'll need to set either `LANG` or `LC_ALL`.

snc.sh
```diff
@@ -124,6 +124,15 @@ fi
 rm id_rsa_crc* || true
 ssh-keygen -N "" -f id_rsa_crc
 
+# Set NetworkManager DNS overlay file
+cat << EOF | sudo tee /etc/NetworkManager/dnsmasq.d/openshift.conf
+server=/crc.testing/192.168.126.1
+address=/apps-crc.testing/192.168.126.11
```
cfergeau: I'd use `CRC_VM_NAME` and `BASE_DOMAIN` rather than `crc` and `testing`

snc.sh
```diff
@@ -124,6 +124,15 @@ fi
...
+EOF
+
+# Reload the NetworkManager to make DNS overlay effective
+sudo systemctl reload NetworkManager
+
```
cfergeau: why only this specific change? This assumes there is already `/etc/NetworkManager/conf.d/openshift.conf`, and there is more pre-configuration needed (firewalld changes, libvirtd changes). I'm not saying I'm against it, but a more detailied rationale in the commit log would be useful.

snc.sh
```diff
@@ -124,6 +124,15 @@ fi
 rm id_rsa_crc* || true
 ssh-keygen -N "" -f id_rsa_crc
 
+# Set NetworkManager DNS overlay file
+cat << EOF | sudo tee /etc/NetworkManager/dnsmasq.d/openshift.conf
```
anjannath: How was it working without this configuration before?

anjannath: Ahh, its part of the pre-requisite one time configs.

snc.sh
cfergeau: An alternative to doing the config ourselves would be to only test that the prerequisite are met (in other words, try to resolve `aaa.crc.testing` and error out if this is not using the DNS server we expect, try to connect to the libvirt port with nc, ...)

praveenkumar: @cfergeau I just want to have this part not other pre-requisites as of now since I mostly use the GCP where the pre-requisites are set except this one. I will just update the commit message as of now for it.

PR #43, opened 2019-06-13
Issue 22 Stop the kubelet service and remove the pods.

praveenkumar:


createdisk.sh
```diff
@@ -144,8 +146,16 @@ fi
 VM_PREFIX=${CRC_VM_NAME}-${random_string}
 
 # Disable kubelet service and pull dnsmasq image from quay.io/crcon/dnsmasq
-ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo systemctl disable kubelet
-ssh -o StrictHostKeyChecking=no -i id_rsa_crc core@api.crc.testing -- sudo podman pull quay.io/crcont/dnsmasq:latest
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable kubelet
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont/dnsmasq:latest
+
```
cfergeau: I would move these 2 changes to the next commit I think

praveenkumar: Done :)

PR #46, opened 2019-06-18
Issue #44 Scale down the ingress router from 2 to 1.

praveenkumar:
As of now default ingress operator deployment have 2 replica for
ingress (which provides routing) since in single node cluster second
one is not going to come up and work.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -199,3 +199,6 @@ ${OC} scale --replicas=0 deployment --all -n openshift-machine-config-operator
 
 # Scale route deployment from 2 to 1
 ${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/default -n openshift-ingress-operator
+
+# Set default route for registry CRD from false to true.
+${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
--- a/snc.sh
+++ b/snc.sh
@@ -196,3 +196,6 @@ ${OC} delete pods -l 'app in (installer, pruner)' -n openshift-kube-controller-m
 # Disable the deployment/replicaset for openshift-machine-api and openshift-machine-config-operator
 ${OC} scale --replicas=0 deployment --all -n openshift-machine-api
 ${OC} scale --replicas=0 deployment --all -n openshift-machine-config-operator
+
+# Scale route deployment from 2 to 1
+${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/default -n openshift-ingress-operator


snc.sh
```diff
@@ -196,3 +196,9 @@ ${OC} delete pods -l 'app in (installer, pruner)' -n openshift-kube-controller-m
 # Disable the deployment/replicaset for openshift-machine-api and openshift-machine-config-operator
 ${OC} scale --replicas=0 deployment --all -n openshift-machine-api
 ${OC} scale --replicas=0 deployment --all -n openshift-machine-config-operator
+
+# Scale route deployment from 2 to 1
+${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/default -n openshift-ingress-operator
```
anjannath: Is this different than using `oc scale` command directly?

praveenkumar: yes, this is not deployment scale it is the cluster operator config update.

praveenkumar:
> Why isn't this set to 'true' by default? Ie are there any side-effects to doing that? 

@cfergeau this is by design from openshift side.

cfergeau:
> > Why isn't this set to 'true' by default? Ie are there any side-effects to doing that?
> 
> @cfergeau this is by design from openshift side.

Yes, but why? :) are there security implications, which is why they disable it? or are there other reasons? And why is it ok for us to go against their design decisions? I'm not saying we should not be doing it, but ideally this would be documented if you did some research on this.

praveenkumar:
> Yes, but why? :) are there security implications, which is why they disable it? or are there other reasons? And why is it ok for us to go against their design decisions? I'm not saying we should not be doing it, but ideally this would be documented if you did some research on this.

@cfergeau https://github.com/openshift/openshift-docs/pull/15370 is doc PR which doesn't list all those details which you are asking. In production side it is not good practice to enable to insecure route for registry instead create route cert and use it. In our case since it is for local development we don't really care about atm to create a new cert and use but instead just allow insecure registry. Does that makes sense?

cfergeau:
> > Yes, but why? :) are there security implications, which is why they disable it? or are there other reasons? And why is it ok for us to go against their design decisions? I'm not saying we should not be doing it, but ideally this would be documented if you did some research on this.
> 
> @cfergeau [openshift/openshift-docs#15370](https://github.com/openshift/openshift-docs/pull/15370) is doc PR which doesn't list all those details which you are asking. In production side it is not good practice to enable to insecure route for registry instead create route cert and use it. In our case since it is for local development we don't really care about atm to create a new cert and use but instead just allow insecure registry. Does that makes sense?

Ah right, since you need to disable TLS verification to access this, it's indeed not so great to have it enabled by defalt in openshift. I agree that in our case, it's not that bad. I'd document this in the commti log though ;) apart  from the misc comments on the logs, this series looks good to me.

PR #48, opened 2019-06-20
Implementation of tarball creation for hyperkit

praveenkumar:


createdisk.sh
```diff
@@ -184,6 +210,14 @@ copy_additional_files $1 $libvirtDestDir
 
 tar cJSf $libvirtDestDir.tar.xz $libvirtDestDir
 
+# HyperKit image generation
+# This must be done after the generation of libvirt image as it resuse some of
+# the content of $libvirtDestDir
```
cfergeau: Given that the libvirt and hyperkit disk images are going to be the same, we could use the same bundle for both. When used with libvirt, it will contain a few unneeded  files (the kernel and initramfs), but that will not make a significant difference in disk usage.

praveenkumar: @cfergeau it is not only the extra files but the metadata content also, I prefer to have separate tarball for separate hypervisor.

cfergeau: The extra metadata can be ignored too. Main advantage of a single tarball is less download time if you use both macos and linux.

praveenkumar: @cfergeau so you want to say have only single libvirt tarball which going to work with linux and mac with hyperkit?

cfergeau: yes, I think this is what I would have done, less tarballs to generate, to download, ... But if we are going to ship crc+bundle as a single "binary", then it's really not important whether the libvirt and hyperkit bundles are the same or are different.

praveenkumar:
Created the required directory have all the required info now.

```
$ ls -l crc_hyperkit_2019-06-20/
total 8202304
-rw-rw-r-- 1 prkumar prkumar       1420 Jun 20 10:02 crc-bundle-info.json
-rw-r--r-- 1 prkumar prkumar 8335261696 Jun 20 10:02 crc.qcow2
-r-------- 1 prkumar prkumar       1831 Jun 20 10:02 id_rsa_crc
-rw-r--r-- 1 prkumar prkumar   55997484 Jun 20 10:02 initramfs-4.18.0-80.1.2.el8_0.x86_64.img
-rw-r--r-- 1 prkumar prkumar         23 Jun 20 10:02 kubeadmin-password
-rw-r--r-- 1 prkumar prkumar       8935 Jun 20 10:02 kubeconfig
-rwxr-xr-x 1 prkumar prkumar    7868768 Jun 20 10:02 vmlinuz-4.18.0-80.1.2.el8_0.x86_64
```

createdisk.sh
```diff
@@ -156,9 +176,15 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crictl pods -q)'
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q)'
 
-# Get rhcos ostree Hash ID
+# Get the rhcos ostree Hash ID
```
cfergeau: This belongs to the previous commit, but I'd just squash these 2 commits together.

createdisk.sh
```diff
@@ -156,9 +176,15 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
...
 ostree_hash=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline | grep -oP "(?<=rhcos-).*(?=/vmlinuz)"')
 
+# Get the rhcos kernel release
+kernel_release=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'uname -r')
+
+# SCP the vmlinuz/initramfs from VM to Host in provided folder.
+${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/rhcos-${ostree_hash}/* $1
+
```
cfergeau: vmlinuz/initramfs are the only 2 files contained in `/boot/ostree/rhcos-${ostree_hash}/`? or do we need to be more specific in what we copy?

createdisk.sh
```diff
@@ -184,6 +210,14 @@ copy_additional_files $1 $libvirtDestDir
 
 tar cJSf $libvirtDestDir.tar.xz $libvirtDestDir
 
+# HyperKit image generation
+# This must be done after the generation of libvirt image as it resuse some of
```
cfergeau: `resuse` -> `reuses`

createdisk.sh
```diff
@@ -105,6 +106,25 @@ function generate_vbox_directory {
...
 
+function generate_hyperkit_directory {
+    srcDir=$1
+    destDir=$2
+    tmpDir=$3
+
+    cp $srcDir/kubeadmin-password $destDir/
+    cp $srcDir/kubeconfig $destDir/
+    cp $srcDir/id_rsa_crc $destDir/
+    cp $srcDir/${CRC_VM_NAME}.qcow2 $destDir/
```
cfergeau: maybe create a hardlink instead? should be faster and consume less disk space

crc-bundle-info.json.sample
```diff
@@ -43,6 +43,12 @@
       # Disk image used by the node, the 'storage' object will contain more
       # details about its format
       "diskImage": "crc.qcow2"
+      # kernel command line of the node
+      "kernelCmdLine": "BOOT_IMAGE=/ostree/rhcos-bf3b38268ddb2a2070dc587b361ce45c46a6f9e3606bbd3b5e15b2e6d3d47e5d/vmlinuz-4.18.0-80.1.2.el8_0.x86_64 console=tty0 console=ttyS0,115200n8 rootflags=defaults,prjquota rw root=UUID=a8fbdcb1-63ea-421e-8f6f-dac9cbbcc822 ostree=/ostree/boot.0/rhcos/bf3b38268ddb2a2070dc587b361ce45c46a6f9e3606bbd3b5e15b2e6d3d47e5d/0 coreos.oem.id=qemu ignition.platform.id=qemu"
+      # initramfs file of the node
+      "initramfs": "initramfs-4.18.0-80.1.2.el8_0.x86_64.img"
+      # vmlinuz file of the node
+      "vmlinuz": "vmlinuz-4.18.0-80.1.2.el8_0.x86_64"
```
cfergeau: I'd change "vmlinuz" to "kernel"

createdisk.sh
```diff
@@ -105,6 +106,25 @@ function generate_vbox_directory {
...
+    cp $tmpDir/vmlinuz-${kernel_release} $destDir/
+    cp $tmpDir/initramfs-${kernel_release}.img $destDir/
+
+    cat $srcDir/crc-bundle-info.json \
+        | ${JQ} ".nodes[0].diskImage = \"${CRC_VM_NAME}.qcow2\"" \
+        | ${JQ} ".storage.diskImages[0].name = \"${CRC_VM_NAME}.qcow2\"" \
+        | ${JQ} '.storage.diskImages[0].format = "qcow2"' \
+        >$destDir/crc-bundle-info.json
```
cfergeau: I believe the diskImage metadata is already present in $srcDir/crc-bundle-info.json? (and you remove this in the next commit)

createdisk.sh
praveenkumar: yes it is only 2 files which are present there.

Issue #54, opened 2019-06-26
Project/namespace deletion goes to hanging state.


praveenkumar:
On CRC side when a user create a new project/namespace and want to delete then it goes to hanging state forever. Looking on the kubernetes side and found out it is occur because of CRD's [0][1], as part of our CRC creation we scale down the  cluster wide monitoring operator[2] which serve 'v1beta1.metrics.k8s.io` as per my understanding. if we also delete the apiservice that would not break anything for the users and if a user want to try out the cluster wide monitoring then it will again created as part of CRD's. 

[0] https://github.com/kubernetes/kubernetes/issues/60807
[1] https://github.com/kubernetes/kubernetes/issues/60807#issuecomment-429754733
[2] https://github.com/code-ready/snc/blob/master/snc.sh#L184-L186

praveenkumar:
resolved by #55

PR #55, opened 2019-06-26
Issue #54 Delete the v1beta1.metrics.k8s.io apiservice

praveenkumar:
On CRC side when a user create a new project/namespace and want to delete,
it goes to hanging state forever. Looking on the kubernetes side and found out
it is occur because of CRD's [0][1], as part of our CRC creation we scale down
the cluster wide monitoring operator[2] which serve 'v1beta1.metrics.k8s.io`.

[0] https://github.com/kubernetes/kubernetes/issues/60807
[1] https://github.com/kubernetes/kubernetes/issues/60807#issuecomment-429754733
[2] https://github.com/code-ready/snc/blob/master/snc.sh#L184-L186

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -202,3 +202,7 @@ ${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/d
 
 # Set default route for registry CRD from false to true.
 ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
+
+# Delete the v1beta1.metrics.k8s.io apiservice since we are already scale down cluster wide monitioring.
+# Since this CRD block namespace deletion forever.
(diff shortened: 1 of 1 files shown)


snc.sh
```diff
@@ -202,3 +202,7 @@ ${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/d
 
 # Set default route for registry CRD from false to true.
 ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
+
+# Delete the v1beta1.metrics.k8s.io apiservice since we are already scale down cluster wide monitioring.
+# Since this CRD block namespace deletion forever.
+${OC} delete apiservice v1beta1.metrics.k8s.io
```
anjannath: :+1:

Issue #57, opened 2019-07-04
Remove pull secret from the disk image.


praveenkumar:
We need to replace the pull secret from the disk image so we can easily share it, right now we have 2 different places which it is listed and should be removed.

```
$ cat pull-secret.yaml 
apiVersion: v1
data:
  .dockerconfigjson: e30K
kind: Secret
metadata:
  name: pull-secret
  namespace: openshift-config
type: kubernetes.io/dockerconfigjson

$ oc get secrets pull-secret -n openshift-config -oyaml
```
- `/var/lib/kubelet/config.json`

anjannath:
FIxed in #58

PR #58, opened 2019-07-05
Issue #57 Remove pull secret from k8s secret resource

praveenkumar:
We have pull secret part of etcd db which is replaced by an empty
json string `{}` and also remove it from the VM disk side.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -7,6 +7,7 @@ set -x
 
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
+OC=${OC:-oc}
 
 function get_git_tag {
     GIT_TAG=$(git describe --exact-match HEAD) || GIT_TAG=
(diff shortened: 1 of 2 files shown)


createdisk.sh
```diff
@@ -166,6 +167,11 @@ if [ -z $random_string ]; then
 fi
 VM_PREFIX=${CRC_VM_NAME}-${random_string}
 
+# Replace pull secret to a null json string '{}'
```
cfergeau: 'replace ... with a null json string...'

createdisk.sh
```diff
@@ -166,6 +167,11 @@ if [ -z $random_string ]; then
...
+${OC} --config $1/auth/kubeconfig replace -f pull-secret.yaml
+# Remove pull secret from the VM
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -f /var/lib/kubelet/config.json
```
cfergeau: This should probably go after the `systemctl stop kubelet` call (and after making sure it really is stopped)? I'm wondering if it's possible that something recreates that file before it was removed. Maybe we should remove it with guestfish once the VM is shutdown?
What about https://github.com/cri-o/cri-o/pull/2510/files which apparently will duplicate the pull secret in a 3rd place? How do we ensure that from now on, the pull secret won't be duplicated in more places?

praveenkumar: @cfergeau https://github.com/cri-o/cri-o/pull/2510/files in this PR it doesn't duplicate the pull secret in 3rd place but use the existing one.  Yes, we need to keep the on disk removal part after the kublet service is stopped and all the containers/pods are deleted

cfergeau: Yup, sorry, I didn't look close enough :-/

createdisk.sh
praveenkumar: Done.

gbraad:
OK to merge or wait for the CRC side?

PR #64, opened 2019-07-19
Add support to generate hyperV disk images

praveenkumar:


createdisk.sh
```diff
@@ -199,6 +216,19 @@ kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cm
...
+sudo yum install --downloadonly --downloaddir $1/hyperv hyperv-daemons
+
+# SCP the downloaded rpms to VM
+${SCP} -r $1/hyperv core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
+
+# Install the hyperV rpms to VM
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/core/hyperv/*.rpm'
+
+# Remove the packages from VM
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/hyperv
```
gbraad: :+1

cfergeau: This means we will have the hyperv-daemons on libvirt, vbox and hyperkit too? Probably not a huge deal, but maybe we'll want to avoid these in the future.

createdisk.sh
```diff
@@ -136,7 +136,30 @@ function generate_hyperkit_directory {
         | ${JQ} ".nodes[0].kernel = \"vmlinuz-${kernel_release}\"" \
         | ${JQ} ".nodes[0].initramfs = \"initramfs-${kernel_release}.img\"" \
         | ${JQ} ".nodes[0].kernelCmdLine = \"${kernel_cmd_line}\"" \
-	| ${JQ} '.driverInfo.name = "hyperkit"' \
+        | ${JQ} '.driverInfo.name = "hyperkit"' \
+        >$destDir/crc-bundle-info.json
+}
+
```
cfergeau: I'd prefer that these space changes go in a separate commit

PR #65, opened 2019-07-19
Don't include crc_libvirt.sh script to final disk tarball

praveenkumar:
Right now when a user extract the bundle it finds out there
is a crc_libvirt.sh file and try to execute it which is not
expected from crc side so we should don't include it to the
final tarball.

On the other side this script is nice to do quick testing
on the generated disk image in case any debugging required so
just have it part of source directory.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -26,7 +26,7 @@ function get_git_tag {
 }
 
 function create_crc_libvirt_sh {
-    destDir=$1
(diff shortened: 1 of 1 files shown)


praveenkumar:
@cfergeau can you again have a look to it?

cfergeau:
Still looks good to me, ack.

Issue #66, opened 2019-07-24
Update kubelet.service in 99_master-kubelet-no-taint.yaml file.


praveenkumar:
We are using an older service file which is not going to work with `4.2.0` side if we don't make changes since now by default the `crio` service is not running and this is something specified in the `Wants=rpc-statd.service crio.service` and `After=crio.service` as part of the kubelet unit file
. 

Current kubelet unit file on the 4.2 side looks like.

```
$ sudo systemctl cat kubelet
# /etc/systemd/system/kubelet.service
[Unit]
Description=Kubernetes Kubelet
Wants=rpc-statd.service crio.service
After=crio.service

[Service]
Type=notify
ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
EnvironmentFile=/etc/os-release
EnvironmentFile=-/etc/kubernetes/kubelet-workaround
EnvironmentFile=-/etc/kubernetes/kubelet-env

ExecStart=/usr/bin/hyperkube \
    kubelet \
      --config=/etc/kubernetes/kubelet.conf \
      --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig \
      --rotate-certificates \
      --kubeconfig=/var/lib/kubelet/kubeconfig \
      --container-runtime=remote \
      --container-runtime-endpoint=/var/run/crio/crio.sock \
      --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
      --minimum-container-ttl-duration=6m0s \
      --cloud-provider= \
      --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
       \
      --register-with-taints=node-role.kubernetes.io/master=:NoSchedule \
      --v=3 \

Restart=always
RestartSec=10

[Install]
WantedBy=multi-user.target
```

codingarchitect:
Thanks a lot for this, this saved a few hours.

praveenkumar:
fixed by #70

PR #70, opened 2019-07-29
Issue #66 Update kubelet unit file

praveenkumar:
With openshift 4.2 current kubelet unit file which is part of
machineconfig resource need update. On 4.2 nodes by default doesn't
have `crio` service up and running and kubelet unit file now start
it since it have `wants` section. This changes also work with 4.1
since in that case it just get the `crio` already running.

### DIFF
--- a/99_master-kubelet-no-taint.yaml
+++ b/99_master-kubelet-no-taint.yaml
@@ -13,12 +13,14 @@ spec:
       - contents: |
           [Unit]
           Description=Kubernetes Kubelet
-          Wants=rpc-statd.service
+          Wants=rpc-statd.service crio.service
+          After=crio.service
 
           [Service]
           Type=notify
           ExecStartPre=/bin/mkdir --parents /etc/kubernetes/manifests
           ExecStartPre=/bin/rm -f /var/lib/kubelet/cpu_manager_state
+          EnvironmentFile=/etc/os-release
           EnvironmentFile=-/etc/kubernetes/kubelet-workaround
           EnvironmentFile=-/etc/kubernetes/kubelet-env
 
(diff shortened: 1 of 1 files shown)


99_master-kubelet-no-taint.yaml
```diff
@@ -31,13 +33,14 @@ spec:
                 --container-runtime=remote \
                 --container-runtime-endpoint=/var/run/crio/crio.sock \
                 --allow-privileged \
-                --node-labels=node-role.kubernetes.io/master \
+                --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=${ID} \
```
anjannath: What is the `ID` here?

praveenkumar:
```
$ cat /etc/os-release 
NAME="Red Hat Enterprise Linux CoreOS"
VERSION="410.8.20190710.0"
VERSION_ID="4.1"
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 410.8.20190710.0 (Ootpa)"
ID="rhcos"
ID_LIKE="rhel fedora"
ANSI_COLOR="0;31"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="OpenShift Container Platform"
REDHAT_BUGZILLA_PRODUCT_VERSION="4.1"
REDHAT_SUPPORT_PRODUCT="OpenShift Container Platform"
REDHAT_SUPPORT_PRODUCT_VERSION="4.1"
OSTREE_VERSION=410.8.20190710.0
```

ID here `rhcos`

anjannath: Hmm Thanks,.. and this line sources that file: `EnvironmentFile=/etc/os-release`

Issue #68, opened 2019-07-25
Add virt-sparsify 


praveenkumar:
```
$ qemu-img info crc.qcow2 
image: crc.qcow2
file format: qcow2
virtual size: 31G (33285996544 bytes)
disk size: 9.0G
cluster_size: 65536
Format specific information:
    compat: 1.1
    lazy refcounts: true
    refcount bits: 16
    corrupt: false

$ virt-sparsify ./crc.qcow2 ./crc_sparse.qcow2 --tmp tmp/
[   0.1] Create overlay file in tmp/ to protect source disk
[   0.1] Examine source disk
 100% ⟦▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒⟧ --:--
[   5.9] Fill free space in /dev/sda2 with zero
[   6.7] Fill free space in /dev/sda3 with zero
[...]

PR #71, opened 2019-08-01
Issue #68 Use virt-sparsify on created qcow2 image.

praveenkumar:
virt-sparsify can be used to create vm disk sparse. It is
reducing around ~700MB disk size and also help us to have
smaller tarball.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -10,7 +10,7 @@ SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 OC=${OC:-oc}
 
 function get_git_tag {
-    GIT_TAG=$(git describe --exact-match HEAD) || GIT_TAG=
+    GIT_TAG=$(git describe --exact-match --tags HEAD) || GIT_TAG=
 
     # Based on code from git-version-gen
     # Don't declare a version "dirty" merely because a time stamp has changed
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -54,6 +54,10 @@ function create_qemu_image {
     ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 31G
     ${VIRT_RESIZE} --expand /dev/sda3 $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
 
+    # Make sure you have enough space in /tmp to run VIRT_SPARSIFY
+    # Read limitation section of `man virt-sparsify`.
+    ${VIRT_SPARSIFY} --in-place $destDir/${CRC_VM_NAME}.qcow2
+
     rm -fr $destDir/${VM_PREFIX}-master-0 $destDir/${VM_PREFIX}-base
 }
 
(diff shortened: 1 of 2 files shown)


createdisk.sh
```diff
@@ -54,6 +54,10 @@ function create_qemu_image {
     ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 31G
     ${VIRT_RESIZE} --expand /dev/sda3 $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
 
+    # Make sure you have enough space to tmp to run VIRT_SPARSIFY
+    # Read limitation section of `man virt-sparsify`.
+    ${VIRT_SPARSIFY} --in-place $destDir/${CRC_VM_NAME}.qcow2
```
cfergeau: Previous code used to set TMPDIR to the current dir:
```
    # TMPDIR must point at a directory with as much free space as the size of the
    # image we want to sparsify
    TMPDIR=$(pwd)/$destDir ${VIRT_SPARSIFY} $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
    rm -fr $destDir/.guestfs-*
```
https://github.com/code-ready/snc/commit/f0768092dc2f0ffc78853dc2165ff109f0e5a0bf

createdisk.sh
```diff
@@ -228,7 +228,7 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crict
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q)'
```
cfergeau: Maybe commit log should be "We are not using sudo to remove the `config.json` file from `/etc/kubernetes`. This silently fails, and the resulting bundle contains this file."

createdisk.sh
praveenkumar: @cfergeau right but right now we are using the `--inplace` which doesn't work with `tmp` dir. I did try using `--tmp` option along with `--in-place` and that error out.

cfergeau: setting `TMPDIR` should not error out. However, 
`If you are using the --in-place option, then large amounts of temporary space are not required.` so I'm not sure the comment you added is needed at all?

praveenkumar: @cfergeau I just put `enough space` not large amount which you are right required only if we don't use the `inplace` option.

cfergeau: "to tmp" -> "in /tmp" if you want to keep the comment

createdisk.sh
```diff
@@ -10,7 +10,7 @@ SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 OC=${OC:-oc}
 
 function get_git_tag {
-    GIT_TAG=$(git describe --exact-match HEAD) || GIT_TAG=
+    GIT_TAG=$(git describe --exact-match --tags HEAD) || GIT_TAG=
```
cfergeau: Why don't  we sign our tags though? :)

createdisk.sh
cfergeau: commit log does not seem to have been adjusted in the latest version?

createdisk.sh
gbraad: I can not judge about this, as this is not something I can test now

createdisk.sh
praveenkumar: @cfergeau it's updated :) https://github.com/code-ready/snc/pull/71/commits/4d78be025616ae861e60bf22c2439ac9f6bde4e4

Issue #72, opened 2019-08-08
Block the createdisk script if cert is not rotated for 30 days


praveenkumar:
Right now we don't have any check for cert rotation, we should first check if the cert rotated successfully for 30 days and then only start the disk creation process otherwise error out and tell user to wait for it. It will help our CI and also users who want to create the bundle using this script.

gbraad:
Right, i wanted ot ask about this for the automated snc generation., You need to poll UNTIL the cert has rotated before the build can continue to package the images.

gbraad:
@praveenkumar see https://bugzilla.redhat.com/show_bug.cgi?id=1736800#c9
```
oc get secret -n openshift-kube-apiserver -A -o json | jq -r '.items[] | select(.metadata.annotations."auth.openshift.io/certificate-not-after" | .!=null and fromdateiso8601<='$( date --date='+1year' +%s )') | "-n \(.metadata.namespace) \(.metadata.name)"' | xargs -n3 oc patch secret -p='{"metadata": {"annotations": {"auth.openshift.io/certificate-not-after": null}}}'
```

anjannath:
When i tested this out, it didn't work for me, waited for 15-20mins and the certs were not rotated.

praveenkumar:
Fixed #74

PR #74, opened 2019-08-09
Issue #72 Add check for cert rotation

praveenkumar:
Check if the cert is rotated initially and now have 30 days validity.
Exit early in case the rotation doesn't happened yet.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -210,6 +210,15 @@ if [ -z $random_string ]; then
 fi
 VM_PREFIX=${CRC_VM_NAME}-${random_string}
 
+# First check if cert rotation happened.
+# Initial certificate is only valid for 24 hours, after rotation, it's valid for 30 days.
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -210,6 +210,12 @@ if [ -z $random_string ]; then
 fi
 VM_PREFIX=${CRC_VM_NAME}-${random_string}
 
+# First check if cert rotation happened successfully and valid for 30 days
```
cfergeau: The comment is misleading as this only checks that the certificate will be valid for more than 24 hours starting from the time of the check. You could check for 25*86400 or something like this, and be more explicit in the comment or the log.
`# First check if cert rotation happened. Initial certificate is only valid for 24 hours, after rotation, it's valid for 30 days. We check if it's valid for more than 25 days rather than 30 days to give us some leeway regarding when we run the check with respect to rotation time`

PR #76, opened 2019-08-13
Issue #73 Add an user developer with password developer

### LINKED ISSUE #73: Create a user developer with password as developer

anjannath:
Fixes #73

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -8,6 +8,7 @@ set -x
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 OC=${OC:-oc}
+DEVELOPER_USER_PASS='developer:$2y$05$paX6Xc9AiLa6VT7qr2VvB.Qi.GJsaqS80TR3Kb78FEIlIL0YyBuyS'
 
 function get_git_tag {
     GIT_TAG=$(git describe --exact-match --tags HEAD) || GIT_TAG=
@@ -218,6 +219,11 @@ if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checken
     echo "Certs are not yet rotated to have 30 days validity"
     exit 1;
 fi
+
+# Add a user developer:developer with htpasswd identity provider and give it sudoer role
+${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
+${OC} --config $1/auth/kubeconfig apply -f htpasswd_cr.yaml
+${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterrole=sudoer --user=developer
 
 # Replace pull secret with a null json string '{}'
 ${OC} --config $1/auth/kubeconfig replace -f pull-secret.yaml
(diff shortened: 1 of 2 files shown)


htpasswd_cr.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: cluster
+spec:
+  identityProviders:
+  - name: htpasswd_provider 
+    mappingMethod: claim 
+    type: HTPasswd
+    htpasswd:
+      fileData:
+        name: htpass-secret
```
gbraad: missimg something?

anjannath: Nope, its done!
Wait did you find something?

createdisk.sh
```diff
@@ -225,6 +226,10 @@ ${OC} --config $1/auth/kubeconfig replace -f pull-secret.yaml
 # Remove the Cluster ID with a empty string.
 ${OC} --config $1/auth/kubeconfig patch clusterversion version -p '{"spec":{"clusterID":""}}' --type merge
 
+# Add a user developer:developer with htpasswd identity provider
+${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
+${OC} --config $1/auth/kubeconfig apply -f htpasswd_cr.yaml
+
```
praveenkumar: @anjannath can you also make this user to act as sudo, to do that `oc create clusterrolebinding developer --clusterrole=sudoer --user=developer` you just need this one https://docs.openshift.com/container-platform/4.1/authentication/impersonating-system-admin.html

htpasswd_cr.yaml
gbraad: It says: `filedata`, but i do not see a file ... ?

createdisk.sh
anjannath: Okay, doing it right away!

htpasswd_cr.yaml
anjannath: So the htpasswd file content is stored as a secret in OpenShift which has the name `htpass-secret`, also instead of adding a new file (with just one line), the file's content is stored in a variable, both these changes are in the `createdisk.sh` file.

createdisk.sh
```diff
@@ -225,6 +226,11 @@ ${OC} --config $1/auth/kubeconfig replace -f pull-secret.yaml
 # Remove the Cluster ID with a empty string.
 ${OC} --config $1/auth/kubeconfig patch clusterversion version -p '{"spec":{"clusterID":""}}' --type merge
 
+# Add a user developer:developer with htpasswd identity provider
+${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
```
gbraad: ah, here ...

createdisk.sh
praveenkumar: @anjannath these changes you should do before we remove cluster pull secret and cluster ID because we don't know if there is any downtime for apiserver once we make those changes. should come https://github.com/code-ready/snc/blob/master/createdisk.sh#L221 here

anjannath: @praveenkumar Done!

createdisk.sh
```diff
@@ -219,6 +220,11 @@ if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checken
     exit 1;
 fi
 
+# Add a user developer:developer with htpasswd identity provider and give it sudoer role
+${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
+${OC} --config $1/auth/kubeconfig apply -f htpasswd_cr.yaml
+${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterrole=sudoer --user=developer
```

Issue #80, opened 2019-08-22
kernel command line argument should be taken after installing any packages and performing a reboot


praveenkumar:
https://github.com/code-ready/crc/issues/420 was caused because we installed the hyperV packages which change the ostree. Since we didn't rebooted the system it doesn't switch the ostree and caused this issue. We need to either get that kernel line args before or reboot the system and then get it.

cfergeau:
I don't think getting the kernel commandline before package installation will help. We either need to build it from `ostree admin status`, or reboot after installing the hyperv packages. I'd go with the latter.

praveenkumar:
@cfergeau I didn't check before creating the issue looks like we are already get the kernel cmdline https://github.com/code-ready/snc/blob/master/createdisk.sh#L248-L249 before even install the packages https://github.com/code-ready/snc/blob/master/createdisk.sh#L261-L262 which means issue is somewhere else. I will take closer look.

cfergeau:
Getting the command line before installing the packages is wrong, as installing the packages will:
- create a new ostree hash so that it's possible to get a tree with the packages installed
- (I think) trigger a switch between boot.0 and boot.1 on next reboot so that the new tree is picked

We want the new tree to be picked as the one before installing the packages will not contain the newly installed packages. So we need to install the packages, reboot, and finally get the kernel command line.

praveenkumar:
> Getting the command line before installing the packages is wrong, as installing the packages will:

@cfergeau Right, new tree is anyway picked in case of libvirt and hyperv or VB since when next reboot happens it will happen on the latest tree, problem is only with hyperkit since it need kernel cmdline which is different than the rebooted system. so we do need to do a reboot and then get the kernel cmdline.

praveenkumar:
fixed by #81

PR #81, opened 2019-08-22
Issue #80 Restart the VM after installing hyperV rpms.

praveenkumar:
Installation of hyperV daemons rpms add a new ostree layer which need
to be consumed to get the proper value of vmlinuz, initrd and kernel
cmdline.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -239,18 +239,6 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl 
 # Remove pull secret from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
 
-# Get the rhcos ostree Hash ID
-ostree_hash=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline | grep -oP "(?<=rhcos-).*(?=/vmlinuz)"')
-
-# Get the rhcos kernel release
-kernel_release=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'uname -r')
-
-# Get the kernel command line arguments
-kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline')
-
-# SCP the vmlinuz/initramfs from VM to Host in provided folder.
-${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/rhcos-${ostree_hash}/* $1
-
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -264,7 +252,35 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
...
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/hyperv
 
-# Shutdown the instance
+# Shutdown and Start the VM after installing the hyperV daemon packages.
+# This is required to get the latest ostree layer which have those installed packages.
+sudo virsh shutdown ${VM_PREFIX}-master-0
+# Wait till instance started successfully
+until sudo virsh domstate ${VM_PREFIX}-master-0 | grep shut; do
+    echo " ${VM_PREFIX}-master-0 still running"
+    sleep 10
```
cfergeau: That's a bit long in my opinion, I'd go with just 1 or 2 seconds

createdisk.sh
```diff
@@ -264,7 +252,35 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
 # Remove the packages from VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/hyperv
 
-# Shutdown the instance
+# Shutdown and Start the VM after installing the hyperV daemon packages.
+# This is required to get the latest ostree layer which have those installed packages.
+sudo virsh shutdown ${VM_PREFIX}-master-0
+# Wait till instance started successfully
+until sudo virsh domstate ${VM_PREFIX}-master-0 | grep shut; do
```
cfergeau: you need LC_ALL=C or similar, on my machine `domstate` reports `fermé` for a shut off VM.

createdisk.sh
```diff
@@ -264,7 +252,35 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
...
+    echo " ${VM_PREFIX}-master-0 still running"
+    sleep 10
+done
+
+sudo virsh start ${VM_PREFIX}-master-0
+# Wait till it is started properly.
+until ping -c1 api.${CRC_VM_NAME}.${BASE_DOMAIN} >/dev/null 2>&1; do
+    echo " ${VM_PREFIX}-master-0 still booting"
+    sleep 10
```
cfergeau: Same comment about 10s being a bit too much.

createdisk.sh
praveenkumar: @cfergeau `export LC_ALL=C` and `export LANG=C` area already as part of initial of the script so we don't need for every command.

createdisk.sh
praveenkumar: sure  I can put 2-3 sec.

PR #83, opened 2019-08-23
Issue 82 Implement to take git tag as argument

praveenkumar:
In CI side this will be useful since we can use single job pipeline
to create the builds for master (in that case unreleased openshift)
and also tagged one.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -8,7 +8,7 @@ JQ=${JQ:-jq}
 OC=${OC:-oc}
 YQ=${YQ:-yq}
 OPENSHIFT_INSTALL=${OPENSHIFT_INSTALL:-./openshift-install}
-OPENSHIFT_RELEASE_VERSION=$(git describe --abbrev=0 --tags 2>/dev/null) || OPENSHIFT_RELEASE_VERSION=
+OPENSHIFT_RELEASE_VERSION=$(git describe --exact-match --tags HEAD 2>/dev/null) || OPENSHIFT_RELEASE_VERSION=
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
--- a/snc.sh
+++ b/snc.sh
@@ -26,7 +26,6 @@ function create_json_description {
             | ${JQ} ".clusterInfo.clusterName = \"${CRC_VM_NAME}\"" \
             | ${JQ} ".clusterInfo.baseDomain = \"${BASE_DOMAIN}\"" \
             | ${JQ} ".clusterInfo.appsDomain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
-    #        |${JQ} '.buildInfo.ocGetCo = "snc"' >${INSTALL_DIR}/crc-bundle-info.json
 }
 
 function generate_pv() {
--- a/snc.sh
+++ b/snc.sh
@@ -4,7 +4,6 @@ export LC_ALL=C
 export LANG=C
 
 INSTALL_DIR=crc-tmp-install-data
-INSTALLER_RELEASE=v0.14.0
 JQ=${JQ:-jq}
 OC=${OC:-oc}
 YQ=${YQ:-yq}
--- a/snc.sh
+++ b/snc.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 export LC_ALL=C
 export LANG=C


createdisk.sh
```diff
@@ -10,20 +10,18 @@ SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 OC=${OC:-oc}
```
cfergeau: From the commit log, I'm not sure I understand the issue there is with CI :) It seems you want to pretend you have a tag when nothing was tagged?

createdisk.sh
```diff
@@ -221,15 +236,15 @@ if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checken
 fi
 
 # Add a user developer:developer with htpasswd identity provider and give it sudoer role
-${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
-${OC} --config $1/auth/kubeconfig apply -f htpasswd_cr.yaml
-${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterrole=sudoer --user=developer
+${OC} --config $clusterDirectory/auth/kubeconfig create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
+${OC} --config $clusterDirectory/auth/kubeconfig apply -f htpasswd_cr.yaml
+${OC} --config $clusterDirectory/auth/kubeconfig create clusterrolebinding developer --clusterrole=sudoer --user=developer
```
cfergeau: I probably would have split the "clusterDirectory" change and the --help addition to a separate commit, it's then easier to see the "important" changes made by this commit (ie the tag-related ones).

createdisk.sh
```diff
@@ -170,6 +168,19 @@ function generate_hyperv_directory {
         >$destDir/crc-bundle-info.json
 }
 
+function usage()
```
gbraad: :-O...

createdisk.sh
praveenkumar: @cfergeau `$currentDirectory` comes from the tag related changes. since I have to implement the parse flag and this should be part of it. I am not sure if I split it out then the script even works as expected.

cfergeau: You can start by adding usage() which will document -d / -h, and s/$1/$clusterDirectory, and then do the tag related changes on top of it. Or am I missing something which would prevent this from working?

praveenkumar: @cfergeau no, I can't do that since then I also have to remove otherwise any option we provide along with `cluster directory` this will not allow it to pass. And if I remove that then there `$1` is something is not an expected value so it is tie with current commit.

```bash
if [[ $# -ne 1 ]]; then
   echo "You need to provide the running cluster directory to copy kubeconfig"	
fi
```

cfergeau:
Still trying to wrap my head around this change and why this is needed.

Before these changes, snc.sh looks up the latest tag in the current branch in order to set OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to the correct openshift version, and to download the appropriate oc version.
One thing to note is that OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE is already a workaround, we should be specifying the image to use when building openshift-install rather than overriding it.

After the changes in this PR, I'm under the impression that the behaviour without the --tag parameter has changed, it's no longer use the latest tag set, but instead will use the image the installer was built against.


Regarding createdisk.sh, the tag is used to name the tarball and the directory it contains. If the current commit is tagged, then we will use the tag name in the tarball name. If it's not tag, it's a git snapshot, and we will use append the date to the tarball name.

After the changes in this PR, if --tag was used and the tag is known, then we use this as the tarball name (regardless of the current commit being tagged or not), if --tag was not used, then we used the date.


All the above means there's an assumption that we'll build the images for a given openshift releases from *one* snc commit. This is not necessarily true, sometimes we'll want to regenerate the images for a given openshift release from a different commit than the one which was tagged.

What this commit is doing to remove this assumption is that it drops all the guessing of the openshift release version/git snapshot by looking at `git tag`, and instead decides between release build/snapshot build depending on whether --tag was specified on the command line or not. This is a fine way to go if we want maximum flexibility, but if we go this way, I believe we should drop everything tag related, looking at the git tags is not really useful, and does not seem to be what we want to do. So just a --openshift-release flag would do the trick.

However, at the moment, I'm not sure it's useful to support arbitrary openshift releases, if we have to rebuild a bundle because of issue on the openshift/coreos/.. side, wouldn't we still base our work against the latest 4.1.x image? or are there situations when we'd really need to rebuild an older 4.1.y image when a newer 4.1.x release exists? I'm under the impression that what we need *now* is a way to rebuild a bundle against the latest openshift release, but with a snc commit which is not the not that we tagged. If that's enough, couldn't we only add a `--release` parameter (with no version after it) which we'll only specify in such situations, and keep all the automatic git stuff to detect snapshots/releases?

praveenkumar:
> All the above means there's an assumption that we'll build the images for a given openshift releases from one snc commit. This is not necessarily true, sometimes we'll want to regenerate the images for a given openshift release from a different commit than the one which was tagged.

Right, I thought it over and whatever discussion we had and I think what I just need on the CI side or as a user to create a openshift cluster from non released bits and right now we are holding this because of `git describe --abbrev=0 --tags` in the `snc.sh` since it will always assign `OPENSHIFT_RELEASE_VERSION` to latest tag even with the master branch but we actually need something like `git describe --exact-match --tags HEAD` which will make sure until I checked out a release tag I am able to use the master like the way we are using with `createdisk.sh` side.

My intention is not to have arbitrary cluster deployment with a single snc commit because that will defeat the purpose of this repo. I want to have it proper release tags as we move so that when  other changes are 4.2.x and 4.3.x we can still able to deploy cluster using our 4.1.x release tags.

I am going to update this PR with my current understanding and hope that would be less distractive then this one. Thanks @cfergeau for detailed view.

snc.sh
```diff
@@ -8,7 +8,7 @@ JQ=${JQ:-jq}
 OC=${OC:-oc}
 YQ=${YQ:-yq}
 OPENSHIFT_INSTALL=${OPENSHIFT_INSTALL:-./openshift-install}
-OPENSHIFT_RELEASE_VERSION=$(git describe --abbrev=0 --tags 2>/dev/null) || OPENSHIFT_RELEASE_VERSION=
+OPENSHIFT_RELEASE_VERSION=$(git describe --exact-match --tags HEAD 2>/dev/null) || OPENSHIFT_RELEASE_VERSION=
```
cfergeau: Sure, though I'm worried this might need to be more sophisticated, as we want totest 4.1.latest, and probably y4.2.latest. Maybe this should be handled through installer build parameters rather than with version overrides here

cfergeau: I just mean that I foresee more work needed in that area, this particular commit is fine with me though, and the patch series too.

Issue #84, opened 2019-08-25
Add qemu-image option lazy_refcounts=on for virt-sparsify 


praveenkumar:
We need to update the virt-sparsify command since it is causing issue for created disk images on the hyperkit. In 4.1.6 tag we didn't used `virt-sparsify` and it was working without any issue for `stop => start` but 4.1.9 is not working. Looking closer looks like if we are using the `virt-sparsify` without `qemu-img` options `lazy_refcounts` then after stop the VM the disk image no longer a qcow2 and it failed to boot.

From cache when we fresh create it have `qcow2` so that works but `stop=>start` not works.
```
$ file ~/.crc/cache/crc_hyperkit_4.1.11/crc.qcow2 
/Users/prkumar/.crc/cache/crc_hyperkit_4.1.11/crc.qcow2: QEMU QCOW Image (v3), 33285996544 bytes
```

After stop and start.

```
$ sudo file /Users/prkumar/.crc/machines/crc/crc.disk 
/Users/prkumar/.crc/machines/crc/crc.disk: data

[...]

PR #85, opened 2019-08-27
 Issue #84 Add lazy_refcounts option to virt-sparsify cmd

praveenkumar:
virt-sparsify command allow most of the `qemu-img` options
but we can't use it for `--in-place` so need to remove it and
add `lazy_refcounts=on`. This will enable hyperkit to behave
sane when vm is started from stop stage.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -55,11 +55,18 @@ function create_qemu_image {
     ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 31G
     ${VIRT_RESIZE} --expand /dev/sda3 $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
 
-    # Make sure you have enough space in /tmp to run VIRT_SPARSIFY
+    # TMPDIR must point at a directory with as much free space as the size of the image we want to sparsify
     # Read limitation section of `man virt-sparsify`.
     TMPDIR=$(pwd)/$destDir ${VIRT_SPARSIFY} -o lazy_refcounts=on $destDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}_sparse.qcow2
     rm -f $destDir/${CRC_VM_NAME}.qcow2
     mv $destDir/${CRC_VM_NAME}_sparse.qcow2 $destDir/${CRC_VM_NAME}.qcow2
+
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -61,6 +61,13 @@ function create_qemu_image {
     rm -f $destDir/${CRC_VM_NAME}.qcow2
     mv $destDir/${CRC_VM_NAME}_sparse.qcow2 $destDir/${CRC_VM_NAME}.qcow2
 
+    # Before using the created qcow2 check if it have lazy_refcounts true.
```
cfergeau: "Before using the created qcow2, check if it has lazy_refcounts set to true."

createdisk.sh
```diff
@@ -61,6 +61,13 @@ function create_qemu_image {
...
+    lazy_refcounts=$(${QEMU_IMG} info ${destDir}/${CRC_VM_NAME}.qcow2 | grep -oP "lazy refcounts: true")
+    if [ -z $lazy_refcounts ]; then
+        echo "${CRC_VM_NAME}.qcow2 doesn't have lazy_refcounts enabled"
```
cfergeau: Maybe also explain in the error message that "This is going to cause disk image corruption when using hyperkit"?

createdisk.sh
```diff
@@ -57,7 +57,9 @@ function create_qemu_image {
 
     # Make sure you have enough space in /tmp to run VIRT_SPARSIFY
     # Read limitation section of `man virt-sparsify`.
-    ${VIRT_SPARSIFY} --in-place $destDir/${CRC_VM_NAME}.qcow2
+    ${VIRT_SPARSIFY} -o lazy_refcounts=on $destDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}_sparse.qcow2
```
cfergeau: I'd readd the `TMPDIR=$(pwd)/$destDir` which we had before commit f0768092

createdisk.sh
```diff
@@ -57,7 +57,16 @@ function create_qemu_image {
 
     # Make sure you have enough space in /tmp to run VIRT_SPARSIFY
```
anjannath: Does this comment need to change, since `TMPDIR=$(pwd)/$destDir` is used now?

cfergeau: Ah right. This used to be "TMPDIR must point at a directory with as much free space as the size of the image we want to sparsify" before commit f0768092d

PR #89, opened 2019-08-29
Add clusterInfo.openshiftVersion to crc-bundle-info.json

cfergeau:
It will be useful to crc to let the user know which version of OpenShift
their cluster is running.


snc.sh
```diff
@@ -23,6 +23,7 @@ function create_json_description {
             | ${JQ} ".buildInfo.buildTime = \"$(date -u --iso-8601=seconds)\"" \
             | ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\"" \
             | ${JQ} ".buildInfo.sncVersion = \"git${sncGitHash}\"" \
+            | ${JQ} ".clusterInfo.openshiftVersion = \"${OPENSHIFT_RELEASE_VERSION:-git}\"" \
```
praveenkumar: 👍

Issue #90, opened 2019-09-03
Change the install type for crc so it will list differently to telemeter


praveenkumar:
https://github.com/code-ready/crc/issues/508 this is something we need to do as part of snc side.

anjannath:
Fixed via https://github.com/code-ready/snc/commit/f80d3842d713ca92a479d6f0ee69ec5ce77903e1

PR #91, opened 2019-09-03
Issue #90 Add codeReadyContainer as invoker for installer

praveenkumar:
This change will make it easy to search the codeReadyContainer invokers
from the telemeter side.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -213,6 +213,9 @@ if ! which ${QEMU_IMG}; then
     sudo yum -y install /usr/bin/qemu-img
 fi
 
+# This random_string is created by installer and added to each resource type,
+# in installer side also variable name is kept as `random_string`
+# so to maintain consistancy, we are also using random_string here.
 random_string=$(sudo virsh list --all | grep -oP "(?<=${CRC_VM_NAME}-).*(?=-master-0)")
 if [ -z $random_string ]; then
     echo "Could not find virtual machine created by snc.sh"
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -12,18 +12,6 @@ DEVELOPER_USER_PASS='developer:$2y$05$paX6Xc9AiLa6VT7qr2VvB.Qi.GJsaqS80TR3Kb78FE
 
 function get_git_tag {
     GIT_TAG=$(git describe --exact-match --tags HEAD) || GIT_TAG=
-
-    # Based on code from git-version-gen
-    # Don't declare a version "dirty" merely because a time stamp has changed
-    git update-index --refresh > /dev/null 2>&1
-
-    dirty=`sh -c 'git diff-index --name-only HEAD' 2>/dev/null` || dirty=
-    case "$dirty" in
-        '') ;;
-        *) # Don't build an 'official' version if git tree is dirty
-            GIT_TAG=
-    esac
-    # end of git-version-gen code
 }
 
 function create_crc_libvirt_sh {
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -10,8 +10,11 @@ SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 OC=${OC:-oc}
 DEVELOPER_USER_PASS='developer:$2y$05$paX6Xc9AiLa6VT7qr2VvB.Qi.GJsaqS80TR3Kb78FEIlIL0YyBuyS'
 
-function get_git_tag {
-    GIT_TAG=$(git describe --exact-match --tags HEAD) || GIT_TAG=
+function get_dest_dir {
+    DEST_DIR=$(git describe --exact-match --tags HEAD)
+    if [ -z ${DEST_DIR} ]; then
+        DEST_DIR="$(date --iso-8601)"
+    fi
 }
 
 function create_crc_libvirt_sh {
@@ -298,13 +301,9 @@ done
 crcBundleSuffix=crcbundle
 
 # libvirt image generation
-get_git_tag
+get_dest_dir
+destDirSuffix="${DEST_DIR}"
 
-if [ -z ${GIT_TAG} ]; then
-    destDirSuffix="$(date --iso-8601)"
-else
-    destDirSuffix="${GIT_TAG}"
-fi
 libvirtDestDir="crc_libvirt_${destDirSuffix}"
 mkdir $libvirtDestDir
 
--- a/snc.sh
+++ b/snc.sh
@@ -8,12 +8,20 @@ JQ=${JQ:-jq}
 OC=${OC:-oc}
 YQ=${YQ:-yq}
 OPENSHIFT_INSTALL=${OPENSHIFT_INSTALL:-./openshift-install}
-OPENSHIFT_RELEASE_VERSION=$(git describe --exact-match --tags HEAD 2>/dev/null) || OPENSHIFT_RELEASE_VERSION=
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 
+# If user defined the OPENSHIFT_VERSION environment variable then use it.
+# Otherwise use the tagged version if available
+function get_openshift_version {
+    if [ ${OPENSHIFT_VERSION} != "" ]; then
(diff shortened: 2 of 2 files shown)


createdisk.sh
```diff
@@ -11,19 +11,12 @@ OC=${OC:-oc}
 DEVELOPER_USER_PASS='developer:$2y$05$paX6Xc9AiLa6VT7qr2VvB.Qi.GJsaqS80TR3Kb78FEIlIL0YyBuyS'
```
cfergeau: > Right now, if a user try to use master branch he is hoping get the
cluster deployed without any image override but this is not the case
right now since `git describe --exact-match --tags HEAD` provide latest
released tag and the image override happen.

This is only true if current git master HEAD is a tagged commit. 
```
[teuf@natto snc ((4.1.13))]$ LC_ALL=C git describe --exact-match --tags 
4.1.13
[teuf@natto snc ((4.1.13))]$ git checkout HEAD^

[teuf@natto snc ((83556dc...))]$ LC_ALL=C git describe --exact-match --tags 
fatal: no tag exactly matches '83556dccfa7dc208b3357c3e050dc559d7ef0657'
```
Since issues related to this have come up a few times recently, I'm not sure this git tag versioning is what we want, as apparently you want to be able to build both releases (which for now are tagged) and snapshots from the same snc commit. So maybe it's time to retire this git tag code and replace it with something else? Or maybe keep the tag versioning if this has benefits, and add a --snapshot argument (or env var)?

createdisk.sh
```diff
@@ -11,19 +11,12 @@ OC=${OC:-oc}
...
-    # Don't declare a version "dirty" merely because a time stamp has changed
-    git update-index --refresh > /dev/null 2>&1
-
-    dirty=`sh -c 'git diff-index --name-only HEAD' 2>/dev/null` || dirty=
-    case "$dirty" in
-        '') ;;
-        *) # Don't build an 'official' version if git tree is dirty
-            GIT_TAG=
-    esac
-    # end of git-version-gen code
```
cfergeau: I would not drop this code in this commit, but in a separate one if that's what we want.

createdisk.sh
```diff
@@ -11,19 +11,12 @@ OC=${OC:-oc}
...
+    BRANCH=$(git rev-parse --abbrev-ref HEAD)
+    if [[ "$BRANCH" == "master" ]]; then
+      GIT_TAG=
+      return
+    fi
+    GIT_TAG=$(git describe --exact-match HEAD 2>/dev/null) || GIT_TAG=
```
cfergeau: So this means the git tag versioning code will only work if we create branches for releases? Is this what we are going to do? If latest git master commit is tagged, then `git rev-parse --abbrev-ref HEAD` will return `master`, and we'll never run the 'git describe` code. This means we went from "hard to create snapshots" to "hard to create releases" imo.

createdisk.sh
```diff
@@ -213,6 +213,8 @@ if ! which ${QEMU_IMG}; then
     sudo yum -y install /usr/bin/qemu-img
 fi
 
+# This random_string is created by installer and added to each resource type so
+# to maintain consistancy, we are also using random_string here.
```
cfergeau: "consistency". But I don't understand the clarification the comment is trying to give :( Are you saying internally, the installer calls it "random_string", and that we are using the same name?

snc.sh
```diff
@@ -189,7 +189,7 @@ ${OC} scale --replicas=0 deployment --all -n openshift-monitoring
 ${OC} scale --replicas=0 statefulset --all -n openshift-monitoring
```
cfergeau: "consistency" in the log

snc.sh
```diff
@@ -158,6 +167,10 @@ ${YQ} write --inplace $INSTALL_DIR/manifests/cluster-ingress-02-config.yml spec[
 # Start the cluster with 10GB memory and 4 CPU create and wait till it finish
 export TF_VAR_libvirt_master_memory=10192
 export TF_VAR_libvirt_master_vcpu=4
+
+# Add codeReadyContainer as invoker to identify it with telemeter
+export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
```
anjannath: :+1:

snc.sh
```diff
@@ -158,6 +158,10 @@ ${YQ} write --inplace $INSTALL_DIR/manifests/cluster-ingress-02-config.yml spec[
 # Start the cluster with 10GB memory and 4 CPU create and wait till it finish
```
cfergeau: "to search", and more importantly, "to filter it out" ;)

snc.sh
cfergeau: Any idea which case is typical for the invoker ("CodeReadyContainers" VS "codeReadyContainers", or even with a space "CodeReady Containers"?)

createdisk.sh
praveenkumar: yes Installer also using this as `random_string` so we are using same.

snc.sh
praveenkumar: @cfergeau whatever easy for us to use I mentioned `codeReadyContainers` to original issue and didn't see any push back so went with it :)

createdisk.sh
praveenkumar: @cfergeau right, even if the master branch have commit to a tag then also if a user manually checkout to a released tag it will not able to build the tagged version.

createdisk.sh
cfergeau: This is the important part, that this naming is used by the installer, so I would spell it explicitly in the comment.

createdisk.sh
```diff
@@ -297,12 +297,12 @@ done
...
 # libvirt image generation
-get_git_tag
+get_dest_dir
 
-if [ -z ${GIT_TAG} ]; then
+if [ -z ${DEST_DIR} ]; then
     destDirSuffix="$(date --iso-8601)"
 else
-    destDirSuffix="${GIT_TAG}"
+    destDirSuffix="${DEST_DIR}"
```
cfergeau: get_dest_dir_suffix would be more accurate given the way it's used, but then I would also move this `if [ -z  ${DEST_DIR} ]` block in the helper function.

snc.sh
```diff
@@ -8,12 +8,19 @@ JQ=${JQ:-jq}
...
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 
+# If user defined the OPENSHIFT_VERSION environment variable then use it.
+# Otherwise use the tagged version if available
+function get_openshift_version {
+    OPENSHIFT_RELEASE_VERSION=$(git describe --exact-match --tags HEAD 2>/dev/null)
+    if [ ${OPENSHIFT_VERSION} != "" ]; then
+        OPENSHIFT_RELEASE_VERSION=$OPENSHIFT_VERSION
+    fi
```
cfergeau: No need for the "git describe" when OPENSHIFT_VERSION is set? I believe this could be written as
```
if [ -z ${OPENSHIFT_VERSION} ]; then 
    OPENSHIFT_RELEASE_VERSION=$(git describe ...)
else 
    OPENSHIFT_RELEASE_VERSION=$OPENSHIFT_VERSION
fi
```
Maybe we could use "OPENSHIFT_RELEASE_VERSION" as the name of the environment variable rather than introducing an extra "OPENSHIFT_VERSION"? (just a thought, I'm fine either way)

createdisk.sh
cfergeau: No change in the comment?

snc.sh
praveenkumar: If a user set `OPENSHIFT_VERSION` then we are going to use that instead the tag, otherwise if the master branch have tagged then there is no way to deploy a different version then tagged one until switch to different branch or go to a different commit. So here we will give user the preference how they want to build.

praveenkumar:
> After introducing the OPENSHIFT_VERSION environment variable, are we really going to be using both the env var and the tagging?

@cfergeau yes, if I want to just spin up a cluster which I find in the tag then I checkout the tag (don't set the environment variable) and good to go. If I want to spin cluster (4.1.10) which is not tagged but near to tag-4.1.13 I can checkout that tag and set the environment variable and good to go.

snc.sh
cfergeau: If current git master tip is tagged, we can now build a different openshift version (say 4.1.11 instead of 4.1.13), but we cannot build an unspecified openshift version (ie use whatever version the installer was built to use). Is this enough for what you want to do?

cfergeau: Yes, I was not suggesting you change the behaviour of what you describe, I was just proposing alternative ways of doing the same, and avoid the "git describe" call when not needed.

Issue #75, opened 2019-08-09
snc build:  Unknown name requested, could not find keepalived-ipfailover in UpdatePayload


andrejpetras:
I installed the centos7 on the laptop and installed all dependencies for the openshift-install and snc script.
(Add kvm-common repository to have the latest qemu-kvm)
When I run the snc script. I get the error that the master was not started in the 30min.
```
[user@ce23652 snc]$ sudo virsh list --all
[sudo] password for user: 
 Id    Name                           State
----------------------------------------------------
 3     crc-85mk2-bootstrap            running
 4     crc-85mk2-master-0             running
```
```
[user@ce23652 snc]$ sudo virsh console 3
Connected to domain crc-85mk2-bootstrap
Escape character is ^]
[ 3126.214688] SELinux: mount invalid.  Same superblock, different security settings for (dev mqueue, type mqueue)
[ 3126.884085] SELinux: mount invalid.  Same superblock, different security settings for (dev mqueue, type mqueue)
[ 3127.608045] SELinux: mount invalid.  Same superblock, different security settings for (dev mqueue, type mqueue)
[ 3128.289552] SELinux: mount invalid.  Same superblock, different security settings for (dev mqueue, type mqueue)
```
```
[user@ce23652 snc]$ sudo virsh console 4
Connected to domain crc-85mk2-master-0
Escape character is ^]
[**    ] A start job is running for Ignition (disks) (52min 7s / no limit)[ 3130.116850] ignition[626]: GET https://api-int.crc.testing:22623/config/master: attempt #629
[ 3130.133065] ignition[626]: GET error: Get https://api-int.crc.testing:22623/config/master: dial tcp 192.168.126.10:22623: connect: connection refused
[  *** ] A start job is running for Ignition (disks) (52min 9s / no limit)
```

My crc-xxx-boostrap image does not start.
```
5b81 (image=quay.io/openshift-release-dev/ocp-release:4.1.9, name=stupefied_gagarin)
Aug 08 23:47:59 crc-8mjbc-bootstrap podman[2831]: 2019-08-08 23:47:59.806207983 +0000 UTC m=+0.303240560 container start ba321b1f5f60edd9596199dea4d33999a5d6b57cb9f3852e27d6741ecc645b81 (image=quay.io/openshift-release-dev/ocp-release:4.1.9, name=stupefied_gagarin)
[...]

PR #93, opened 2019-09-09
Issue #75 Add logic to get qcow2 location for 4.2 cluster

praveenkumar:
Due to https://github.com/openshift/installer/pull/1956 , now
by default the location of the qcow2 images are part of
`/var/lib/libvirt/openshift/<cluster_name>_<random_hash>` so
this change will make sure we are copy correct file as part of
our CI.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -38,8 +38,13 @@ function create_crc_libvirt_sh {
 function create_qemu_image {
     local destDir=$1
 
-    sudo cp /var/lib/libvirt/images/${VM_PREFIX}-master-0 $destDir
-    sudo cp /var/lib/libvirt/images/${VM_PREFIX}-base $destDir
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -38,8 +38,13 @@ function create_crc_libvirt_sh {
 function create_qemu_image {
     local destDir=$1
 
-    sudo cp /var/lib/libvirt/images/${VM_PREFIX}-master-0 $destDir
-    sudo cp /var/lib/libvirt/images/${VM_PREFIX}-base $destDir
+    if [ -f /var/lib/libvirt/images/${VM_PREFIX}-master-0 ]; then
```
cfergeau: Would be cleaner to check for libvirt storage pool existence, get the path from there, and use it. But given our use of sh to do all of this, the way you did is much easier/efficient.

PR #101, opened 2019-09-23
Add podman workaround to make it work along with cri-o.

praveenkumar:
Podman and crio both using the CNI networking and that cause the
race condition.

Check https://bugzilla.redhat.com/show_bug.cgi?id=1729603 for detail info.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -301,6 +301,13 @@ kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cm
 # SCP the vmlinuz/initramfs from VM to Host in provided folder.
 ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/rhcos-${ostree_hash}/* $1
 
+# Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1729603
+# TODO: Should be removed once latest podman available or the fix is backported.
+# Issue found in podman version 1.4.2-stable2 (podman-1.4.2-5.el8.x86_64)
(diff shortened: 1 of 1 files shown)


cfergeau:
> Podman and crio both using the CNI networking and that cause the
> race condition.
> 
> Check https://bugzilla.redhat.com/show_bug.cgi?id=1729603 for detail info.

If you are not logged into bugzilla, there are not any details in this bug, so I'd suggest adding a few more details to the log ("the problem is caused because X is started before Y, we want xx to have priority so we make sure it runs first", or something like this).

createdisk.sh
```diff
@@ -301,6 +301,12 @@ kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cm
 # SCP the vmlinuz/initramfs from VM to Host in provided folder.
 ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/rhcos-${ostree_hash}/* $1
 
+# Workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1729603
+# TODO: Should be removed once latest podman available or the fix is backported.
```
cfergeau: Maybe mention the rpm version of podman which we have now?

praveenkumar: @cfergeau done

createdisk.sh
```diff
@@ -301,6 +301,13 @@ kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cm
...
+# Issue found in podman version 1.4.2-stable2
```
cfergeau: Ah, by "rpm version", I was thinking something like `podman-1.4.2-4.el8` (`rpm -q podman` on the node)

Issue #102, opened 2019-09-25
Make registry component configurable in snc script.


praveenkumar:
Right now we are directly using the `quay.io/openshift-release-dev/ocp-release` as hard coded for getting the release information instead of that we should be making it configurable since nightly bits comes from `quay.io/openshift-release-dev/ocp-release-nightly` registry.

praveenkumar:
fixed by #103

PR #103, opened 2019-09-25
Issue #102 Get quay registry from environment variable if set

praveenkumar:
Right now, we are directly using the quay.io/openshift-release-dev/ocp-release
as hard coded for getting the release information instead of that we should
be making it configurable since nightly release bits comes from
quay.io/openshift-release-dev/ocp-release-nightly registry.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -10,6 +10,7 @@ YQ=${YQ:-yq}
 OPENSHIFT_INSTALL=${OPENSHIFT_INSTALL:-./openshift-install}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
+QUAY_REGISTRY=${QUAY_REGISTRY:-quay.io/openshift-release-dev/ocp-release}
 CRC_PV_DIR="/mnt/pv-data"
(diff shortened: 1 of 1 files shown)


praveenkumar:
@cfergeau I can just put that in the README about it but this whole thing going to run in the CI where we set what to build (nightly or released).

cfergeau:
> @cfergeau I can just put that in the README about it but this whole thing going to run in the CI where we set what to build (nightly or released).

So the script on its own becomes more and more dependants on arbitrary tweaks/options injected by the CI, I don't think we should go too far in that direction, it should be equally easy to generate an image on a developer machine and on the CI.

praveenkumar:
> So the script on its own becomes more and more dependants on arbitrary tweaks/options injected by the CI, I don't think we should go too far in that direction, it should be equally easy to generate an image on a developer machine and on the CI.

@cfergeau I agree to your concern, may be just expose a flag to ask which registry a user want in future but that would be after RC side.

cfergeau:
> > So the script on its own becomes more and more dependants on arbitrary tweaks/options injected by the CI, I don't think we should go too far in that direction, it should be equally easy to generate an image on a developer machine and on the CI.
> 
> @cfergeau I agree to your concern, may be just expose a flag to ask which registry a user want in future but that would be after RC side.

ok...

PR #113, opened 2019-10-10
Proper Bash check for missing value

cardil:
* Changing a check for OPENSHIFT_VERSION variable that works if that
variable isn't declared.
 * Changing a git describe to always produce a value.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -17,7 +17,7 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 # Otherwise use the tagged version if available
 function get_openshift_version {
-    if [ ${OPENSHIFT_VERSION} != "" ]; then
+    if [ "${OPENSHIFT_VERSION}" != "" ]; then
         OPENSHIFT_RELEASE_VERSION=$OPENSHIFT_VERSION
     else
         OPENSHIFT_RELEASE_VERSION=$(git describe --exact-match --tags HEAD 2>/dev/null)
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+id_rsa_crc
+id_rsa_crc.pub
+crc-tmp-install-data
+openshift-install


snc.sh
```diff
@@ -17,10 +17,10 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 # Otherwise use the tagged version if available
 function get_openshift_version {
-    if [ ${OPENSHIFT_VERSION} != "" ]; then
+    if [ "${OPENSHIFT_VERSION}X" != "X" ]; then
```
praveenkumar: @cardil why do you think `X` is required why not just compare `"${OPENSHIFT_VERSION}"` ?

snc.sh
```diff
@@ -17,10 +17,10 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
...
         OPENSHIFT_RELEASE_VERSION=$OPENSHIFT_VERSION
     else
-        OPENSHIFT_RELEASE_VERSION=$(git describe --exact-match --tags HEAD 2>/dev/null)
+        OPENSHIFT_RELEASE_VERSION=$(git describe --always --tags HEAD 2>/dev/null)
```
praveenkumar: @cardil this is by design since we don't want to run a tagged version of release so if we merge this then there will be no difference between the master branch or a tagged version.

```
$ git branch 
* master

 $ git describe --exact-match --tags HEAD
fatal: no tag exactly matches '2c2dc7d22758cbe39df1411965d0ed6f1b574165'

$ git checkout 4.1.14 <= or any tag/releases
$ git branch 
* (HEAD detached at 4.1.14)

$ git describe --exact-match --tags HEAD
4.1.14
```

snc.sh
cfergeau: According to https://stackoverflow.com/questions/6852612/test-for-empty-string-with-x 
`[ "X${OPENSHIFT_VERSION}" != "X" ]` is better

cardil: Shellcheck also raports this as invalid.

```bash
$ shellcheck snc.sh

In snc.sh line 19:
    if [ ${OPENSHIFT_VERSION} != "" ]; then
         ^-- SC2086: Double quote to prevent globbing and word splitting.
```

I propose this because without defining `OPENSHIFT_VERSION` script throws:

```
~/git/snc/snc.sh: 20: [: !=: unexpected operator
``` 

Shellcheck proposes to quote and I will do that. That also prevents an error without additional `X` sign, that might be mind bending.

snc.sh
cardil: It will be different:

```
$ git branch 
* master

 $ git describe --always --tags HEAD
4.1.14-8-g037212f

$ git checkout 4.1.14
$ git branch 
* (HEAD detached at 4.1.14)

$ git describe --always --tags HEAD
4.1.14
```

On tagged commit, it will be equal to tag. On other commits it gives a meaningful output.
The current head is based on `4.1.14`, but since it has a few commits on top of that, describe has added the number of additional commits `8` and an abbreviated object name for the commit itself `037212f` at the end.

In that way it's always unique and easy to know is that was official release of build of something else.

cardil: I removed it, as I didn't saw how this is used. Sorry.

snc.sh
```diff
@@ -17,7 +17,7 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_c
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 # Otherwise use the tagged version if available
 function get_openshift_version {
-    if [ ${OPENSHIFT_VERSION} != "" ]; then
+    if [ "${OPENSHIFT_VERSION}" != "" ]; then
```
praveenkumar: :+1:

praveenkumar:
@cardil `git log` has the typo for **works**` and **declared**

`Changing a check for OPENSHIFT_VERSION variable that wotrks if that
variable isn't decraled.`

cardil:
I've fixed the typo.

praveenkumar:
@cardil Still *declared* is not fixed in the commit log and it contains *decraled* :)

Issue #111, opened 2019-10-10
Disable insight operator


praveenkumar:
Right now with 4.2 `insight` operator is also added and we can disable it without causing any issue to the cluster.
```
$ oc get all -n openshift-insights
NAME                                     READY   STATUS    RESTARTS   AGE
pod/insights-operator-56c7c97d87-h2rdd   1/1     Running   0          2d23h

NAME                                READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/insights-operator   1/1     1            1           2d23h

NAME                                           DESIRED   CURRENT   READY   AGE
replicaset.apps/insights-operator-56c7c97d87   1         1         1       2d23h

```

praveenkumar:
Fixed by #114

PR #114, opened 2019-10-11
Issue #111 Scale replica to 0 for insight operator

praveenkumar:


gbraad:
Actually, this information is better to collect in the actual script, as in this case it would then be usable as a form of documentation and discussion artifact.

cfergeau:
> Actually, this information is better to collect in the actual script, as in this case it would then be usable as a form of documentation and discussion artifact.

True... this would indeed have been better :)

Issue #115, opened 2019-10-14
Enable PV for registry storage backend.


praveenkumar:
By default, installer use `EmptyDir` for registry storage and this is not suitable during stop=>start scenario where user have some images to imagestream. We already have 30 pv, one of them can be used for registry side.

```
$ oc get deployment image-registry -oyaml | grep -A3 -i volumes
  volumes:
  - emptyDir: {}
    name: registry-storage
```

praveenkumar:
Fixed by #116

praveenkumar:
Looks like we need to change the `configs.imageregistry.operator.openshift.io` as mentioned in the https://docs.openshift.com/container-platform/4.2/registry/configuring-registry-storage/configuring-registry-storage-baremetal.html documents.

praveenkumar:
Fixed by #121

PR #116, opened 2019-10-16
Issue #115 Enable PV for registry volume

praveenkumar:
Currently `EmptyDir{}` is used as volume for registry which doesn't
able to persist the data from start => stop case.

### DIFF
--- a/registry_pvc.yaml
+++ b/registry_pvc.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: registry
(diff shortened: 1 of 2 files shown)


praveenkumar:
> If they increase disk size manually, why can't they also increase the size of the PVs?

@cfergeau They can or even add more PV's but I don't see any harm to have it around `100Gi` as default since it is just spec.

praveenkumar:
Merging after having discussion on slack with @cfergeau ( He is still feel it is odd :wink: but not strongly disagree.)

Issue #117, opened 2019-10-17
Validate cert for server and client for expiration


praveenkumar:
It is possible that sometime client cert is rotated but server one is not or vice versa. We need to add check for both the certs.

```
$ ssh -i id_rsa_crc -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null core@192.168.126.11 -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem
Certificate will not expire
$ ssh -i id_rsa_crc -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null core@192.168.126.11 -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-server-current.pem
Certificate will expire
```

praveenkumar:
This is already fixed.

PR #120, opened 2019-10-22
Issue #117 Validate kubelet server cert also for expiration

praveenkumar:
It is possible that sometime client cert is rotated but server one is not or vice versa. We need to add check for both the certs.

```
$ ssh -i id_rsa_crc -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null core@192.168.126.11 -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem
Certificate will not expire
$ ssh -i id_rsa_crc -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null core@192.168.126.11 -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-server-current.pem
Certificate will expire
```

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -198,8 +198,13 @@ create_pvs "${CRC_PV_DIR}" 30
 # Once it is finished, disable the CVO
 ${OC} scale --replicas=0 deployment --all -n openshift-cluster-version
 
+# Get the pod name associated with cluster-monitoring-operator deployment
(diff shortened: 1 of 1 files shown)


anjannath:
@praveenkumar when we check this validity from crc we are only checking the client certs, in there also we should check both?

gbraad:
Good point ... has this caused issues so far?

PR #121, opened 2019-10-22
Issue #115 Add pvc as registry storage.

praveenkumar:
This revert the 7c468bde commit since it is not persist during restart.
To have it properly we need to make the change to configs.imageregistry.operator.openshift.io
resources.

- https://docs.openshift.com/container-platform/4.2/registry/configuring-registry-storage/configuring-registry-storage-baremetal.html

### DIFF
--- a/registry_pvc.yaml
+++ b/registry_pvc.yaml
@@ -1,11 +1,11 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: registry
+  name: crc-image-registry-storage	
   namespace: openshift-image-registry
 spec:
   accessModes:
-    - ReadWriteOnce
+    - ReadWriteMany
   resources:
     requests:
       storage: 100Gi
--- a/snc.sh
+++ b/snc.sh
@@ -237,6 +237,11 @@ ${OC} scale --replicas=0 deployment --all -n openshift-cloud-credential-operator
 # Apply registry pvc to bound with pv0001
 ${OC} apply -f registry_pvc.yaml
 
+# Add registry storage to pvc
+${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim": "crc-image-registry-storage"}}]' --type=json
+# Remove emptyDir as storage for registry
+${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "remove", "path": "/spec/storage/emptyDir"}]' --type=json
+
 # Delete the v1beta1.metrics.k8s.io apiservice since we are already scale down cluster wide monitioring.
 # Since this CRD block namespace deletion forever.
 ${OC} delete apiservice v1beta1.metrics.k8s.io


praveenkumar:
> What is being done in this commit does not match what is described in the documentation linked in the commit log?

@cfergeau I used the same, doc describe the way using `oc edit` which need manual intervention and I used the patch command to have it automated.

cfergeau:
> > What is being done in this commit does not match what is described in the documentation linked in the commit log?
> 
> @cfergeau I used the same, doc describe the way using `oc edit` which need manual intervention and I used the patch command to have it automated.

This is not obvious at all, the documentation says
> Leave the claim field blank to allow the automatic creation of an image-registry-storage PVC.

This commit sets the 'claim' field to 'image-registry-storage' rather than blanking it, and it deletes 'storage/emptyDir' which is not mentioned in the documentation.

praveenkumar:
> This commit sets the 'claim' field to 'image-registry-storage' rather than blanking it, and it deletes 'storage/emptyDir' which is not mentioned in the documentation.

@cfergeau when you left `claim` field as blank it auto add `image-registry-storage` which is only able to done when using the `oc edit` since `oc patch` command doesn't allow to have a null string and you will see something like.
```
$ oc patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim"}}]' --type=json
The Config "cluster" is invalid: []: Invalid value: map[string]interface {}{"apiVersion":"imageregistry.operator.openshift.io/v1", "kind":"Config", "metadata":map[string]interface {}{"creationTimestamp":"2019-10-16T10:22:02Z", "finalizers":[]interface {}{"imageregistry.operator.openshift.io/finalizer"}, "generation":26, "name":"cluster", "resourceVersion":"325300", "uid":"cb55fd5e-effe-11e9-82ff-525400b24bd6"}, "spec":map[string]interface {}{"defaultRoute":true, "httpSecret":"ceae8318acc54dc682e186e4be7213b7e576c6649366c646e76b5bef3bd8673f102c67425271b2386e85a52e3493372ed835f52a0e2993ec7066ca6076469ed6", "logging":2, "managementState":"Managed", "proxy":map[string]interface {}{"http":"", "https":"", "noProxy":""}, "readOnly":false, "replicas":1, "requests":map[string]interface {}{"read":map[string]interface {}{"maxInQueue":0, "maxRunning":0, "maxWaitInQueue":"0s"}, "write":map[string]interface {}{"maxInQueue":0, "maxRunning":0, "maxWaitInQueue":"0s"}}, "storage":map[string]interface {}{"pvc":map[string]interface {}{"claim":interface {}(nil)}}}, "status":map[string]interface {}{"conditions":[]interface {}{map[string]interface {}{"lastTransitionTime":"2019-10-22T11:42:57Z", "reason":"PVC Exists", "status":"True", "type":"StorageExists"}, map[string]interface {}{"lastTransitionTime":"2019-10-16T10:26:08Z", "message":"The registry is ready", "reason":"Ready", "status":"True", "type":"Available"}, map[string]interface {}{"lastTransitionTime":"2019-10-22T14:14:52Z", "message":"The registry is ready", "reason":"Ready", "status":"False", "type":"Progressing"}, map[string]interface {}{"lastTransitionTime":"2019-10-16T10:22:06Z", "status":"False", "type":"Degraded"}, map[string]interface {}{"lastTransitionTime":"2019-10-16T10:22:06Z", "status":"False", "type":"Removed"}}, "observedGeneration":25, "readyReplicas":0, "storage":map[string]interface {}{"emptyDir":map[string]interface {}{}, "pvc":map[string]interface {}{"claim":"image-registry-storage"}}, "storageManaged":true}}: validation failure list:
spec.storage.pvc.claim in body must be of type string: "null"
bash-3.2$ oc patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim":""}}]' --type=json
config.imageregistry.operator.openshift.io/cluster patched

$ oc patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/emptyDir", "value": {}}]' --type=json
config.imageregistry.operator.openshift.io/cluster patched

$ oc get configs.imageregistry.operator.openshift.io -o json | jq .items[0].spec.storage
{
  "pvc": {
    "claim": ""
  }
}
```

cfergeau:
The relevant code is https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/pvc/pvc.go#L145
If the name is empty, the operator creates a pvc with the name 'image-registry-storage', and then sets this name as the value for 'claim'.
I'm not sure we are doing that claim at all? If we do the claim manually, I think we should use a different name for it ('crc-image-registry-storage'), and set that name in the claim.

Issue #27, opened 2019-05-20
Force certificate rotation after creating the cluster


cfergeau:
We currently need to wait 24h after creating the cluster before getting certificates valid for one month. We should have a way of forcing the rotation of the certificates without waiting for 24h

Experimenting with the commands below seemed to help:
```
# validity is 30 times the base
oc create -n openshift-config configmap unsupported-cert-rotation-config --from-literal='base=30s'

# forcing rotation
oc get secret -A -o json | jq -r '.items[] | select(.metadata.annotations."auth.openshift.io/certificate-not-after" | .!=null and fromdateiso8601<='$( date --date='+1year' +%s )') | "-n \(.metadata.namespace) \(.metadata.name)"' | xargs -n3 oc patch secret -p='{"metadata": {"annotations": {"auth.openshift.io/certificate-not-after": null}}}'

# Wait ~ 5-10 minutes

# Make sure at least the apiserver serving cert has 15 min validity (change your cluster name based on your kubeconfig)
openssl s_client -connect api.tnozicka-1.devcluster.openshift.com:6443 | openssl x509 -noout -dates


# go back to normal certrotation setting
oc delete -n openshift-config configmap unsupported-cert-rotation-config
```

praveenkumar:
I am all up for it if the resulted cluster is healthy, Conversation which I had sometime back with installer folks look like even when initial `oc get co` says required operator are in healthy state then also it need to give some time to reconcile the cluster before putting a workload, I am not sure how quickly that reconcile can happen when we are putting `~30sec` limit to the cert. Looks like we need to test it before make it part of our bundle creation script.

anjannath:
After doing the above steps, how to check if the certificates were rotated?.. i waited around 30mins and checked the certificates in the master.. and found the same 1 day validity
```
[root@crc-l8gfp-master-0 pki]# openssl x509 -in kubelet-client-current.pem -noout -dates 
notBefore=May 28 08:05:00 2019 GMT
notAfter=May 29 07:48:17 2019 GMT 

$ ./oc get csr
NAME AGE REQUESTOR CONDITION
csr-4mp5f 113m system:node:crc-l8gfp-master-0 Approved,Issued
csr-b5jnq 113m system:serviceaccount:openshift-machine-config-operator:node-bootstrapper Approved,Issued 
```

anjannath:
[...]

PR #124, opened 2019-11-05
Issue #27 Force cert rotation after creating the cluster

praveenkumar:
As of now, we need to wait for 24 hour till the kubelet cert rotation
happen, with help of bootstarp cred manager daemon set, it can be rotated
just after the cluster created and new bootstarp csr generated which will
create 30 days valid certs for the kublet client and server.

- https://blog.openshift.com/enabling-openshift-4-clusters-to-stop-and-resume-cluster-vms/

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -169,6 +169,28 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
 
 ${OPENSHIFT_INSTALL} --dir $INSTALL_DIR create cluster --log-level debug
 
+export KUBECONFIG=$INSTALL_DIR/auth/kubeconfig
+
+oc apply -f kubelet-bootstrap-cred-manager-ds.yaml
+
+# Delete the current csr signer to get new request.
+oc delete secrets/csr-signer-signer secrets/csr-signer -n openshift-kube-controller-manager-operator
+
+# Wait for 5 min to make sure cluster is stable again.
+sleep 300
+
+# Remove the 24 hours certs and bootstrap kubeconfig
+# this kubeconfig will be regenerated and new certs will be created in pki folder
+# which will have 30 days validity.
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo rm -fr /var/lib/kubelet/pki
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo rm -fr /var/lib/kubelet/kubeconfig
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl restart kubelet
+
+# Wait until bootstrap csr request is generated.
+until oc get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done
+oc get csr -oname | xargs oc adm certificate approve
+
+
 # Wait for install to complete, this provide another 30 mins to make resources (apis) stable
 ${OPENSHIFT_INSTALL} --dir $INSTALL_DIR wait-for install-complete --log-level debug
 if [ $? -ne 0 ]; then
@@ -183,8 +205,6 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} sudo hostnamectl set-hostname ${HO
 
 create_json_description
 
-# export the kubeconfig
-export KUBECONFIG=$INSTALL_DIR/auth/kubeconfig
 
 # Create persistent volumes
 create_pvs "${CRC_PV_DIR}" 30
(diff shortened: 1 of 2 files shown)


snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
 
 ${OPENSHIFT_INSTALL} --dir $INSTALL_DIR create cluster --log-level debug
 
+# export the kubeconfig
+export KUBECONFIG=$INSTALL_DIR/auth/kubeconfig
+
+# Add kubelet-bootstrap-cred-manager daemonset
+oc apply -f kubelet-bootstrap-cred-manager-ds.yaml
+
+# Delete the current csr signer to get new request.
```
cfergeau: There's an extra blank space at the end of this line

snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
 
 ${OPENSHIFT_INSTALL} --dir $INSTALL_DIR create cluster --log-level debug
 
+# export the kubeconfig
```
cfergeau: I don't think this comment is useful, the next line says as much

snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
...
+export KUBECONFIG=$INSTALL_DIR/auth/kubeconfig
+
+# Add kubelet-bootstrap-cred-manager daemonset
```
cfergeau: I would remove that comment too

snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
...
+# which will have 30 days validity.
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo rm -fr /var/lib/kubelet/pki
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo rm -fr /var/lib/kubelet/kubeconfig
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl restart kubelet
+
+# Wait until bootstrap csr request is generated.
+until oc get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done
+for CSR in $(oc get csr | grep Pending | awk '{print $1}'); do
+       oc adm certificate approve $CSR;
+done
```
cfergeau: The instructions you link to have an alternative to this command which involves less parsing:
`oc get csr -oname | xargs oc adm certificate approve`

snc.sh
```diff
@@ -239,3 +262,12 @@ ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op":
...
 # Since this CRD block namespace deletion forever.
 ${OC} delete apiservice v1beta1.metrics.k8s.io
+
+# Get the cert pod name
+cert_pod=$(${OC} get pod -l k8s-app=kubelet-bootstrap-cred-manager -o jsonpath="{.items[0].metadata.name}" -n openshift-machine-config-operator)
+# Remove the bootstrap-cred-manager daemonset
+${OC} delete daemonset.apps/kubelet-bootstrap-cred-manager -n openshift-machine-config-operator
+# Wait till the cert pod is removed
+${OC} wait --for=delete pod/$cert_pod --timeout=60s -n openshift-machine-config-operator
+# Remove the cli image which used for the bootstrap-cred-manager daemonset
```
cfergeau: ....which _was_ used for ...

snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
...
+oc apply -f kubelet-bootstrap-cred-manager-ds.yaml
+
+# Delete the current csr signer to get new request. 
+oc delete secrets/csr-signer-signer secrets/csr-signer -n openshift-kube-controller-manager-operator
+
+# Wait for 5 min to make sure cluster is stable again.
+sleep 300
+
+# Remove the 24 hours certs and bootstrap kubeconfig
+# This kubeconfig again get generated and new certs will be created in pki folder
```
cfergeau: "this kubeconfig will be regenerated" instead of "This kubeconfig again get generated "?

snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
...
+export KUBECONFIG=$INSTALL_DIR/auth/kubeconfig
+
+# Add kubelet-bootstrap-cred-manager daemonset
+oc apply -f kubelet-bootstrap-cred-manager-ds.yaml
+
+# Delete the current csr signer to get new request. 
+oc delete secrets/csr-signer-signer secrets/csr-signer -n openshift-kube-controller-manager-operator
+
+# Wait for 5 min to make sure cluster is stable again.
+sleep 300
```
cfergeau: do you think `oc wait --for=condition=Ready pod/xxx` could be used here rather than sleep?

snc.sh
```diff
@@ -169,6 +169,31 @@ export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
...
+
+# Remove the 24 hours certs and bootstrap kubeconfig
+# This kubeconfig again get generated and new certs will be created in pki folder
+# which will have 30 days validity.
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo rm -fr /var/lib/kubelet/pki
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo rm -fr /var/lib/kubelet/kubeconfig
```
cfergeau: Removing these does not seem to be part of the instructions you linked to? https://blog.openshift.com/enabling-openshift-4-clusters-to-stop-and-resume-cluster-vms/
Is this mandatory?

snc.sh
praveenkumar: @cfergeau no, wait is good for a single pod which we know about, here we don't know the which all the pods are going to terminated and come back again so need to put a sleep.

snc.sh
praveenkumar: @cfergeau Without this we need to reboot the cluster to get the new csr atleast the node one so from crc prospective this is mandatory.

cfergeau: It seems unexpected though https://bugzilla.redhat.com/show_bug.cgi?id=1760181#c4
Maybe add a comment that this is a workaround?

snc.sh
cfergeau: Checking that a pod which starts very late is available might be a good addition to the sleep so that we are sure we waited long enough?

praveenkumar: @cfergeau might be, but after the 2-3 mins wait we already check with `openshift-install --wait-for ..` which is more accurate then we are trying to go with different pods and it schedule.

praveenkumar:
> Maybe it would be worth having separate functions for the cert regeneration/cert regeneration cleanup steps, so that they are easily identified in the overall code flow?

Sure, might be refactor at later stage.

Issue #129, opened 2019-11-13
Add workaround for virt-sparsify for 4.3 images


praveenkumar:
More details: https://bugzilla.redhat.com/show_bug.cgi?id=1771257

praveenkumar:
fixed by #130

PR #130, opened 2019-11-13
Issue #129 Workaround for virt-sparsify to work with immutable partition

praveenkumar:
With OpenShift-4.3 the rh-coreos image failed during sparsify because of
the immutable attribute set for /dev/sda4 partition. Need to use guestfish
to remove it and run the virt-sparsify and then apply back.

- https://bugzilla.redhat.com/show_bug.cgi?id=1771257

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -61,12 +61,18 @@ function create_qemu_image {
             exit 1
     fi
 
+    # Remove immutable attribute for OpenShift 4.3+
+    guestfish -a $destDir/${CRC_VM_NAME}.qcow2 -m /dev/sda4 set-e2attrs / i clear:true
+
     # TMPDIR must point at a directory with as much free space as the size of the image we want to sparsify
     # Read limitation section of `man virt-sparsify`.
     TMPDIR=$(pwd)/$destDir ${VIRT_SPARSIFY} -o lazy_refcounts=on $destDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}_sparse.qcow2
     rm -f $destDir/${CRC_VM_NAME}.qcow2
     mv $destDir/${CRC_VM_NAME}_sparse.qcow2 $destDir/${CRC_VM_NAME}.qcow2
     rm -fr $destDir/.guestfs-*
+
+    # Set immutable attribute back for /dev/sda4
+    guestfish -a $destDir/${CRC_VM_NAME}.qcow2 -m /dev/sda4 set-e2attrs / i
 
     # Before using the created qcow2, check if it has lazy_refcounts set to true.
     ${QEMU_IMG} info ${destDir}/${CRC_VM_NAME}.qcow2 | grep "lazy refcounts: true" 2>&1 >/dev/null
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -50,9 +50,12 @@ function create_qemu_image {
     ${QEMU_IMG} rebase -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
     ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
 
+    # Check which partition is labeled as `root`
+    partition=$(${VIRT_FILESYSTEMS} -a $destDir/${VM_PREFIX}-base  -l | grep root | cut -f1 -d' ')
+
     # Resize the image from the default 1+15GB to 1+30GB
     ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 31G
-    ${VIRT_RESIZE} --expand /dev/sda3 $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
+    ${VIRT_RESIZE} --expand $partition $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
     if [ $? -ne 0 ]; then
             echo "${VIRT_RESIZE} call failed, disk image was not properly resized, aborting"
             exit 1
@@ -192,6 +195,7 @@ JQ=${JQ:-jq}
 VIRT_RESIZE=${VIRT_RESIZE:-virt-resize}
 QEMU_IMG=${QEMU_IMG:-qemu-img}
 VIRT_SPARSIFY=${VIRT_SPARSIFY:-virt-sparsify}
+VIRT_FILESYSTEMS=${VIRT_FILESYSTEMS:-virt-filesystems}
 
 if [[ $# -ne 1 ]]; then
    echo "You need to provide the running cluster directory to copy kubeconfig"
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -58,13 +58,19 @@ function create_qemu_image {
...
     # TMPDIR must point at a directory with as much free space as the size of the image we want to sparsify
     # Read limitation section of `man virt-sparsify`.
     TMPDIR=$(pwd)/$destDir ${VIRT_SPARSIFY} -o lazy_refcounts=on $destDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}_sparse.qcow2
     rm -f $destDir/${CRC_VM_NAME}.qcow2
     mv $destDir/${CRC_VM_NAME}_sparse.qcow2 $destDir/${CRC_VM_NAME}.qcow2
     rm -fr $destDir/.guestfs-*
 
+    # Set immutable attribute back for /dev/sda4
+    guestfish -a $destDir/${CRC_VM_NAME}.qcow2 -m /dev/sda4 set-e2attrs / i
+
```
cfergeau: Won't that cause issues with 4.2 which does not set this bit?

praveenkumar: @cfergeau I didn't test that but it shouldn't if we just want to make that partition as immutable, Need to test this out and if this is causing issue we might just create another branch for 4.2 side.

cfergeau: On 4.2 the partition is not immutable, so I don't know how coreos is going to behave if we suddenly make it immutable. We could check whether the partition is immutable or not before making these changes.

praveenkumar: @cfergeau on 4.2 side there is no `/dev/sda4` :(

```
$ guestfish -a crc.qcow2 -m /dev/sda4 set-e2attrs / i
libguestfs: error: vfs_type: vfs_type_stub: /dev/sda4: No such file or directory
libguestfs: error: mount_options: mount_options_stub: /dev/sda4: No such file or directory
guestfish: ‘/dev/sda4’ could not be mounted.
guestfish: Did you mean to mount one of these filesystems?
guestfish: 	/dev/sda1 (unknown)
guestfish: 	/dev/sda2 (ext4)
guestfish: 	/dev/sda3 (xfs)
```

praveenkumar: > On 4.2 the partition is not immutable, so I don't know how coreos is going to behave if we suddenly make it immutable. We could check whether the partition is immutable or not before making these changes.

@cfergeau yes, that would be a better solution in terms of checks. Using the guestfish we can get if there is immutable bit set and then only perform removing that bit and adding it but as I said for the 4.2 side there is no `/dev/sda4`.

```
$ guestfish -a crc.qcow2 -m /dev/sda4 get-e2attrs /
i
```

praveenkumar: @cfergeau 

For 4.3 images
```
guestfish <<_EOF_
 add crc.qcow2
run
list-filesystems
_EOF_
/dev/sda1: ext4
/dev/sda2: vfat
/dev/sda3: unknown
/dev/sda4: xfs
```

For 4.2 images
```
$  guestfish <<_EOF_
>  add crc.qcow2
> run
> list-filesystems
> _EOF_
/dev/sda1: unknown
/dev/sda2: ext4
/dev/sda3: xfs
```

cfergeau: Oh well, if no /dev/sda4 on 4.2, this indeed means my concern is not going to be an issue. We could leave it as it is and ignore the error.
Looks like we should double check we resize the correct partition on 4.3?

praveenkumar: > Looks like we should double check we resize the correct partition on 4.3?

Yes now it should be `/dev/sda4` which we need to resize `/dev/sda3` is the boot partition so I think first we detect which partition is `xfs` and then increase it size.

```
$ sudo fdisk -l
Disk /dev/vda: 31 GiB, 33285996544 bytes, 65011712 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: 00000000-0000-4000-A000-000000000001

Device        Start      End  Sectors  Size Type
/dev/vda1      2048   788479   786432  384M Linux filesystem
/dev/vda2    788480  1048575   260096  127M EFI System
/dev/vda3   1048576 32505215 31456640   15G BIOS boot
/dev/vda4  32505216 65008990 32503775 15.5G Linux filesystem
```

createdisk.sh
```diff
@@ -206,6 +210,10 @@ if ! which ${VIRT_RESIZE}; then
     sudo yum -y install /usr/bin/virt-resize libguestfs-xfs
 fi
 
+if ! which ${VIRT_FILESYSTEMS}; then
+    sudo yum -y install /usr/bin/virt-filesystems
+fi
+
```
cfergeau: It's in the same package as virt-resize, so most of the time this will be skipped.

createdisk.sh
```diff
@@ -50,9 +50,12 @@ function create_qemu_image {
     ${QEMU_IMG} rebase -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
     ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
 
+    # Check which partition is labeled as `root`
+    partition=$(${VIRT_FILESYSTEMS} -a $destDir/${VM_PREFIX}-base  -l | grep root | cut -f1 -d' ')
```
cfergeau: `awk` could be used instead of `cut`, both are fine with me.

createdisk.sh
```diff
@@ -50,9 +50,12 @@ function create_qemu_image {
...
+
     # Resize the image from the default 1+15GB to 1+30GB
     ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 31G
-    ${VIRT_RESIZE} --expand /dev/sda3 $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
+    ${VIRT_RESIZE} --expand $partition $destDir/${VM_PREFIX}-base $destDir/${CRC_VM_NAME}.qcow2
```
cfergeau: The other commit fixes an issue where /dev/sda4 is immutable with openshift 4.3. Can we write to that partition after the cluster is setup? Otherwise, it's not really useful to grow it.

createdisk.sh
```diff
@@ -61,13 +61,19 @@ function create_qemu_image {
             exit 1
     fi
 
+    # Remove immutable attribute for /dev/sda4
```
cfergeau: maybe replace /dev/sda4 with "OpenShift 4.3+"

createdisk.sh
cfergeau: Misunderstanding on my side of how the 'i' attribute works, it's only set on the 'root' directory of the /dev/sda4 partition, so we cannot write to that dir. However, it's not recursive, so this does not impact any of its subdirectories which do not have that bit set. So this is not going to be an issue.

Issue #131, opened 2019-11-22
Cache the recovery image so can be used later.


praveenkumar:
We need the recovery image to be cached as part of our snc and that we can use for cert recovery side.

- https://gist.github.com/cfergeau/fc4ac5379a120dad1381e8ba1fc4789d

praveenkumar:
Fixed by #132

PR #132, opened 2019-11-22
Issue #131 Tag cluster-kube-apiserver-operator image

praveenkumar:
We need the image to recover the cert in case it expire, if we don't
tag it properly then we need to do it on the CRC side and that require
the internet connectivity to fetch the hash for KAO image. Now we can
directly use `openshift/cert-recovery` tag without fetching it from the
release payload.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -255,6 +255,10 @@ ${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-lit
 ${OC} --config $1/auth/kubeconfig apply -f htpasswd_cr.yaml
 ${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterrole=sudoer --user=developer
 
+# Get cluster-kube-apiserver-operator image along with hash and tag it
+certImage=$(${OC} --config $1/auth/kubeconfig adm release info --image-for=cluster-kube-apiserver-operator)
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman tag $certImage openshift/cert-recovery
+
 # Replace pull secret with a null json string '{}'
 ${OC} --config $1/auth/kubeconfig replace -f pull-secret.yaml
 


createdisk.sh
```diff
@@ -255,6 +255,10 @@ ${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-lit
 ${OC} --config $1/auth/kubeconfig apply -f htpasswd_cr.yaml
 ${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterrole=sudoer --user=developer
 
+# Get cluster-kube-apiserver-operator image along with hash and tag it
+certImage=$(${OC} --config $1/auth/kubeconfig adm release info --image-for=cluster-kube-apiserver-operator)
```
cfergeau: Is this going to use `quay.io/openshift-release-dev/ocp-release:4.2.2` by default? (with '4.2.2' being the openshift release we are building the bundle for)
The command I was using so far was
```
${OC} --config $1/auth/kubeconfig adm release info  quay.io/openshift-release-dev/ocp-release:4.2.2 --image-for=cluster-kube-apiserver-operator
```

createdisk.sh
```diff
@@ -255,6 +255,10 @@ ${OC} --config $1/auth/kubeconfig create secret generic htpass-secret --from-lit
...
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman tag $certImage openshift/cert-recovery
```
cfergeau: When I test  this on a CRC image, the tagging goes well, however the tagging does not seem to pull the needed bits to ensure they are available locally? If I run 'pull' after 'tag', I get
```
Trying to pull quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:f955cd757f3a94daca69e3bc9d5f0659c270b411ce27478b8dfa479ca7b1c191...Getting image source signatures                                           
Copying blob 6c6726ab35bd skipped: already exists
Copying blob 1d2c4ce43b78 skipped: already exists
Copying blob 1c9f515fc6ab skipped: already exists
Copying blob c7de79b2356b skipped: already exists
Copying blob dcb1864555fe skipped: already exists
Copying config fff5ef4ac6 done
Writing manifest to image destination
Storing signatures
```
Does this mean it fetched the config from the net?

createdisk.sh
praveenkumar: @cfergeau it is going to use whatever version we deployed on the cluster so for each release it should be it's own version since this is available from 4.2.0 so 4.2.2/4.2.4. ..etc. all have it's own `cluster-kube-apiserver-operator` image.

createdisk.sh
praveenkumar: @cfergeau when you get the image hash and search it with available images using `sudo podman images | grep <hash>` you can see that the image is already available. Can you directly try that image with tag, I am hoping that it doesn't have any bit extra but I have to make sure otherwise I will add `${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull $certImage` before it.

Issue #134, opened 2019-11-28
Update the logic to get right partition to increase the size


praveenkumar:
Since Red Hat CoreOS now moved with encryption so our `virt-filesystem` doesn't list those partitions, we need to make changes in that logic.

- https://github.com/coreos/coreos-assembler/pull/826
- https://github.com/openshift/enhancements/blob/master/enhancements/automated-policy-based-disk-encryption.md
- https://github.com/coreos/coreos-assembler/pull/826/files

praveenkumar:
resolved by #135

PR #135, opened 2019-11-28
Issue #134 Fix partition logic for encrypted partitions

praveenkumar:
For an encrypted disk, virt-filesystem doesn't show all the partition
until specified. Also now there is no `root` label on partition so this logic
need to be changed.
```
$ virt-filesystems -a crc-ssnp7-base -l
Name       Type        VFS   Label       Size       Parent
/dev/sda1  filesystem  ext4  boot        402653184  -
/dev/sda2  filesystem  vfat  EFI-SYSTEM  133169152  -

$ virt-filesystems -a crc-ssnp7-base -l --partitions
Name       Type       MBR  Size         Parent
/dev/sda1  partition  -    402653184    /dev/sda
/dev/sda2  partition  -    133169152    /dev/sda
/dev/sda3  partition  -    1048576      /dev/sda
/dev/sda4  partition  -    16641932800  /dev/sda
```

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -51,7 +51,7 @@ function create_qemu_image {
     ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
 
     # Check which partition is labeled as `root`
-    partition=$(${VIRT_FILESYSTEMS} -a $destDir/${VM_PREFIX}-base  -l | grep root | cut -f1 -d' ')
+    partition=$(${VIRT_FILESYSTEMS} -a $destDir/${VM_PREFIX}-base -l --partitions | sort -rk4 -n | sed -n 1p | cut -f1 -d' ')
 
     # Resize the image from the default 1+15GB to 1+30GB
     ${QEMU_IMG} create -o lazy_refcounts=on -f qcow2 $destDir/${CRC_VM_NAME}.qcow2 31G
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -61,18 +61,7 @@ function create_qemu_image {
             exit 1
     fi
 
-    # Remove immutable attribute for OpenShift 4.3+
-    guestfish -a $destDir/${CRC_VM_NAME}.qcow2 -m /dev/sda4 set-e2attrs / i clear:true
-
-    # TMPDIR must point at a directory with as much free space as the size of the image we want to sparsify
-    # Read limitation section of `man virt-sparsify`.
-    TMPDIR=$(pwd)/$destDir ${VIRT_SPARSIFY} -o lazy_refcounts=on $destDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}_sparse.qcow2
-    rm -f $destDir/${CRC_VM_NAME}.qcow2
-    mv $destDir/${CRC_VM_NAME}_sparse.qcow2 $destDir/${CRC_VM_NAME}.qcow2
     rm -fr $destDir/.guestfs-*
-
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -51,7 +51,7 @@ function create_qemu_image {
     ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
 
     # Check which partition is labeled as `root`
-    partition=$(${VIRT_FILESYSTEMS} -a $destDir/${VM_PREFIX}-base  -l | grep root | cut -f1 -d' ')
+    partition=$(${VIRT_FILESYSTEMS} -a $destDir/${VM_PREFIX}-base -l --partitions | sort -rk4 -n | sed -n 1p | cut -f1 -d' ')
```
cfergeau: the ` sed -n 1p | cut -f1 -d` can alternatively be changed to `awk '{ if (NR == 1)  {print $1}  }'`. I'd at least use `head -1` instead of `sed -n 1p`

praveenkumar:
> What is the impact on overall bundle size?

@cfergeau no idea, we will get to know once we create one using these changes.

createdisk.sh
praveenkumar: >  I'd at least use head -1 instead of sed -n 1p

@cfergeau Is there any reason `head -1` better than `sed -n 1p` ?

praveenkumar:
> The commit log says "Also now there is no root label on partition so this logic need to be changed.", which is not entirely true, we cannot read the label since the partition is encrypted, so we don't know whether it's root or something else

@cfergeau I will remove that comment but I tried with `--keys-from-stdin` and it also doesn't show.

```
$ echo nokey | virt-filesystems --keys-from-stdin -a crc-ssnp7-base -l --extra
Name       Type        VFS      Label       Size       Parent
/dev/sda1  filesystem  ext4     boot        402653184  -
/dev/sda2  filesystem  vfat     EFI-SYSTEM  133169152  -
/dev/sda3  filesystem  unknown  -           1048576    -
```

createdisk.sh
cfergeau: only (bad) reason is that I know what `head -1` is doing without looking at the man page, but I'm not familiar at all with `sed -n 1p`. No idea if it's just me, or if this is going to be typical :)

praveenkumar: @cfergeau I will not take as a reason then, you now know a new way to perform same thing 😉

createdisk.sh
```diff
@@ -61,19 +61,8 @@ function create_qemu_image {
...
 
-    # Remove immutable attribute for OpenShift 4.3+
-    guestfish -a $destDir/${CRC_VM_NAME}.qcow2 -m /dev/sda4 set-e2attrs / i clear:true
-
-    # TMPDIR must point at a directory with as much free space as the size of the image we want to sparsify
-    # Read limitation section of `man virt-sparsify`.
-    TMPDIR=$(pwd)/$destDir ${VIRT_SPARSIFY} -o lazy_refcounts=on $destDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}_sparse.qcow2
-    rm -f $destDir/${CRC_VM_NAME}.qcow2
-    mv $destDir/${CRC_VM_NAME}_sparse.qcow2 $destDir/${CRC_VM_NAME}.qcow2
     rm -fr $destDir/.guestfs-*
```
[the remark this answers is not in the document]
praveenkumar: Right, I still see the folder, didn't check the size so better to remove it.

Issue #136, opened 2019-11-29
Remove virtualbox bundle creation from script


praveenkumar:
Ref: https://github.com/code-ready/crc/issues/838

praveenkumar:
Fixed #138

PR #138, opened 2019-12-03
Issue #136 Stop generating virtualbox bundle

praveenkumar:
As part of CRC we disable some of operators like (monitoring/machine-config ..etc.) but
the images are always present on the node since there is no knob on installer side right now
to disable those operators from starting (iirc.) Now overall disk size is increases,
till 4.2 our final disk size was around 2GB but with 4.3 it is increasing around 3GB
and that is because all the other images are added as part of CVO payload.

crictl version used in the RHCOS is `0.1.0` which doesn't have fix
for kubernetes-sigs/cri-tools#399 one yet so
using kubernetes-sigs/cri-tools#399 (comment) as workaround.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -114,29 +114,6 @@ function copy_additional_files {
     update_json_description $srcDir $destDir
 }
 
-function generate_vbox_directory {
-    local srcDir=$1
-    local destDir=$2
-
-    cp $srcDir/kubeadmin-password $destDir/
-    cp $srcDir/kubeconfig $destDir/
-    cp $srcDir/id_rsa_crc $destDir/
-
-    ${QEMU_IMG} convert -f qcow2 -O vmdk $srcDir/${CRC_VM_NAME}.qcow2 $destDir/${CRC_VM_NAME}.vmdk
-    
-    diskSize=$(du -b $destDir/${CRC_VM_NAME}.vmdk | awk '{print $1}')
-    diskSha256Sum=$(sha256sum $destDir/${CRC_VM_NAME}.vmdk | awk '{print $1}')
-
-    cat $srcDir/crc-bundle-info.json \
-        | ${JQ} ".nodes[0].diskImage = \"${CRC_VM_NAME}.vmdk\"" \
-        | ${JQ} ".storage.diskImages[0].name = \"${CRC_VM_NAME}.vmdk\"" \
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -247,6 +247,9 @@ ${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterr
 certImage=$(${OC} --config $1/auth/kubeconfig adm release info --image-for=cluster-kube-apiserver-operator)
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman tag $certImage openshift/cert-recovery
 
+# Remove unused images from container stroage
```
cfergeau: 'storage'

createdisk.sh
```diff
@@ -247,6 +247,9 @@ ${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterr
...
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl images -q | xargs -n 1 sudo crictl rmi 2>/dev/null'
```
cfergeau: I would run this after the `systemctl stop kubelet` to avoid the (probably only theoritical for us) issue mentioned here https://github.com/kubernetes-sigs/cri-tools/issues/399#issuecomment-435562365
It probably would not hurt to explain in the commit log or in a comment that this is attempting to remove *all* images, but that crictl will not fail to remove images which are being used.

praveenkumar: @cfergeau If we run it after stopping the kubelet then it will remove the kubelet image also since as workaround it is just removing the images which are not used by any containers. Also `kubelet would pull an image before creating the container. If the prune happens during them, the container creation may be failed.` it mention that corner case happen if we are pulling the image during pruning the images and our case that is not true.

cfergeau: > If we run it after stopping the kubelet then it will remove the kubelet image 

oh ok, we indeed don't want that :)

> it mention that corner case happen if we are pulling the image during pruning the images and our case that is not true.

we are not doing it, but some operator could theoritically decide to do that in the background for whatever reason. I agree this is most likely not going to happen, and since we cannot easily avoid it, this should be good enough for now.

praveenkumar:
@cfergeau since after removing the images doesn't make a difference on final tarball so I am not going to add this patch and rebase it with only virtualbox removal code.

Issue #139, opened 2019-12-11
Add a wait around 2-3 seconds in between `crictl stopp` and `crictl rmp` command


praveenkumar:
Right now during the disk creation  `crictl rmp` run just after `crictl stopp` which fails if the containers are not stopped properly. So we need some wait time to gracefully stop those pods and may be a retry.

praveenkumar:
Fixed by #140

PR #140, opened 2019-12-11
Issue #139 Add a retry and wait time for `crictl rmp`

praveenkumar:
Right now during the disk creation `crictl rmp` run just after `crictl stopp`
which fails with following errors if the containers are not stopped properly.

```
$ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc core@api.crc.testing -- 'sudo crictl rmp $(sudo crictl pods -q)'
time="2019-12-11T05:13:15Z" level=fatal msg="removing the pod sandbox \"8412e11d8371c2d12cc2866db64f5b4678c3a31980d4008c2e0fbeafb1c86b47\" failed: rpc error: code = Unknown desc = failed to remove pod sandbox 8412e11d8371c2d12cc2866db64f5b4678c3a31980d4008c2e0fbeafb1c86b47: unlinkat /var/run/containers/storage/overlay-containers/8412e11d8371c2d12cc2866db64f5b4678c3a31980d4008c2e0fbeafb1c86b47/userdata/shm: device or resource busy"
```

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -239,7 +239,7 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
 # Remove all the pods from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crictl pods -q)'
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q)'
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'for i in {1..3}; do sudo crictl rmp $(sudo crictl pods -q) && break || sleep 2; done'
 
 # Remove pull secret from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'


createdisk.sh
```diff
@@ -239,7 +239,7 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
 # Remove all the pods from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crictl pods -q)'
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q)'
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'for i in {1..3}; do sudo crictl rmp $(sudo crictl pods -q) && break || sleep 2; done'
```
cfergeau: Something checking that `crictl ps` no longer has running pods would be more robust than an arbitrary loop+delay

cfergeau: Actually I'm not able to reproduce this, so I don't know if this suggestion would work out

praveenkumar: > Something checking that crictl ps no longer lists running pods would be more robust than an arbitrary loop+delay

@cfergeau This is kind of awkward situation that even `crictl ps` doesn't show the pods since those are going to be in stop state but not properly stopped, It is kind of hard to produce it but on the CI it does happen time to time. On my mac I am also not able to reproduce it.

> Actually I'm not able to reproduce this, so I don't know if this suggestion would work out

The way I tried it out is keep the `kubelet` running so that all the pods are not stopped correctly since kubelet try to start them again, during that you can see this command is tried 3 times.

cfergeau: > > Actually I'm not able to reproduce this, so I don't know if this suggestion would work out
> 
> The way I tried it out is keep the `kubelet` running so that all the pods are not stopped correctly since kubelet try to start them again, during that you can see this command is tried 3 times.

Testing this while kubelet is running is going to be racy, as we run `$(sudo crictl pods -q)` at 2 different times, so the pods listed there are going to be different.
Maybe ```
pods=$(sudo crictl pods -q)
sudo crictl stopp $pods
sudo crictl rmp $pods
```
would work better?

praveenkumar: @cfergeau I am pointing running kubelet just for test since we are not able to reproduce it but it actually happen on the CI times multiple time. Getting the pods and stop/delete doesn't different that what we have right now, what we need is retry option.

cfergeau: Not clear whether this is a workaround for a crictl bug, or something which is expected though :-/

PR #142, opened 2020-01-16
Manually sparsify luks partitions

cfergeau:
OpenShift 4.3 moved their root partition to a LUKS volume so that they
can reencrypt it more easily. One problem this causes is that
virt-sparsify/virt-resize do not deal very well with such partitions
(see 782ace66 "Remove virt-sparsify because of disk encryption" for
example).
This has a bad side-effect: the resulting bundle ends up being more than
4GB, which is far too big.
Another smaller issue is that virt-resize does not call xfs-growfs after
having grown the FS so the root filesystem is not actually bigger.

This commit shrinks the bundle to about 2.3GB by doing manually the same
steps as virt-sparsify after mounting the luks partition. It also calls
xfs-growfs to avoid the issue mentioned above.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -35,6 +35,59 @@ function create_crc_libvirt_sh {
     chmod +x $destDir/crc_libvirt.sh
 }
 
+function resize_and_sparsify {
+    local baseDir=$1
+    local srcFile=$2
+    local destFile=$3
+    local tmpFile=$(mktemp $baseDir/crc-resizedXXXX.qcow2)
+
+    # Check which partition is labeled as `root`
+    partition=$(${VIRT_FILESYSTEMS} -a $baseDir/$srcFile -l --partitions | sort -rk4 -n | sed -n 1p | cut -f1 -d' ')
+
+    # Resize the root partition from the default 1+15GB to 1+30GB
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -35,6 +35,59 @@ function create_crc_libvirt_sh {
...
+zero-free-space /boot/
+EOF
+    if [ $? -ne 0 ]; then
+            echo "Failed to resize and sparsify $baseDir/$srcFile, aborting"
+            exit 1
+    fi
+
+    ${GUESTFISH} --remote -- exit
+
+    ${QEMU_IMG} convert -p -f qcow2 -O qcow2 -o lazy_refcounts=on $tmpFile $baseDir/$destFile
```
praveenkumar: :+1:

createdisk.sh
```diff
@@ -35,6 +35,59 @@ function create_crc_libvirt_sh {
...
+    if [ $? -ne 0 ]; then
+            echo "Failed to sparsify $tmpFile, aborting"
+            exit 1
+    fi
+
+    rm $tmpFile
+    rm -fr $baseDir/.guestfs-*
```
praveenkumar: :+1:

Issue #137, opened 2019-12-03
Remove unused images before creating the bundle


praveenkumar:
As part of CRC we disable some of operators like (monitoring /machine-config ..etc.) but the images are always present on the node since there is no knob on installer side right now to disable those
operators from starting (iirc.) Now overall disk size is increases, till 4.2 our final disk size was around 2GB but with 4.3 it is increasing around 3GB and that is because all the other images are added as part of CVO payload.

```
=== For 4.2.x ==
# du -sh /var/lib/containers/
5.8G /var/lib/containers/

$ LC_ALL=C virt-df -a ~/.crc/cache/crc_libvirt_4.2.8/crc.qcow2  -h
Filesystem                                Size       Used  Available  Use%
crc.qcow2:/dev/sda3                        30G       8.1G        22G   28%

=== For 4.3.x ===
# du -sh containers/
6.6G containers/

[core@crc-ssnp7-master-0 ~]$ sudo df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/dm-0        16G   11G  4.6G  71% /sysroot
```

I am looking at pruning the unused images using the `crictl` itself after we disable the operators and set the deployments for those to 0. As per [0] there is `crictl images prune` command available but that is for new versions. So I am thinking about using a workaround `crictl images -q | xargs -n 1 crictl rmi` which is listed on the issue.

[0] https://github.com/kubernetes-sigs/cri-tools/issues/399

praveenkumar:
Fixed by #143

PR #143, opened 2020-01-17
Issue #137 Remove unused images from container storage

praveenkumar:
As part of CRC we disable some of operators like (monitoring/machine-config ..etc.) but
the images are always present on the node since there is no knob on installer side right now
to disable those operators from starting (iirc.) Now overall disk size is increases,
till 4.2 our final disk size was around 2GB but with 4.3 it is increasing around 3GB
and that is because all the other images are added as part of CVO payload.

crictl version used in the RHCOS is `0.1.0` which doesn't have fix
for kubernetes-sigs/cri-tools#399 one yet so
using kubernetes-sigs/cri-tools#399 (comment) as workaround.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -271,6 +271,9 @@ ${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterr
 certImage=$(${OC} --config $1/auth/kubeconfig adm release info --image-for=cluster-kube-apiserver-operator)
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman tag $certImage openshift/cert-recovery
 
+# Remove unused images from container storage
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -271,6 +271,9 @@ ${OC} --config $1/auth/kubeconfig create clusterrolebinding developer --clusterr
 certImage=$(${OC} --config $1/auth/kubeconfig adm release info --image-for=cluster-kube-apiserver-operator)
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman tag $certImage openshift/cert-recovery
 
+# Remove unused images from container stroage
```
cfergeau: `storage`

cfergeau:
Apart from this comments, looks good to me.

praveenkumar:
> crictl images will list all images, even the one being used, but crictl rmi will only be able to remove the unused images, and will error out on the other images. Is that correct?

@cfergeau yes that's right.

Issue #144, opened 2020-01-21
Add libvarlink-util package to work with podman remote


praveenkumar:
libvarlink-util package need to be part of the image so that it can work with podman-remote

praveenkumar:
Fixed by #145

PR #145, opened 2020-01-21
Issue #144 Add libvarlink-util package

praveenkumar:
podman-remote needs libvarlink-util to be part of VM so it can
able to connect the podman socket running inside the VM.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -287,6 +287,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
 # Stop the kubelet service so it will not reprovision the pods
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
+# Enable the io.podman.socket service
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable io.podman.socket
+
 # Remove all the pods from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crictl pods -q)'
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q)'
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -287,16 +287,19 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
...
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q)'
 
 # Remove pull secret from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
 
-# Download the hyperV daemons dependency on host
+# Download the hyperV daemons and libvarlink-util dependency on host
 mkdir $1/hyperv
-sudo yum install -y --downloadonly --downloaddir $1/hyperv hyperv-daemons
+sudo yum install -y --downloadonly --downloaddir $1/hyperv hyperv-daemons libvarlink-util
```
gbraad: You add the package to download, but you aren't installing it into the image with:

```
${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install *.rpm'
```
 ?

praveenkumar: @gbraad it is downloaded in same directory so all the `*.rpm` should be installed.

gbraad: I notice, but in that case please rename paths:

https://github.com/code-ready/snc/blob/846938326d2184545866d823738365c72d674c16/createdisk.sh#L297-L308

This shouldbe called `packages` or similar and not `hyperv` as this gives the wrong idea for future maintenance

praveenkumar: @gbraad done.

createdisk.sh
```diff
@@ -298,17 +298,17 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl
...
+mkdir $1/packages
+sudo yum install -y --downloadonly --downloaddir $1/packages hyperv-daemons libvarlink-util
 
 # SCP the downloaded rpms to VM
-${SCP} -r $1/hyperv core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
+${SCP} -r $1/packages core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
 
-# Install the hyperV rpms to VM
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/core/hyperv/*.rpm'
+# Install the hyperV and libvarlink-util rpms to VM
```
cfergeau: This line probably belongs to the previous commit

Issue #147, opened 2020-01-24
Remove pods before deleting the containers using crictl


praveenkumar:
We should remove the pods before deleting the container using crictl, this way old pods configuration/logs are not be part of our image.

praveenkumar:
This is not required anymore since the pods are cleaned when we shutdown the node gracefully.

PR #148, opened 2020-01-24
Issue #147 Remove pods before delete using crictl

praveenkumar:


createdisk.sh
```diff
@@ -287,6 +287,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
 # Stop the kubelet service so it will not reprovision the pods
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
+# Remove all the pod before delete them using crio
```
cfergeau: "deleting"

createdisk.sh
```diff
@@ -287,6 +287,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
...
+${OC} --config $1/auth/kubeconfig delete pods --all --all-namespaces --timeout=50s
+# Remove the pods from pending state
```
cfergeau: "in pending state"?

createdisk.sh
```diff
@@ -287,6 +287,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
...
+${OC} --config $1/auth/kubeconfig delete pods --all-namespaces --field-selector=status.phase=Pending
```
cfergeau: dunno if we can have a field selector matching both pods in Pending state, and in 'default' state so that we use a single command? (not really important, just random thought :)

PR #149, opened 2020-01-24
snc.sh: Get pullspec from mirrored release.txt

wking:
Add a `MIRROR` variable in case folks want to point at a different mirror for whatever reason.  If neither `OPENSHIFT_VERSION` nor a Git tag are available, use the `latest/` entry on the mirror to fetch the latest release (note that this is not pinned to a particular 4.y, it's just the latest stable release).  Drop `get_openshift_version` and just set up `OPENSHIFT_RELEASE_VERSION` instead; no need to delay that.  Echo the chosen release version and how it was determined to be more transparent and reduce surprise.  Drop empty-`OPENSHIFT_RELEASE_VERSION` checks from the rest of the script, now that `OPENSHIFT_RELEASE_VERSION` is always set.

Pull the release pullspec from the mirror's release.txt for the chosen version.  This decouples us from the Quay tagging scheme ([rhbz#794466][1], openshift/openshift-docs#19266), decouples us from the particular repository (doesn't even have to get pushed to Quay), and protects us from malicious registries that could point the mutable tag at a malicious image.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1794466

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -10,19 +10,30 @@ YQ=${YQ:-yq}
 OPENSHIFT_INSTALL=${OPENSHIFT_INSTALL:-./openshift-install}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
-QUAY_REGISTRY=${QUAY_REGISTRY:-quay.io/openshift-release-dev/ocp-release}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/clients/ocp}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 # Otherwise use the tagged version if available
-function get_openshift_version {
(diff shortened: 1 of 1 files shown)


snc.sh
```diff
@@ -10,19 +10,30 @@ YQ=${YQ:-yq}
 OPENSHIFT_INSTALL=${OPENSHIFT_INSTALL:-./openshift-install}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
-QUAY_REGISTRY=${QUAY_REGISTRY:-quay.io/openshift-release-dev/ocp-release}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/clients/ocp}
```
praveenkumar: @wking So is this correct assumption that `mirror.openshift.com` is source of truth when it comes to OCP releases? Recently when 4.3 rc bits happened those were not appeared on mirror.openshift.com atleast for a day or two. If you says this is way forward and every bits should be present here then we can merge it.

wking: > Recently when 4.3 rc bits happened those were not appeared on mirror.openshift.com at least for a day or two...

They're supposed to get pushed up around when the release is tagged into the candidate channel, with `latest*` symlinks bumped around when a release is tagged into fast/stable.  If that's not what's happening, talk to ART.

PR #150, opened 2020-01-24
snc.sh: Set pipefail, exit on errors and undefined variables

wking:
So the script doesn't blindly continue if something important like `install cluster` dies.  There may be commands where failure is acceptable, and this change will break those.  You can work around it with:

```sh
command-where-failure-is-ok || echo "this thing failed, but that is ok"
```

I'll leave those off in this commit, because I don't understand this script well enough to know which commands should be allowed to fail.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -euo pipefail
+
 export LC_ALL=C
 export LANG=C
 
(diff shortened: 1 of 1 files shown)


cfergeau:
I've tested this change, and had to ignore a few errors, but overall it worked fine. https://github.com/cfergeau/snc/commit/f36b2d54a47a969f19a1638c3db964e213e3a32d has the changes I applied on top of your branch.
There is one `mkdir` which I'm not sure should be ignored or not when it fails, or if we should fail earlier, ..

wking:
> cfergeau@f36b2d5 has the changes I applied on top of your branch.

I'm leaving `createdisk.sh` alone.  Feel free to file a separate PR to add `pipefail`, etc. to that script.

I've picked up your two `snc.sh` fallbacks (with explanatory `echo`, hopefully I got the "why we are ignoring this failure" reasoning right) with 16a0bb7 -> ecbb5d9.  How does that look?

cfergeau:
> I'm leaving createdisk.sh alone. Feel free to file a separate PR to add pipefail, etc. to that script.
Yup, I was not asking to add it to this PR, just some notes I took during my testing :) One more note about OPENSHIFT_VERSION which should be checked early, or have a default value in createdisk.sh.

PR #152, opened 2020-02-03
[WIP] Scale down the deployments and daemonset

praveenkumar:
This patch will scale down all the deployments and daemonset for
apiserver, node-exporter. This will allow us to scale only the deployment
which we want to start.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -280,6 +280,13 @@ ${OC} --config $1/auth/kubeconfig replace -f pull-secret.yaml
 # Remove the Cluster ID with a empty string.
 ${OC} --config $1/auth/kubeconfig patch clusterversion version -p '{"spec":{"clusterID":""}}' --type merge
 
+# Scale down deployments to 0
+${OC} --config $1/auth/kubeconfig get ns -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | xargs -n 1 ${OC} --config $1/auth/kubeconfig scale --replicas=0 deployment --all -n
+
+# Scale down daemonset to 0
+${OC} --config $1/auth/kubeconfig -n openshift-apiserver patch daemonset apiserver -p '{"spec": {"template": {"spec": {"nodeSelector": {"non-existing": "true"}}}}}'
+${OC} --config $1/auth/kubeconfig -n openshift-monitoring patch daemonset node-exporter -p '{"spec": {"template": {"spec": {"nodeSelector": {"non-existing": "true"}}}}}'
+
 # Disable kubelet service and pull dnsmasq image from quay.io/crcon/dnsmasq
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable kubelet
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont/dnsmasq:latest


praveenkumar:
> We actually need to keep the output of this visible to make sure we can compare what has been disabled and what therefore needs to be enabled.

As of now we are disabling each and every deployment and 2 different daemonsets.

cfergeau:
Couple of comments:
- It's not clear why we are doing this (Praveen explained that his goal is to ensure the pods are recreated from scratch when crc is first started to try to get them to drop state they may have inherited from running at install/snc time)
- it also may help with the 4.3 bug we're currently having with the apiserver not being functional after first boot with crc
- a small mention that the apiserver uses daemonsets rather than deployments so must be disabled tihs way could be useful (unless this is k8s 101 and really is stating the obvious)
- there is some code which becomes ineffective after this change (manual scaling to 0 of various deployments), it could be removed
- and same as Gerard, I'd feel more comfortable if we kept a list of the operators we disabled so that crc knows for sure what to reenable (more future-proof)

cfergeau:
Regarding the daemonsets, any reason you only disable 2 of them?
```shell
$ oc get daemonset --all-namespaces  -oname
daemonset.extensions/apiserver
daemonset.extensions/tuned
daemonset.extensions/controller-manager
daemonset.extensions/dns-default
daemonset.extensions/node-ca
daemonset.extensions/machine-config-daemon
daemonset.extensions/machine-config-server
daemonset.extensions/node-exporter
daemonset.extensions/multus
daemonset.extensions/multus-admission-controller
daemonset.extensions/ovs
daemonset.extensions/sdn
daemonset.extensions/sdn-controller
```

praveenkumar:
```
daemonset.extensions/multus
daemonset.extensions/multus-admission-controller
daemonset.extensions/ovs
daemonset.extensions/sdn
daemonset.extensions/sdn-controller
```

^^ These are required for networking so can't disable those by default, until we know for sure which part of networking they are responsible without breaking any networking stuff in our CRC cluster.

```
daemonset.extensions/machine-config-daemon
daemonset.extensions/machine-config-server
daemonset.extensions/node-exporter
```

^^ These we can disable without any issue since we are disabling the MCO and also monitoring one.

```
daemonset.extensions/tuned
```
^^ As per docs https://github.com/openshift/openshift-tuned it is required for kernel tuning, we can disable it also by default but need to mention it somewhere in the doc if a user need it.

```
daemonset.extensions/node-ca
```
^^ This is part of image registry and used when a custom registry cert need to be added. 

```
daemonset.extensions/controller-manager
daemonset.extensions/dns-default
```

^^ These are for openshift controller-manager and openshift-dns side, these can be come up with respective operator if we disable those.

cfergeau:
By "disabling", I really meant "disable them at snc time before reenabling them at crc start time", not "disable them permanently", your answer seems to be about the latter situation?

praveenkumar:
> By "disabling", I really meant "disable them at snc time before reenabling them at crc start time", not "disable them permanently", your answer seems to be about the latter situation?

@cfergeau I was just pointing out what we can disable as default even on snc side, except networking component we pretty much able to disable all the daemonset.

praveenkumar:
Putting this back to WIP, until we decide to go with this path.

praveenkumar:
/close

This was created because of a POC and we don't need it anymore.

PR #157, opened 2020-02-25
CVO override

praveenkumar:


cvo_override.yaml
```diff
@@ -0,0 +1,38 @@
...
+    namespace: openshift-monitoring
+    unmanaged: true
+  - kind: Deployment
+    group: apps/v1
+    name: machine-config-operator
+    namespace: openshift-machine-config-operator
+    unmanaged: true
+  - kind: Deployment
+    group: apps/v1
+    name: etcd-quorum-guard
```
cfergeau: If we 'unmanage'  `machine-config-operator`, do we also need to unmanage this one?

cvo_override.yaml
```diff
@@ -0,0 +1,38 @@
...
+    namespace: openshift-machine-api
+    unmanaged: true
+  - kind: Deployment
+    group: apps/v1
+    name: insights-operator
+    namespace: openshift-insights
+    unmanaged: true
+  - kind: Deployment
+    group: apps/v1
+    name: prometheus-k8s
```
cfergeau: Is there a more generic monitoring operator we want to disable? or is this the "main" one?

snc.sh
```diff
@@ -223,26 +223,30 @@ create_pvs "${CRC_PV_DIR}" 30
 # https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#setting-objects-unmanaged
 ${OC} patch clusterversion version --type json -p "$(cat cvo_override.yaml)"
 
-# Get the pod name associated with cluster-monitoring-operator deployment
```
cfergeau: For each of these blocks, I think I'd keep a `# cluster-monitoring-operator` comment. Or maybe this could be a helper func?

snc.sh
```diff
@@ -223,26 +223,30 @@ create_pvs "${CRC_PV_DIR}" 30
...
 cmo_pod=$(${OC} get pod -l app=cluster-monitoring-operator -o jsonpath="{.items[0].metadata.name}" -n openshift-monitoring)
```
cfergeau: have you considered using `-oname` instead?

snc.sh
```diff
@@ -223,26 +223,30 @@ create_pvs "${CRC_PV_DIR}" 30
...
-# Disable the deployment/replicaset/statefulset config for openshift-monitoring namespace
-${OC} scale --replicas=0 deployment --all -n openshift-monitoring
-# Wait till the cluster-monitoring-operator pod is deleted
+${OC} delete deployment cluster-monitoring-operator -n openshift-monitoring
+# Wait till the cluster-monitoring-operator pod is deleted before deleting other resources
```
cfergeau: I'd replace `other resources` with `the resources it manages`

snc.sh
```diff
@@ -223,26 +223,30 @@ create_pvs "${CRC_PV_DIR}" 30
...
 
-# Set replica to 0 for openshift-insights
-${OC} scale --replicas=0 deployment --all -n openshift-insights
+mco_pod=$(${OC} get pod -l k8s-app=machine-config-operator -o jsonpath="{.items[0].metadata.name}" -n openshift-machine-config-operator)
+${OC} delete deployment machine-config-operator -n openshift-machine-config-operator
+# Wait till the machine-config-operator pod is deleted before deleting other resources
+${OC} wait --for=delete pod/$mco_pod --timeout=60s -n openshift-machine-config-operator
+${OC} delete statefulset,deployment,daemonset --all -n openshift-machine-config-operator
+
+${OC} delete statefulset,deployment,daemonset --all -n openshift-insights
```
cfergeau: no need to delete the operator first for openshift-insights?

snc.sh
cfergeau: Regarding the commit log
>  If a user want that specific operator he can just remove it from CVO's override list and all the resources are repopulated.

How to do this *must* be documented in `crc`

snc.sh
```diff
@@ -251,12 +251,6 @@ ${OC} delete statefulset,deployment,daemonset --all -n openshift-insights
 # Scale route deployment from 2 to 1
 ${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/default -n openshift-ingress-operator
 
-# Scale console deployment from 2 to 1
-${OC} scale --replicas=1 deployment.apps/console -n openshift-console
-
-# Scale console download deployment from 2 to 0
-${OC} scale --replicas=0 deployment.apps/downloads -n openshift-console
-
```
cfergeau: How many console deployments/pods/... will we get after this change? Does this have an adverse impact on our cluster memory footprint?

snc.sh
```diff
@@ -268,11 +271,5 @@ ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op":
 # Since this CRD block namespace deletion forever.
 ${OC} delete apiservice v1beta1.metrics.k8s.io
 
-# Get the cert pod name
-cert_pod=$(${OC} get pod -l k8s-app=kubelet-bootstrap-cred-manager -o jsonpath="{.items[0].metadata.name}" -n openshift-machine-config-operator)
-# Remove the bootstrap-cred-manager daemonset
-${OC} delete daemonset.apps/kubelet-bootstrap-cred-manager -n openshift-machine-config-operator
-# Wait till the cert pod is removed
-${OC} wait --for=delete pod/$cert_pod --timeout=120s -n openshift-machine-config-operator
 # Remove the cli image which was used for the bootstrap-cred-manager daemonset
```
cfergeau: why do we delete this image?

snc.sh
```diff
@@ -224,10 +224,20 @@ create_pvs "${CRC_PV_DIR}" 30
...
 # Wait till the cluster-monitoring-operator pod is deleted before deleting other resources
 ${OC} wait --for=delete pod/$cmo_pod --timeout=120s -n openshift-monitoring
+${OC} delete deployment prometheus-operator -n openshift-monitoring
+# Wait till the prometheus operator pod is deleted before deleteing other resoureces
+${OC} wait --for=delete pod/$prometheus_op_pod -n openshift-monitoring
 ${OC} delete deployment,statefulset,daemonset --all -n openshift-monitoring
+# Wait till the prometheus-adapter pods (part of statefulset) deleted
+${OC} wait --for=delete pod/$prometheus_pod --timeout=120s -n openshift-monitoring
+# Wait till the alertmanager pods (part of statefulset) deleted
+${OC} wait --for=delete pod/$alertmanager_pod --timeout=120s -n openshift-monitoring
```
cfergeau: are these wait required ?

cvo_override.yaml
praveenkumar: @cfergeau yes some of the operators which are part of different operators managed by CVO, I think those operators which are part of same namespace.

cvo_override.yaml
praveenkumar: @cfergeau the generic one is openshift-monitioring but it also consist prometheus one which again maintained by CVO so this again need to be disabled.

snc.sh
praveenkumar: Yes it should go in the doc side.

snc.sh
praveenkumar: @cfergeau yes we can use that, I will put another PR for that.

snc.sh
praveenkumar: It is standalone operator and only have single operator deployment so no need to do it.

snc.sh
praveenkumar: Because this image is extra which we download to rotate the certs.

snc.sh
praveenkumar: @cfergeau yes those are required because sometime pod does take a lot of time for graceful shutdown. (default wait time is around 30sec)

cvo_override.yaml
cfergeau: Are you saying `etcd-quorum-guard` is an operator of its own, and that it's not managed by `machine-config-operator`?

snc.sh
cfergeau: Even if we don't need to do it now, isn't it more future-proof (and more consistent) to handle it in the same way as all other operators?

cvo_override.yaml
praveenkumar: @cfergeau it is managed by CVO also.

snc.sh
praveenkumar: @cfergeau no the memory footprint is almost same as we have previously.

snc.sh
praveenkumar: sure, we can add this as part of new PR.

cfergeau: Since you add this code as part of this PR, it would have made much more sense to add it to this one rather than changing it afterwards.

PR #160, opened 2020-03-04
Issue #159 Fetch pre-built installer from openshift images 

praveenkumar:


snc.sh
```diff
@@ -177,12 +177,6 @@ ${YQ} write --inplace $INSTALL_DIR/manifests/cluster-ingress-02-config.yml spec[
 ${YQ} write --inplace $INSTALL_DIR/openshift/99_openshift-cluster-api_master-machines-0.yaml spec.providerSpec.value[domainMemory] 12288
 ${YQ} write --inplace $INSTALL_DIR/openshift/99_openshift-cluster-api_master-machines-0.yaml spec.providerSpec.value[domainVcpu] 6
 
-# Start the cluster with 10GB memory and 4 CPU create and wait till it finish
-# Todo: we need to remove this once stop building 4.2 bits.
-# For 4.3 this is ignored.
-export TF_VAR_libvirt_master_memory=10192
-export TF_VAR_libvirt_master_vcpu=4
-
```
cfergeau: Maybe mention in the log that snc is already modifying `openshift/99_openshift-cluster-api_master-machines-0.yaml` so that it uses 12GB RAM and 6 CPUs

cfergeau: But feel free to push this (`Remove env variable for master resources (mem/cpu) `) now even without the other patch which I haven't looked at yet :)

snc.sh
```diff
@@ -137,7 +134,19 @@ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -l "${MIRROR}/${OPENSHIFT_RELEA
 export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE
 echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
 
+# Extract openshift-install binary if not present in current direcory
+if ! which $OPENSHIFT_INSTALL; then
+    echo ${OPENSHIFT_PULL_SECRET} > pull-secret
+    baremetal_installer_image=$(oc adm release -a pull-secret info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=baremetal-installer)
+    oc image -a pull-secret extract ${baremetal_installer_image} --confirm --path /usr/bin/openshift-install:.
+    chmod +x openshift-install
+    rm pull-secret
```
cfergeau: Not sure it should go to current directory, or to some subdirectory. We need to set OPENSHIFT_INSTALL to the path to the installer after doing this though, as the `which` test could fail because `OPENSHIFT_INSTALL` was set to some wrong value.

snc.sh
```diff
@@ -137,7 +134,19 @@ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -l "${MIRROR}/${OPENSHIFT_RELEA
 export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE
 echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
 
+# Extract openshift-install binary if not present in current direcory
+if ! which $OPENSHIFT_INSTALL; then
+    echo ${OPENSHIFT_PULL_SECRET} > pull-secret
```
cfergeau: `echo "Using installer from OpenShift baremetal-installer image"` ?

snc.sh
cfergeau: Seeing .gitignore, we also use the current directory for the generated ssh keys. Maybe it's a good opportunity to start using `$INSTALL_DIR/snc_data/` and put the pull secret and the installer there?

cfergeau: Actually, one advantage of not putting it in ./openshift-install is that snc will fetch the image/the binary every time if ./openshift-install does not exist, which ensures we are not using an outdated binary. But that can be improved later if you prefer.

praveenkumar: @cfergeau even you put an `$INSTALL_DIR/snc_data` then a user need to delete this directory before caching another installer binary right or you mean to say that then irrespective to data in the directory we are always going to rewrite the installer binary? In the current form if I want to test a different installer binary with different payload I can just compile the installer binary and put it in snc directory so the extraction process ignore it (This would be good for debugging) but we might lose that if we always force to rewrite installer binary.

cfergeau:
```
if ! which $OPENSHIFT_INSTALL; then 
# put installer in somewhere/openshift-install
# OPENSHIFT_INSTALL=somewhere/openshift-install
fi
```
if OPENSHIFT_INSTALL is set/./openshift_install exist, they are going to be used. If they don't exist, then the installer will always be extracted from the baremetal image.
After we've tested this PR for a while, we probably can remove the default value for OPENSHIFT_INSTALL

snc.sh
```diff
@@ -137,7 +134,20 @@ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -l "${MIRROR}/${OPENSHIFT_RELEA
...
 echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
 
+# Extract openshift-install binary if not present in current direcory
+if ! which $OPENSHIFT_INSTALL; then
+    echo "Extracting installer binary from OpenShift baremetal-installer image"
+    echo ${OPENSHIFT_PULL_SECRET} > pull-secret
+    baremetal_installer_image=$(oc adm release -a pull-secret info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=baremetal-installer)
+    oc image -a pull-secret extract ${baremetal_installer_image} --confirm --path /usr/bin/openshift-install:.
+    chmod +x openshift-install
+    rm pull-secret
```
cfergeau: Still missing `OPENSHIFT_INSTALL="./openshift-install"` or `OPENSHIFT_INSTALL="$(pwd)/openshift-install"`

zeenix:
README needs to be updated too.

PR #165, opened 2020-03-19
Add rolebinding for developer user to be able to list installed operators

anjannath:
To view installed operators access to resources from operators.coreos.com
is needed, this patch adds a cluster role for this and binds it to the
developer user.

Related to: https://github.com/code-ready/crc/issues/1096


anjannath:
Closing as this is a bug in the webconsole UI and should be handled there.. https://github.com/code-ready/crc/issues/1096

PR #166, opened 2020-03-23
Update memory from 12GB to 14GB

praveenkumar:
In 4.3.8 looks like the cluster is taking almost full allocated memory
so to start the cluster in sane state we need to increase the memory
resource.


snc.sh
```diff
@@ -234,7 +234,7 @@ ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests || exit 1
 ${YQ} write --inplace ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml spec[domain] apps-${CRC_VM_NAME}.${BASE_DOMAIN}
 # Add master memory to 12 GB and 6 cpus
```
jmazzitelli: FYI: I think you want to change this comment now - should be 14GB

PR #169, opened 2020-04-16
Add instance IP to metadata

praveenkumar:
From 4.4 release etcd certs are now created for
internal IP instead dns srv record like it used to, in case
of the libvirt the internal IP is same as VM ip.

### DIFF
--- a/crc-bundle-info.json.sample
+++ b/crc-bundle-info.json.sample
@@ -45,6 +45,8 @@
       # Disk image used by the node, the 'storage' object will contain more
       # details about its format
       "diskImage": "crc.qcow2"
+      # Internal IP for which etcd certs are valid
+      "internalIP": "192.168.126.11"
       # kernel command line of the node
       "kernelCmdLine": "BOOT_IMAGE=/ostree/rhcos-bf3b38268ddb2a2070dc587b361ce45c46a6f9e3606bbd3b5e15b2e6d3d47e5d/vmlinuz-4.18.0-80.1.2.el8_0.x86_64 console=tty0 console=ttyS0,115200n8 rootflags=defaults,prjquota rw root=UUID=a8fbdcb1-63ea-421e-8f6f-dac9cbbcc822 ostree=/ostree/boot.0/rhcos/bf3b38268ddb2a2070dc587b361ce45c46a6f9e3606bbd3b5e15b2e6d3d47e5d/0 coreos.oem.id=qemu ignition.platform.id=qemu"
       # initramfs file of the node
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -130,6 +130,7 @@ function update_json_description {
         | ${JQ} '.nodes[0].kind[1] = "worker"' \
         | ${JQ} ".nodes[0].hostname = \"${VM_PREFIX}-master-0\"" \
         | ${JQ} ".nodes[0].diskImage = \"${CRC_VM_NAME}.qcow2\"" \
+        | ${JQ} ".nodes[0].internalIP = \"${INTERNAL_IP}\"" \
         | ${JQ} ".storage.diskImages[0].name = \"${CRC_VM_NAME}.qcow2\"" \
         | ${JQ} '.storage.diskImages[0].format = "qcow2"' \
         | ${JQ} ".storage.diskImages[0].size = \"${diskSize}\"" \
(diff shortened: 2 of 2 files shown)


praveenkumar:
@cfergeau it is still required to put entry on the dns side like this one https://github.com/code-ready/crc/pull/1154/commits/007e19216272b7cc11311da4589e78a73a353cb5#diff-4a1356e5ac05231bb116112b03f61e34R27 otherwise how the hostname resolved to the dummy IP.

createdisk.sh
```diff
@@ -338,6 +341,9 @@ kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cm
 # SCP the vmlinuz/initramfs from VM to Host in provided folder.
 ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/rhcos-${ostree_hash}/* $1
 
+# Get the IP of the VM
+internalIP=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'hostname -i')
```
cfergeau, line 345: This returns multiple IP addresses on my crc VM and my laptop, how do we pick the right one? The manpage also recommends `-I` rather than `-i` If the etcd certificate or port is easy to access, we could get the IP to use directly from it.

createdisk.sh
```diff
@@ -370,7 +376,7 @@ mkdir $libvirtDestDir
 
 create_qemu_image $libvirtDestDir
 
-copy_additional_files $1 $libvirtDestDir
+copy_additional_files $1 $libvirtDestDir $internalIP
```
cfergeau, line 385: we just use global variables for the other variables in the metadata file, so it's not strictly required to explicity pass it here.

crc-bundle-info.json.sample
```diff
@@ -45,6 +45,8 @@
       # Disk image used by the node, the 'storage' object will contain more
```
cfergeau, line 45: > From 4.4 release etcd certs are now created for internal IP instead dns srv record like it used to, in case of the libvirt the internal IP is same as VM ip.

Did the certificate directly refer to the SRV records? My understanding is that the SRV records were used to get the hostname of the etcd server, and then the certificates were valid for that hostname.
Maybe
>  Starting from the 4.4 release, etcd certs are now only valid for etcd IP. In previous releases, the etcd instances were accessed through hostnames so the etcd IP was not that important. Now, crc will need to know for which IP these certificates are valid. In the libvirt case, the internal IP is same as VM ip.

createdisk.sh
```diff
@@ -347,6 +347,9 @@ internalIP=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'hostname -i')
 # Add a dummy network interface with internalIP
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo nmcli conn add type dummy ifname eth10 con-name dummy ip4 ${internalIP}/24  && sudo nmcli conn up dummy"
 
+# Add internalIP as node IP for kubelet systemd unit file
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo sed -i.back '/kubelet /a\      --node-ip="$ip" \\\' /etc/systemd/system/kubelet.service"
+
```
cfergeau, line 352: > This will make sure when kubelet service run then node internal IP would be what defined here instead of detecting it automatic.
"service runs ... what is defined here ... automatically"

createdisk.sh
```diff
@@ -344,6 +344,9 @@ ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/rhcos-${ostree_hash}/
 # Get the IP of the VM
 internalIP=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'hostname -i')
 
+# Add a dummy network interface with internalIP
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo nmcli conn add type dummy ifname eth10 con-name dummy ip4 ${internalIP}/24  && sudo nmcli conn up dummy"
```
cfergeau, line 348: we can use a more descriptive name than `dummy`. Maybe `internalEtcd` since this is what it's going to be used for?

cfergeau: I believe a much more detailed commit log would be useful here

>  Add a dummy network using internal IP

> As part of etcd the certs are created for internalIP instead of srv records. This network IP is going to be instance IP so that the certs are always valid for it even running on different platform.

The problem we are trying to solve is that the cluster/node IP is going to change depending on the platform CRC runs on, and the etcd certificate is IP only. How we are solving it, by adding a dummy network interface on the node whose IP will match the one expected by etcd, and we'll make sure the cluster uses this internally. All of this is worthwhile to have in the log in my opinion.

createdisk.sh
praveenkumar: Alright I will use something like `internalIP=$(${DIG} +short api.${CRC_VM_NAME}.${BASE_DOMAIN})` then.

PR #171, opened 2020-04-20
Fix snc.sh fails if oc not in $PATH

noseka1:


snc.sh
```diff
@@ -115,7 +115,7 @@ function create_pvs() {
 function renew_certificates() {
     # Get the cli image from release payload and update it to bootstrap-cred-manager resource
     echo ${OPENSHIFT_PULL_SECRET} > pull-secret
-    cli_image=$(oc adm release -a pull-secret info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=cli)
+    cli_image=$(${OC} adm release -a pull-secret info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=cli)
```
praveenkumar, line 118: :+1:

cfergeau:
`git grep '\<oc\>'` returns 2 more instances where this would be an issue, do you mind changing these too? I can send a follow-up patch if you prefer.

noseka1:
If you mean these two:
```
...
$ git grep '\<oc\>'
snc.sh:    if [[ ! -e oc ]] ; then
...
snc.sh:        curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-linux-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar zx oc
...
```
Those should stay as they are.

cfergeau:
> > 
> > Those should stay as they are.
> 
> Nope, the last 2 ones below:

Sorry for the noise, I missed one hunk in your PR with 2 hunks /o\

PR #177, opened 2020-05-13
Issue #170 Add apache 2.0 license

### LINKED ISSUE #170: Include a License
CRC is using Apache v2. Should the same be here?

Thanks.

anjannath:
But the generated bundle/image are not apache licensed

Fixes #170

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+                                Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,202 @@
+Apache License
```
gbraad, line 1: indent wrong?

anjannath:
@gbraad its there by default, even in the CRC repo, https://github.com/code-ready/crc/blob/master/LICENSE

anjannath:
@gbraad Fixed the indentation, not sure from where we inherited it!

mtarsel:
Sorry but if this is taken from CRC - shouldn't the name and year be added too? https://github.com/code-ready/crc/blob/master/LICENSE#L189 (same as this PR).

mtarsel:
Disregard my last comment as that is only for adding the license to specific files.

PR #181, opened 2020-06-09
oc flag --config has been deprecated, use --kubeconfig instead

praveenkumar:
With new version of oc `--config` flag might not even available so
better to update it to `--kubeconfig` flag.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -269,22 +269,22 @@ if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checken
 fi
 
 # Add a user developer:developer with htpasswd identity provider and give it sudoer role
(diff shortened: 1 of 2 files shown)


cfergeau:
It seems `--kubeconfig` is fairly new (oc 4.4 and newer?), and https://github.com/openshift/oc/pull/256 has no details about the deprecation plan. I'd keep using `--config` for now, easy enough to fix when needed.

praveenkumar:
> It seems --kubeconfig is fairly new (oc 4.4 and newer?)

@cfergeau it is not new but added around 4.1 time frame itself also our current master is for 4.4 and new version image generation, we tag/branch when we start creating new version of openshift bundle.

cfergeau:
Oh ok, 4.1 is fine, from looking at the git log I had the feeling this was very new :) Worth mentioning in the commit log imo.

PR #183, opened 2020-06-15
Fix shellcheck error SC1075: Use 'elif' instead of 'else if'

guillaumerose:
This is the only shellcheck error. 4 warnings are still present, I will take of them after.

Perhaps we could configure some CI to run shellcheck ? (a github action maybe?)


cfergeau:
github actions for shellcheck and similar sounds good. And the patch looks good too.

PR #185, opened 2020-06-25
Add preflight checks

cfergeau:
When alternating between running snc.sh and running snc, and rebooting
the machine in between, it's easy to get the host configuration wrong,
and to have the libvirt installer fails after running for an hour or so.
This commit adds checks for the more common mistakes in host
configuration, the only one which is not checked yet is whether the
192.168.126.0/24 network can access the 192.168.122.0/24 network.

This also adds a dependency on xmlstarlet but there is no check for it yet.
This could alternatively probably be handled with a relatively simple go code,
the resulting binary could be automatically downloaded, not sure which way is
best.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -79,19 +79,22 @@ function run_preflight_checks() {
         # check if firewalld is configured to allow traffic from 192.168.126.0/24 to 192.168.122.1
         # this check is very basic and expects the configuration to match
         # https://github.com/openshift/installer/tree/master/docs/dev/libvirt#firewalld
-        local zone
-        if firewall-cmd -h >/dev/null; then
-                # With older libvirt, the 'libvirt' zone will not exist
-                if firewall-cmd --get-zones |grep '\<libvirt\>'; then
-                        zone=libvirt
-                else
-                        zone=dmz
-                fi
-                if ! firewall-cmd --zone=${zone} --list-services | grep '\<libvirt\>'; then
-                        preflight_failure "firewalld is available, but it is not configured to allow 'libvirt' traffic in either the 'libvirt' or 'dmz' zone, please check https://github.com/openshift/installer/tree/master/docs/dev/libvirt#firewalld"
-                        return
-                fi
-        fi
+        # Disabled for now as on stock RHEL8 installs, additional permissions are needed for
+        # firewall-cmd --list-services, so this test fails for unrelated reasons
+        #
+        #local zone
+        #if firewall-cmd -h >/dev/null; then
+        #        # With older libvirt, the 'libvirt' zone will not exist
+        #        if firewall-cmd --get-zones |grep '\<libvirt\>'; then
+        #                zone=libvirt
+        #        else
+        #                zone=dmz
+        #        fi
+        #        if ! firewall-cmd --zone=${zone} --list-services | grep '\<libvirt\>'; then
+        #                preflight_failure "firewalld is available, but it is not configured to allow 'libvirt' traffic in either the 'libvirt' or 'dmz' zone, please check https://github.com/openshift/installer/tree/master/docs/dev/libvirt#firewalld"
+        #                return
+        #        fi
+        #fi
 
         echo "libvirt and DNS configuration successfully checked"
 }
--- a/snc.sh
+++ b/snc.sh
@@ -8,6 +8,7 @@ export LANG=C
 INSTALL_DIR=crc-tmp-install-data
 JQ=${JQ:-jq}
 OC=${OC:-oc}
+XMLLINT=${XMLLINT:-xmllint}
 YQ=${YQ:-yq}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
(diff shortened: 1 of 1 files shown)


praveenkumar:
> When alternating between running snc.sh and running snc, and rebooting
the machine in between.

@cfergeau In commit logs ` running snc.sh and running snc` doesn't seems right you mean crc ?

Also we should add the `xmlstarlet` check like https://github.com/code-ready/snc/blob/master/snc.sh#L184-L186 one so if it not present by default should be installed,

cfergeau:
> > When alternating between running snc.sh and running snc, and rebooting
> > the machine in between.
> 
> @cfergeau In commit logs ` running snc.sh and running snc` doesn't seems right you mean crc ?

Yes, thanks for catching this :)

> 
> Also we should add the `xmlstarlet` check like https://github.com/code-ready/snc/blob/master/snc.sh#L184-L186 one so if it not present by default should be installed,

Yeah, I know we need a check, I was not sure if it was preferred to depend on xmlstarlet, or to write our own tool, I should have marked it as WIP. I'll add this.

cfergeau:
Another thing which is missing is that for now it only prints a message. Should this exit instead? Or check an env var to decide between warn/exit?

snc.sh
```diff
@@ -35,6 +36,64 @@ else
...
+
+        LIBVIRT_URI=qemu+tcp://localhost/system
+
+        # check if libvirtd is listening on a TCP socket
+        if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
+                preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
+        fi
+
+        # check if libvirtd has access to an x86_64 hypervisor
+        if ! virsh -c ${LIBVIRT_URI} capabilities | ${XMLSTARLET} sel -t -c '/capabilities/guest/arch[@name="x86_64"]' >/dev/null; then
```
praveenkumar, line 58: Shouldn't this check will fail for the folks who want to use snc for powerpc or other arch ?

snc.sh
```diff
@@ -35,6 +36,64 @@ else
...
+        if ! virsh -c ${LIBVIRT_URI} capabilities | ${XMLSTARLET} sel -t -c '/capabilities/guest/arch[@name="x86_64"]/domain[@type="kvm"]' >/dev/null; then
+                preflight_failure "Your platform does not provide a hardware-accelerated hypervisor, it's strongly recommended to enable it before running SNC. Check virt-host-validate for more detailed diagnostics"
+                return
+        fi
+
+        # check that api.crc.testing either can't be resolved, or resolves to 192.168.126.1[01]
+        local ping_status
+        ping_status="$(ping -c1 api.crc.testing | head -1 || true >/dev/null)"
+        if echo ${ping_status} | grep "PING api.crc.testing (" && ! echo ${ping_status} | grep "192.168.126.1[01])"; then
+                preflight_failure "DNS setup seems wrong, api.crc.testing resolved to an IP which is neither 192.168.126.10 nor 192.168.126.11, please check your NetworkManager configuration and /etc/hosts content"
```
praveenkumar, line 73: :+1:

snc.sh
```diff
@@ -35,6 +36,64 @@ else
...
+        # this check is very basic and expects the configuration to match
+        # https://github.com/openshift/installer/tree/master/docs/dev/libvirt#firewalld
+        local zone
+        if firewall-cmd -h >/dev/null; then
+                # With older libvirt, the 'libvirt' zone will not exist
+                if firewall-cmd --get-zones |grep '\<libvirt\>'; then
+                        zone=libvirt
+                else
+                        zone=dmz
+                fi
```
praveenkumar, line 87: what if a user have created new zone manually and configure the libvirt traffic to that zone or we just want a check here which is listed to the https://github.com/openshift/installer/tree/master/docs/dev/libvirt#firewalld doc?

snc.sh
```diff
@@ -185,6 +244,12 @@ if ! which ${JQ}; then
     sudo yum -y install /usr/bin/jq
 fi
 
+if ! which ${XMLSTARLET}; then
+    sudo yum -y install /usr/bin/xmlstarlet
```
praveenkumar, line 248: :+1:

snc.sh
cfergeau: Yes it will fail for these users, which is why there's a way to disable the failed preflights.
I'd need to change this test to
```
arch=$(virsh  capabilities | xmlstarlet sel -t -c '/capabilities/host/cpu/arch')
if [ arch != "x86_64 ]; then
    # warn
fi
```
and then use $arch in the kvm test instead of hardcoding `x86_64`

snc.sh
cfergeau: I'm aware of this limitation which is mentioned in the comment before this test. I'm not sure many people are going to do a different configuration :) If you have suggestions for alternate ways of testing this, I can try a different test, I was short on ideas on how to achieve this one.

snc.sh
```diff
@@ -35,6 +36,66 @@ else
...
+        # https://github.com/openshift/installer/tree/master/docs/dev/libvirt#firewalld
+        local zone
+        if firewall-cmd -h >/dev/null; then
+                # With older libvirt, the 'libvirt' zone will not exist
+                if firewall-cmd --get-zones |grep '\<libvirt\>'; then
+                        zone=libvirt
+                else
+                        zone=dmz
+                fi
+                if ! firewall-cmd --zone=${zone} --list-services | grep '\<libvirt\>'; then
```
cfergeau, line 90: This does not work so well on a stock rhel8 install, fails with
```
Authorization failed.
    Make sure polkit agent is running or run the application as superuser.
```
Maybe best to skip this check for now.

snc.sh
```diff
@@ -35,6 +36,66 @@ else
...
+        if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
+                preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
+        fi
+
+        # check if libvirtd has access to an x86_64 hypervisor
+        local arch
+        arch=$(virsh  capabilities | xmlstarlet sel -t -v '/capabilities/host/cpu/arch')
+        if [ ${arch} != "x86_64" ]; then
+                # warn only
+                echo "The host architecture is ${arch}, SNC has only been tested on x86_64"
```
praveenkumar, line 62: :+1:

snc.sh
praveenkumar: @cfergeau sure, let's update the PR with the changes.

praveenkumar:
@cfergeau this need rebase and also shouldn't both commit squashed and be only one adding preflight check since firewall check anyway is commented?

cfergeau:
> @cfergeau this need rebase and also shouldn't both commit squashed and be only one adding preflight check since firewall check anyway is commented?

squashing/not squashing, I'm open to both options, with the separate commit it's obvious why it's commented out, and easy to revert :)

PR #186, opened 2020-06-26
Disk resize related changes

praveenkumar:


snc.sh
```diff
@@ -257,6 +257,7 @@ ${YQ} write --inplace ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml spe
 # This is only valid for openshift 4.3 onwards
 ${YQ} write --inplace ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml spec.providerSpec.value[domainMemory] 14336
 ${YQ} write --inplace ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml spec.providerSpec.value[domainVcpu] 6
+${YQ} write --inplace ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml spec.providerSpec.value.volume[volumeSize] 32212254720
```
cfergeau, line 260: This is 30GiB (might be worth a comment, as  32212254720 is hard to parse ;), We used to have 31GiB I think.
You could also be more specific in the commit log about the openshift versions this is going to work with, so that people find answers to questions like "I'm generating 4.4 images, will this negatively impact me?".
We are extracting openshift-install from released bare metal container images, do these already have the required change?

createdisk.sh
```diff
@@ -64,11 +54,10 @@ function resize_and_sparsify {
     fi
 
     guestfish --remote <<EOF
-add-drive $tmpFile
+add-drive $baseDir/$srcFile
 run
 luks-open $partition coreos-root
 mount /dev/mapper/coreos-root /
-xfs-growfs /
```
cfergeau, line 71: Maybe give slightly more details in the commit log about how this works? The installer provisions a VM with the correct disk size from the start, and on initial boot, RHCOS expands its filesystem to fills the whole disk, so we no longer need to do the resizing ourselves.

createdisk.sh
```diff
@@ -64,11 +54,10 @@ function resize_and_sparsify {
...
 zero-free-space /boot/
 EOF
     if [ $? -ne 0 ]; then
```
cfergeau, line 63: Just below that line, the error message is "Failed to resize and sparsify $baseDir/$srcFile, aborting", you can remove the "resize" mention  from it.

snc.sh
praveenkumar: > We are extracting openshift-install from released bare metal container images, do these already have the required change?

from master side it already have those changes, on 4.5 side it will be part of next rc since we just miss one day to get it merge on 4.5 branch.

cfergeau: I'd document this in the commit log so that we can have a clue why resizing no longer happens on 4.4

praveenkumar: I usually branch out snc for each openshift release so the user don't worry about what works and what not with master branch, I can add the note in commit.

PR #187, opened 2020-06-26
Add Dockerfile for openshift CI

praveenkumar:


images/openshift-ci/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# This Dockerfile is used by openshift CI
+# It builds an image containing snc code and nss-wrapper for remote deployments, as well as the google cloud-sdk for nested GCE environments.
+FROM scratch AS builder
+WORKDIR /code-ready/snc
+COPY . .
+
+FROM centos:7
```
cfergeau, line 7: centos:8? I'm only running snc.sh on rhel8/fedoras, so I don't know how well el7 works.

images/openshift-ci/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# This Dockerfile is used by openshift CI
```
cfergeau, line 1: Can this be named `Containerfile`?

images/openshift-ci/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+COPY --from=builder /code-ready/snc/images/openshift-ci/mock-nss.sh /bin/mock-nss.sh
+COPY --from=builder /code-ready/snc/images/openshift-ci/google-cloud-sdk.repo /etc/yum.repos.d/google-cloud-sdk.repo
+
+RUN yum update -y && \
+    yum install --setopt=tsflags=nodocs -y \
+    gettext \
+    google-cloud-sdk \
+    nss_wrapper \
+    openssh-clients && \
+    yum clean all && rm -rf /var/cache/yum/*
```
cfergeau, line 18: Have you looked how hard it would be to get a base image for this? 2nd time I see this code duplicated, and I think the first time I saw it, it was already copied and pasted from another place..

images/openshift-ci/Dockerfile
gbraad: Any particular reason? This remains consistent with: https://github.com/code-ready/crc/tree/master/images

images/openshift-ci/Dockerfile
gbraad: I also would like to see a consistent image got this between CRC and SNC

images/openshift-ci/google-cloud-sdk.repo
```diff
@@ -0,0 +1,8 @@
+[google-cloud-sdk]
+name=Google Cloud SDK
+baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el7-x86_64
```
gbraad, line 3: the reason to base on EL7 ?

images/openshift-ci/Dockerfile
gbraad: centos 7 works... so should centos 8, but anyhing particular reason?

images/openshift-ci/Dockerfile
cfergeau: We could rename the crc one if this works :) Hopefully OpenShift CI is not using `docker` but `podman`, so the `Dockerfile` name is misleading. But it's just a personal preference to see how far I can go without having `docker` on my machine.

images/openshift-ci/Dockerfile
cfergeau: If Praveen is not testing el7 on a regular basis, this means it's almost never tested on el7, so we could hit unexpected issues. In a way that's good, but given rhel7 age, I'm not sure it's really worth making sure el7 works in addition to el8.

images/openshift-ci/google-cloud-sdk.repo
praveenkumar: we do have https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64 one but we do need to try that.

images/openshift-ci/Dockerfile
praveenkumar: To get a base image we need to maintain it, and one of the conversation which I had with Alex they really want to have base image as minimal as posible so in this case if centos:7 have new tag then our stuff auto run on new image but if we create ourself a base image name it `foo` then if there is any update on `centos:7` which will be base for `foo` first need to be updated so  I am OK with duplication as long as it is not maintenance burden.

images/openshift-ci/Dockerfile
praveenkumar: This image just carry the snc code not going to run it on el7 but run it on GCP VM which is rhel8 so I don't care what we have here, if you folks want `8` then I can change it.

images/openshift-ci/Dockerfile
praveenkumar: Yes, I can rename it but ci-operator (used for openshift-ci) have `dockerfile_path` which required Dockerfile path we can try to change it but all the project under openshift org use `Dockerfile` so I just used the same, we can try to rename it and if CI fails then change it back to Dockerfile (but that would be a separate PR since we need this in to test) ?

cfergeau: Not worth spending time on it :)

images/openshift-ci/Dockerfile
cfergeau: I'd explain in the commit log how this image is going to be used then.

PR #194, opened 2020-06-30
Rework hiding of pull secret from stdout

cfergeau:
a572aab "Hide the pull secret from stdout" prevents the pull secret from
being shown on stdout, but at the cost of not showing
'${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret'
when the script is run with 'set -x' / 'set -o xtrace'.
This commit sets a dummy value in install-config.yaml which is then
replaced with the actual pull secret in a function which makes sure the
pull secret is never shown on stdout.

'set -x' output is now more explicit:

+ cp install-config.yaml crc-tmp-install-data
+ ./yq write --inplace crc-tmp-install-data/install-config.yaml baseDomain testing
+ ./yq write --inplace crc-tmp-install-data/install-config.yaml metadata.name crc
+ ./yq write --inplace crc-tmp-install-data/install-config.yaml 'compute[0].replicas' 0
+ ./yq write --inplace crc-tmp-install-data/install-config.yaml pullSecret @HIDDEN_PULL_SECRET@
+ replace_pull_secret crc-tmp-install-data/install-config.yaml
+ set +x

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -99,6 +99,15 @@ function run_preflight_checks() {
         echo "libvirt and DNS configuration successfully checked"
 }
 
+function replace_pull_secret() {
+        # Hide the output of 'cat $OPENSHIFT_PULL_SECRET_PATH' so that it doesn't
+        # get leaked in CI logs
+        set +x
+        local filename=$1
+        sed -i "s!@HIDDEN_PULL_SECRET@!$(cat $OPENSHIFT_PULL_SECRET_PATH)!" $filename
+        set -x
+}
+
 function apply_bootstrap_etcd_hack() {
         # This is needed for now due to etcd changes in 4.4:
         # https://github.com/openshift/cluster-etcd-operator/pull/279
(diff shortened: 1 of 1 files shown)


praveenkumar:
@cfergeau I am still not convince here about the justification of {YQ} command because in this PR we are now hiding the `sed` command which actually replacing the pull secret. So if the command fail we will anyway get the error output in both the implementation.

cfergeau:
Currently, someone looking at the logs will not even see that we are trying to add a `pullSecret` field to `install-config.yaml` while we try to do it, With this change, this will appear in the logs, as well as a call to a `replace_pull_secret` function. Then the logs don't show how the pull secret is actually replaced, but the logs still tells us what was being done. 
In other words, with this change, if I grep the logs for 'pull.*secret' I will see where the script changed the pull secret, without this change, I won't have any indication if/when the pull secret was set if I only look at the logs

PR #195, opened 2020-07-02
Misc changes to accommodate Ben's changes so bundle can be used as is. 

praveenkumar:
- https://github.com/bbrowning/snc/commit/6d416a5dca837ef2bf42c6269a2010a239caf965

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -19,13 +19,13 @@ metadata:
   name: crc
 networking:
   clusterNetwork:
-  - cidr: 10.128.0.0/14
+  - cidr: 10.116.0.0/14
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
   networkType: OpenShiftSDN
   serviceNetwork:
-  - 172.30.0.0/16
+  - 172.25.0.0/16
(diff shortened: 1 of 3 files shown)


install-config.yaml
```diff
@@ -19,13 +19,13 @@ metadata:
...
-  - cidr: 10.128.0.0/14
+  - cidr: 10.116.0.0/14
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
   networkType: OpenShiftSDN
   serviceNetwork:
-  - 172.30.0.0/16
+  - 172.25.0.0/16
 platform:
```
cfergeau, line 29: I assume these values were just picked "randomly"?

guillaumerose:
It might worth a release note when released with crc.

/lgtm

install-config.yaml
praveenkumar: yes these are randomly picked.

guillaumerose: we could choose a single IP block (large enough) and say explicitly in the documentation that this IP block is reserved for crc. 
Inside this block, we can do what we want.

PR #196, opened 2020-07-02
Add VALID_CERT bool env to validate the cert

praveenkumar:
In the CI environment we don't really need to validate the cluster
kubelet cert and also current cert rotation method is not work properly
so with this env we can choose to validate or not the cluster certs.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -253,7 +253,10 @@ VM_PREFIX=${CRC_VM_NAME}-${random_string}
 # leeway regarding when we run the check with respect to rotation time
 if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem; then
     echo "Certs are not yet rotated to have 30 days validity"
-    exit 1;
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -247,15 +247,17 @@ if [ -z $random_string ]; then
...
-    exit 1;
+# Only validate the cert expire time if VALIDATE_CERT is set to true
+if [ "${VALIDATE_CERT:-false}" = true ]; then
+    # First check if cert rotation happened.
+    # Initial certificate is only valid for 24 hours, after rotation, it's valid for 30 days.
+    # We check if it's valid for more than 25 days rather than 30 days to give us some
+    # leeway regarding when we run the check with respect to rotation time
+    if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem; then
+        echo "Certs are not yet rotated to have 30 days validity"
+        exit 1;
```
cfergeau, line 258: I'd keep the check so that we have a warning in the logs that the certificates did not rotate. And I'd actually consider reusing `SNC_NON_FATAL_PREFLIGHT_CHECKS` which is introduced in the preflight commit as the cert validity check is arguably some kind of preflight check.
        fi`

PR #200, opened 2020-07-03
Add OWNERS

guillaumerose:
Required to please the openshift-ci bot.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,6 @@
+approvers:
+  - cfergeau
+  - praveenkumar
+  - anjannath
+  - guillaumerose
+  - gbraad


OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+  - cfergeau
+  - praveenkumar
+  - anjannath
+  - guillaumerose
+  - gbraad
```
praveenkumar, line 6: :+1:

praveenkumar: @guillaumerose I think you also need reviewers details to this file.

guillaumerose: I took the example from https://github.com/tektoncd/cli/blob/master/OWNERS. It seems to work fine for them.

praveenkumar: @guillaumerose alright. looks like the bot will not able to assign the reviewer if it is not there but that's OK since we don't want the bot to be that intelligent.

PR #201, opened 2020-07-03
Use `oc scale` instead of `oc patch` for scaling down ingress controller

guillaumerose:
According to the docs, it's the correct way to scale reduce the number of routers. See https://github.com/openshift/cluster-ingress-operator#managing.

For me, `oc patch` still left a broken pod, `oc scale` correctly remove it.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -405,7 +405,7 @@ ${OC} delete statefulset,deployment,daemonset --all -n openshift-kube-storage-ve
 ${OC} delete apiservice v1beta1.metrics.k8s.io
 
 # Scale route deployment from 2 to 1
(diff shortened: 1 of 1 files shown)


snc.sh
```diff
@@ -405,7 +405,7 @@ ${OC} delete statefulset,deployment,daemonset --all -n openshift-kube-storage-ve
 ${OC} delete apiservice v1beta1.metrics.k8s.io
 
 # Scale route deployment from 2 to 1
-${OC} patch --patch='{"spec": {"replicas": 1}}' --type=merge ingresscontroller/default -n openshift-ingress-operator
+${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-operator
```
praveenkumar, line 408: :+1:

PR #202, opened 2020-07-04
Use threads option for xz compression

praveenkumar:
Using the threads option for xz we can reduce the bundle creation
time significantly. This will also help us to stop timeout for CI.
```
$ time xz --keep --threads=0 crc.qcow2
real    4m36.255s
user    54m17.698s
sys     0m3.416s

$ time xz --keep crc.qcow2
real    34m48.959s
user    34m46.237s
sys     0m2.710s

$ time xz --threads=0 crc_libvirt_4.4.8.tar
real	15m8.081s
user	107m18.482s
sys	0m18.774s
```

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -380,7 +380,7 @@ create_qemu_image $libvirtDestDir
 
 copy_additional_files $1 $libvirtDestDir
 
-tar cJSf $libvirtDestDir.$crcBundleSuffix $libvirtDestDir
+tar cSf - $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix
 
 # HyperKit image generation
 # This must be done after the generation of libvirt image as it reuse some of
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -380,7 +387,7 @@ create_qemu_image $libvirtDestDir
 
 copy_additional_files $1 $libvirtDestDir
 
-tar cJSf $libvirtDestDir.$crcBundleSuffix $libvirtDestDir
+create_tarball $libvirtDestDir
```
cfergeau, line 390: This can be `tar cSf - $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix`

praveenkumar: @cfergeau is there any improvement using this over executing the commands one by one, just curious :)

cfergeau: Main improvement is that you don't need an intermediate file that you need to remove afterwards. But not a big difference between the 2.

PR #204, opened 2020-07-06
Use unsupported config map to force the cert rotation

praveenkumar:


snc.sh
```diff
@@ -227,31 +227,57 @@ function create_pvs() {
...
 
-    # Wait until bootstrap csr request is generated with 5 min timeout
-    timeout 300 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
-    ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
+    if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem; then
+        preflight_failure "Certs are not yet rotated to have 30 days validity"
+    fi
 
-    delete_operator "daemonset/kubelet-bootstrap-cred-manager" "openshift-machine-config-operator" "k8s-app=kubelet-bootstrap-cred-manager"
+    # go back to normal certrotation setting (remove unsupported-cert-rotation-config)
```
guillaumerose, line 279: What do we loose if we stay in this mode ? Certs will always be longer than usual? I guess it's a good thing no?

snc.sh
```diff
@@ -227,31 +227,57 @@ function create_pvs() {
...
+	else
+            break
+        fi
+	i=$[$i+1]
+    done
 
-    # Wait until bootstrap csr request is generated with 5 min timeout
-    timeout 300 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
-    ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
+    if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem; then
```
guillaumerose, line 275: With the new hack, I guess we could do a more wide check: look at all certs in /etc/kubernetes and check their date. (clearly not for this PR, just a remark).

snc.sh
```diff
@@ -227,31 +227,57 @@ function create_pvs() {
...
-    # Get the cli image from release payload and update it to bootstrap-cred-manager resource
-    cli_image=$(${OC} adm release -a ${OPENSHIFT_PULL_SECRET_PATH} info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=cli)
-    ${YQ} write kubelet-bootstrap-cred-manager-ds.yaml.in spec.template.spec.containers[0].image ${cli_image} >kubelet-bootstrap-cred-manager-ds.yaml
+    # Create the unsupported cert rotation config for a day (validity will increase by 30 times so it will become 30days)
+    ${OC} create -n openshift-config configmap unsupported-cert-rotation-config --from-literal='base=86400s'
 
-    ${OC} apply -f kubelet-bootstrap-cred-manager-ds.yaml
-    rm kubelet-bootstrap-cred-manager-ds.yaml
+    # Delete the pods to have cert rotation
+    ${OC} delete pods --all -A --force --grace-period=0
```
guillaumerose, line 234: Here I guess you want to delete pods that load the configmap, not all of them. That's fine but maybe it can save some time.

Consumers:
cluster-kube-apiserver-operator
cluster-kube-controller-manager-operator 

Am I wrong? (Don't bother changing this, I am just trying to understand)

praveenkumar: @guillaumerose I am not sure if only those are the consumers but this is something I got from the openshift QE folks so used this as it is but we can try that as separate experiment.

snc.sh
praveenkumar: sure, we can do that.

guillaumerose:
I took the bundle from the CI and the bundle is really better.

```
find /etc/kubernetes -iname "*.crt" -exec /bin/sh -c "echo {}; openssl x509 -checkend 2160000 -noout -in  {}" \; | grep "will expire" -B1 
```

returns only the kubelet serving cert to be expired. Compared to before it is a huge improvement.

snc.sh
```diff
@@ -227,31 +227,57 @@ function create_pvs() {
...
+     done
+     echo "API server is up"
+
+    # Force the rotation for secrets
+    ${OC} get secret -A -o json | ${JQ} -r '.items[] | select(.metadata.annotations."auth.openshift.io/certificate-not-after" | .!=null and fromdateiso8601<='$( date --date='+1year' +%s )') | "-n \(.metadata.namespace) \(.metadata.name)"' | \
+        xargs -n3 ${OC} patch secret -p='{"metadata": {"annotations": {"auth.openshift.io/certificate-not-after": null}}}'
 
-    # Wait for 5 min to make sure cluster is stable again.
+    # Delete the current csr signer to get new request.
+    ${OC} delete secrets/csr-signer-signer secrets/csr-signer -n openshift-kube-controller-manager-operator || true
```
guillaumerose, line 247: These secrets are also affected by the action just before. Maybe it is not needed?

PR #205, opened 2020-07-06
[WIP] move cvo override between `create manifest` and `create cluster`

guillaumerose:
It works but I'm not sure this is the right path to follow. 

Pros:
* Unwanted operators never run
* It's what we aim with cluster profile
* Simpler snc script

Cons:
* Still leave broken contents (ns with services but not the operator, etc.)
* CRDs created by operators are not present (PrometheusRule and ServiceMonitor for example)
* It doesn't speed up snc
* Override list is hard to maintain.

### DIFF
--- a/cvo_override.yaml
+++ b/cvo_override.yaml
@@ -13,11 +13,6 @@
     unmanaged: true
   - kind: Deployment
     group: apps/v1
-    name: machine-api-operator
-    namespace: openshift-machine-api
-    unmanaged: true
-  - kind: Deployment
-    group: apps/v1
     name: cluster-autoscaler-operator
     namespace: openshift-machine-api
     unmanaged: true
--- a/cvo_override.yaml
+++ b/cvo_override.yaml
@@ -8,11 +8,6 @@
     unmanaged: true
   - kind: Deployment
     group: apps/v1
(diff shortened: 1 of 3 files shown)


praveenkumar:
Another con to this approach.

If a user want to enable monitoring or any disable operator, it is quite easy override list till now to remove required component from the override list and CVO again start to bring that operator to the cluster ( https://code-ready.github.io/crc/#starting-monitoring-alerting-telemetry_gsg ). I think with this PR it would be very difficult.

guillaumerose:
We could also introduce a crc command to do it. I just added in the overrides:
* all PrometheusRule
* all ServiceMonitor
* some ClusterOperators

It's a just a big list but we could automate that by listing PrometheusRule and ServiceMonitor in the release image.

praveenkumar:
> It's a just a big list but we could automate that by listing PrometheusRule and ServiceMonitor in the release image.

@guillaumerose you mean using https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusterversion.md#setting-objects-unmanaged ( oc adm release extract ..) and parse the Rules ?

guillaumerose:
Not better than before, closing.

PR #206, opened 2020-07-08
Use raw image on macos

cfergeau:
Newer macos release uses a filesystem which supports sparse files. qcow2
support in hyperkit has known issues, and using it is not recommended. A
sparse raw file should not be much different from a qcow2 file with
respect to disk usage, so we can switch to using this.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -155,15 +155,20 @@ function generate_hyperkit_directory {
     cp $srcDir/kubeadmin-password $destDir/
     cp $srcDir/kubeconfig $destDir/
     cp $srcDir/id_rsa_crc $destDir/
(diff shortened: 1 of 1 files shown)


praveenkumar:
> Nice! Is .img the right extension? I recall Docker Desktop using .raw.

@guillaumerose As per https://en.wikipedia.org/wiki/IMG_(file_format) looks like it is raw disk format extension.

cfergeau:
> > Nice! Is .img the right extension? I recall Docker Desktop using .raw.
> 
> @guillaumerose As per https://en.wikipedia.org/wiki/IMG_(file_format) looks like it is raw disk format extension.

Yes, I think  the tools built around qemu use `.img` as the extension, quick check of `qemu-img` and `virt-install` man pages show that they use `.img` in their examples, and no results when searching for `.raw`

PR #209, opened 2020-07-09
Enable Multi-Architecture Support

mtarsel:
Enable support for both ppc64le and s390x using what’s available from the mirror. Creates a global $ARCH variable and then download proper Openshift binaries from the official mirror. Remove the local arch variable added by preflight check introduced from commit b03abd4. Use proper arch name for yq binary - x86_64 is labeled as amd64.  Also includes the extra step in README to change value of architecture key in install-config.yaml.

@Prashanth684

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -3,13 +3,13 @@
 apiVersion: v1
 baseDomain: testing
 compute:
-- architecture: amd64
+- architecture: 
   hyperthreading: Enabled
   name: worker
   platform: {}
   replicas: 1
 controlPlane:
-  architecture: amd64
+  architecture: 
   hyperthreading: Enabled
   name: master
   platform: {}
--- a/snc.sh
+++ b/snc.sh
@@ -12,9 +12,16 @@ XMLLINT=${XMLLINT:-xmllint}
 YQ=${YQ:-yq}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/clients/ocp}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
+ARCH=$(uname -m)
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+
+yq_ARCH=${ARCH}
+# yq and install_config.yaml use amd64 as arch for x86_64
+if [ "${ARCH}" == "x86_64" ]; then
+    yq_ARCH="amd64"
+fi
 
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 # Otherwise use the tagged version if available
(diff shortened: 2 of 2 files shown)


snc.sh
```diff
@@ -54,17 +61,9 @@ function run_preflight_checks() {
                 preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
         fi
 
-        # check if libvirtd has access to an x86_64 hypervisor
-        local arch
-        arch=$(virsh -c ${LIBVIRT_URI} capabilities | ${XMLLINT} --xpath '/capabilities/host/cpu/arch/text()' -)
```
Prashanth684, line 59: maybe keep this check anyway for x86,ppc64le and s390x so other arches like arm will be excluded (till we have support)

README.md
```diff
@@ -4,6 +4,7 @@
 - Make sure the one time setup is satisfied. (https://github.com/openshift/installer/blob/master/docs/dev/libvirt/README.md#one-time-setup)
 - Clone this repo `git clone https://github.com/code-ready/snc.git`
 - `cd <directory_to_cloned_repo>`
+- Change the architecture in the [install-config.yaml](install-config.yaml)
 - `./snc.sh`
 
 ## How to create disk image?
```
Prashanth684, line 10: maybe mention here that disk image creation is not supported for arches other than x86_64 ?

snc.sh
```diff
@@ -12,10 +12,17 @@ XMLLINT=${XMLLINT:-xmllint}
...
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/clients/ocp}
+ARCH=$(uname -m)
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 
+# Point to proper directory on mirror based on architecture
+if [ "${ARCH}" == "ppc64le" ] || [ "${ARCH}" == "s390x" ]; then
+    MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
```
cfergeau, line 21: https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/ also exists, so I would unconditionnally use `https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp`

snc.sh
```diff
@@ -54,17 +61,9 @@ function run_preflight_checks() {
                 preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
         fi
 
-        # check if libvirtd has access to an x86_64 hypervisor
-        local arch
-        arch=$(virsh -c ${LIBVIRT_URI} capabilities | ${XMLLINT} --xpath '/capabilities/host/cpu/arch/text()' -)
-        if [ "${arch}" != "x86_64b" ]; then
```
cfergeau, line 60: ugh, `x86_64b` looks like some debug leftover?

snc.sh
cfergeau: Yeah, would be good to keep such a warning (or even failure, as it's only a warning to avoid blocking ppc64 efforts)

README.md
cfergeau: Why isn't this step automated? And what's wrong with `create_disk.sh`?

snc.sh
gbraad: Please add a switch like structure to determine the ARCH and if supported.

gbraad:
Hold off till after the release

README.md
mtarsel: I was able to create a disk image tarball on ppc64le. I will automate this with yq and remove the line in README

snc.sh
praveenkumar: Looks like ppc64le arc is not upto date, if you check x86_64 you can find the 4.5 bits but not present for ppc64 side :(

Prashanth684: yeah i think for 4.4. and 4.5 ppc64le/s390x will GA later than x86 and 4.6 onward they should release at the same time.

README.md
Prashanth684: when trying it a while ago - i hit this issue where virt-resize was renaming the partitions. For example on s390x, there are only 2 partitions with id 1 and 4. when copying, virt-resize changed that to 1 and 2 causing issues. In the latest commit I see that logic removed, so it should work now!

mtarsel:
Split the changes into 3 commits. Wasn't really sure what the echos should say in the case statement for the arch's support but hope it works.

praveenkumar:
@mtarsel Can you please squash it and then push since all individual commits doesn't make sense here, there are all the changes needed to have the other platform (ppc/s390) support.

README.md
cfergeau: If create_disk.sh was not working yet, I would not necessarily see this as a blocker for that PR, but this would need to be documented.

install-config.yaml
```diff
@@ -3,13 +3,13 @@
 apiVersion: v1
 baseDomain: testing
 compute:
-- architecture: amd64
+- architecture:
```
cfergeau, line 6: I think we leave default values in `install-config.yaml`, so you can keep the `amd64` there.

snc.sh
```diff
@@ -321,6 +327,8 @@ sudo systemctl reload NetworkManager
 
 # Create the INSTALL_DIR for the installer and copy the install-config
 rm -fr ${INSTALL_DIR} && mkdir ${INSTALL_DIR} && cp install-config.yaml ${INSTALL_DIR}
+${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml compute[0].architecture ${ARCH}
+${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml controlPlane.architecture ${ARCH}
```
praveenkumar, line 331: looks like for `x86_64` valid value is `amd64` otherwise it fails our current tests fails because of this.

```
"failed to fetch Master Machines: failed to load asset \"Install Config\": invalid \"install-config.yaml\" file: [controlPlane.architecture: Unsupported value: \"x86_64\": supported values: \"amd64\", \"s390x\", \"ppc64le\", compute[0].architecture: Unsupported value: \"x86_64\": supported values: \"amd64\", \"s390x\", \"ppc64le\"
```

cfergeau: indeed, this can also be seen in the install-config.yaml changes in this PR

mtarsel: fixed with $yq_ARCH variable in squashed commit

praveenkumar:
The e2e-snc failure is not related to this PR. merging it.

PR #213, opened 2020-07-17
Update the mirror url to use latest-4.5

praveenkumar:
We started using candidate for 4.5 because that time latest link
was not available and now it is so switching to latest so our CI
also can test what is comming in latest-4.5 side.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -33,7 +33,7 @@ else
     if test -n "${OPENSHIFT_RELEASE_VERSION}"; then
         echo "Using release ${OPENSHIFT_RELEASE_VERSION} from local Git tags"
     else
-        OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/candidate/release.txt | sed -n 's/^ *Version: *//p')"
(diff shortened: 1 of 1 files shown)


cfergeau:
Isn't `candidate-4.5` closest to what we want to test regularly, as this is what our next release would be based on? I don't really know the difference between candidate and latest :)

guillaumerose:
I don't know if it's possible with the CI but we could also choose to create a 4.5 branch now and put this commit in.

praveenkumar:
> Isn't candidate-4.5 closest to what we want to test regularly, as this is what our next release would be based on? I don't really know the difference between candidate and latest :

 @cfergeau latest-<version> and stable-<version> are what are part of stable channel and candidate is which is not out as part of stable channel and stress testing is still in progress and we always release based on the latest or stable. we only do the release with candidate when and upcoming release (like this time 4.5) is available to mirror but not part of stable yet (going to be soon). All the z stream releases we do using stable/latest till now.

> I don't know if it's possible with the CI but we could also choose to create a 4.5 branch now and put this commit in.

@guillaumerose I would keep master with 4.5 till there is 4.6 candidate available and the do this switch.

PR #215, opened 2020-07-21
Cleanup packages and vmlinuz/initramfs files after bundle tarball creation

praveenkumar:


guillaumerose:
/lgtm

Are there some files we should add to .gitignore ?

PR #216, opened 2020-07-22
Modifications to support creating okd bundles

anjannath:
Introduce a flag for the scripts `--install-type=ocp|okd` to switch between okd and ocp bundle generation

Overview of the changes we need in snc.sh:
- Do not remove the pull-secret config and file, as we inject a dummy pull-secret for okd installation
- Switch the mirror to download okd bits 

Overview of the changes we need in createdisk.sh:
- Change the ostree hash greping pattern, it starts with `fedora-coreos` in okd
- The root partition is not luks-encryption ready so modify guestfish commands accordingly.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -54,14 +54,24 @@ function sparsify {
             echo "${GUESTFISH} failed to start, aborting"
             exit 1
     fi
-
-    guestfish --remote <<EOF
+    
+    if [ ${OKD_INSTALL} == false ]; then
+        guestfish --remote <<EOF
 add-drive $baseDir/$srcFile
 run
 luks-open $partition coreos-root
 mount /dev/mapper/coreos-root /
 zero-free-space /boot/
 EOF
+    else
+        guestfish --remote <<EOF
+add-drive $baseDir/$srcFile
+run
+mount $partition /
+zero-free-space /boot/
+EOF
+    fi
+
     if [ $? -ne 0 ]; then
             echo "Failed to sparsify $baseDir/$srcFile, aborting"
             exit 1
@@ -92,9 +102,7 @@ function create_qemu_image {
     sudo chown $USER:$USER -R $destDir
     ${QEMU_IMG} rebase -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
     ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
-
     sparsify $destDir ${VM_PREFIX}-base ${CRC_VM_NAME}.qcow2
-
     # Before using the created qcow2, check if it has lazy_refcounts set to true.
     ${QEMU_IMG} info ${destDir}/${CRC_VM_NAME}.qcow2 | grep "lazy refcounts: true" 2>&1 >/dev/null
     if [ $? -ne 0 ]; then
@@ -202,10 +210,28 @@ QEMU_IMG=${QEMU_IMG:-qemu-img}
 VIRT_FILESYSTEMS=${VIRT_FILESYSTEMS:-virt-filesystems}
 GUESTFISH=${GUESTFISH:-guestfish}
 DIG=${DIG:-dig}
+# OKD_INSTALL: variable to switch between ocp/okd bundle creation
+OKD_INSTALL=false
 
-if [[ $# -ne 1 ]]; then
+if [[ $# -eq 0 ]]; then
    echo "You need to provide the running cluster directory to copy kubeconfig"
    exit 1
+fi
+
+if [[ $# -eq 2 ]]; then
+    case $2 in
+        --install-type=okd)
+            echo "OKD install requested"
+            OKD_INSTALL=true
+            ;;
+        --install-type=ocp)
+            echo "Doing default OCP install"
+            ;;
+        *)
+            echo "To switch between OKD/OCP install, Use flag '--install-type=ocp|okd'"
+            exit 1
+            ;;
+    esac
 fi
 
 if ! which ${JQ}; then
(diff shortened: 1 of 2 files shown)


createdisk.sh
```diff
@@ -54,14 +54,24 @@ function sparsify {
...
 run
 luks-open $partition coreos-root
 mount /dev/mapper/coreos-root /
 zero-free-space /boot/
 EOF
+    else
+        guestfish --remote <<EOF
+add-drive $baseDir/$srcFile
+run
+mount $partition /
```
praveenkumar, line 70: @anjannath if it is same partition then same might work for RCOS we can try with `mount $partition` on RCOS also.

createdisk.sh
```diff
@@ -274,7 +300,12 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman tag $certImage open
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl images -q | xargs -n 1 sudo crictl rmi 2>/dev/null'
 
 # Replace pull secret with a null json string '{}'
-${OC} --kubeconfig $1/auth/kubeconfig replace -f pull-secret.yaml
+# If its an OKD install we don't need to remove as that's not a real pull-secret
+if [ ${OKD_INSTALL} == false]; then
+    ${OC} --kubeconfig $1/auth/kubeconfig replace -f pull-secret.yaml
+    # Remove pull secret from the VM
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
+fi
```
praveenkumar, line 308: This can be ignored in case of FCOS because there is no pull secret but need to check if there is configmap for same. Remove pull secret from the VM can be ignored because it just delete the file if exist if not it will just run the `rm` command and do nothing.

createdisk.sh
```diff
@@ -329,7 +364,11 @@ until ping -c1 api.${CRC_VM_NAME}.${BASE_DOMAIN} >/dev/null 2>&1; do
 done
 
 # Get the rhcos ostree Hash ID
-ostree_hash=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline | grep -oP "(?<=rhcos-).*(?=/vmlinuz)"')
+if [ ${OKD_INSTALL} == false ]; then
```
praveenkumar, line 367: I would just do something like.
```bash
ostree_dir=rhcos
if [ ${OKD_INSTALL} ]; then
ostree_dir= fedora-coreos
fi

ostree_hash=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline | grep -oP "(?<=${ostree_dir}-).*(?=/vmlinuz)"')

 ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${ostree_dir}-${ostree_hash}/* $1
```

createdisk.sh
```diff
@@ -296,21 +327,25 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable io.podman
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crictl pods -q)'
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'for i in {1..3}; do sudo crictl rmp $(sudo crictl pods -q) && break || sleep 2; done'
 
-# Remove pull secret from the VM
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
 
-# Download the hyperV daemons and libvarlink-util dependency on host
-mkdir $1/packages
-sudo yum install -y --downloadonly --downloaddir $1/packages hyperv-daemons libvarlink-util
+if [ ${OKD_INSTALL} == false ]; then
```
praveenkumar, line 331: This can be replaced with `if ! ${OKD_INSTALL}`

createdisk.sh
anjannath: But RHCOS still has luks encryption, its only fcos that doesn't have it

createdisk.sh
anjannath: Both the file and config map exists in FCOS too, we just don't need to remove them, as those are dummy, you still need a valid json in the install-config for the pull-secret.

createdisk.sh
anjannath: Thanks, so all the `if [ ${OKD_INSTALL} == false ]` can be replaced with this.. _shellNoob_ 🙅

anjannath: so `if [ ! {OKD_INSTALL} ]` evaluates to true if a value is set for `OKD_INSTALL` doesn't matter `false` or `true` i don't want that, so we have to explicitly check with `if [  {OKD_INSTALL} == false ]`

createdisk.sh
cfergeau: I believe I would add one layer of indirection here,
```
if [ ${OKD_INSTALL} == false ]; then
   use_luks=false
else
  use_luks=true
fi
```
and then check `use_luks` in the guestfish code.
You probably can split the `guestfish --remote` invocation to keep the common part
```
guestfish --remote <<EOF
add-drive $baseDir/$srcFile
run
EOF

if [ -n "${use_luks}" ]; then
    guestfish --remote luks-open $partition coreos-root
fi
guestfish --remote <<EOF
mount /dev/mapper/coreos-root /
zero-free-space /boot/
EOF
```
```

File: createdisk.sh (hunk 4 above)
Comment (cfergeau) [reply createdisk.sh:331]: I think I would also favour doing tests for variable empty/not empty rather than checking explicitly for `true`/`false`

File: snc.sh [hunk 5]
```diff
@@ -333,8 +359,13 @@ ${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml controlPlane.architectu
...
 ${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml compute[0].replicas 0
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret "@HIDDEN_PULL_SECRET@"
-replace_pull_secret ${INSTALL_DIR}/install-config.yaml
+# if OKD install type was requested inject dummy pull-secret
+if [ ${OKD_INSTALL} == false ]; then
+    ${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret "@HIDDEN_PULL_SECRET@"
+    replace_pull_secret ${INSTALL_DIR}/install-config.yaml
+else
+    ${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret '{"auths":{"fake":{"auth": "bar"}}}'
+fi
```
cfergeau, line 368: Might be more elegant (or too convoluted??) to do:
```
${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret '{"auths":{"fake":{"auth": "bar"}}}'    
if [ ${OKD_INSTALL} == false ]; then
    replace_pull_secret ${INSTALL_DIR}/install-config.yaml
fi
```
and to change the @HIDDEN_PULL_SECRET@ pattern which is currently used in `replace_pull_secret`. Maybe `replace_pull_secret` could be always called, and have an early return when it's an okd install?

cfergeau:
In this PR or the other similar PR, is there any indication in `crc-bundle-info.json` that the bundle is an openshift bundle or an okd one?

anjannath:
> In this PR or the other similar PR, is there any indication in crc-bundle-info.json that the bundle is an openshift bundle or an okd one?

No there's no explicit indication, as part of both the PRs. However the `openshiftVersion` will contain the _okd_ string as the versions are like that, e.g `4.5.0-0.okd-2020-08-12-020541`

praveenkumar:
/close

Since we have other PR in for same.

PR #217, opened 2020-07-22
Disable the network time sync for the bundle

praveenkumar:
This will help us to test the cert rotation effectively since
we can then on the crc codebase side enable it or keep it disabled.


anjannath:
just curious, are we going to add a flag to start, or config for not-starting this from crc?

cfergeau:
For what it's worth, having to change both `snc` and `crc` for an integration test does not seem so great, an integration test should be able to test an unmodified binary/bundle, not something specifically tweaked for the test needs.

PR #219, opened 2020-07-31
Use oc binary from the payload instead user provided

praveenkumar:
If user is using older version of oc binary then some of the
functionality doesn't work as expected due to server and client
mismatch.

```
$ oc adm certificate approve csr-chcck
No resources found
error: no kind "CertificateSigningRequest" is registered for version "certificates.k8s.io/v1" in scheme "k8s.io/kubectl/pkg/scheme/scheme.go:28"
```

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -297,6 +297,15 @@ if test -z ${OPENSHIFT_INSTALL-}; then
     OPENSHIFT_INSTALL=./openshift-install
 fi
 
+# Extract oc binary from the payload and use it for all following operations
+if ! test -f oc; then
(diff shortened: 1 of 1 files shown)


cfergeau:
Seems the preexisting logic around finding the `oc` binary could be simplified if we always want to use this binary. Actually, isn't https://github.com/code-ready/snc/blob/master/snc.sh#L245-L253 conflicting with this new code https://github.com/code-ready/snc/blob/master/snc.sh#L299-L308 ?

cfergeau:
See https://github.com/code-ready/snc/pull/221

PR #220, opened 2020-07-31
Remove unused code

praveenkumar:
This code used to just show an message in case the error was
because of MCo and it is already fixed long time ago and we now
also exit from the script as soon as any command fails.


cfergeau:
Looks good to me, I was slightly confused by the log message initially, you mention that 'we now
also exit from the script as soon as any command fails.', maybe also be explicit that this means we'll never trigger the `if [ $? -ne 0 ]; then` condition.

PR #221, opened 2020-08-04
Never consider 'oc' binaries present in $PATH

cfergeau:
Some distros only have packages for old 'oc' versions (3.11), which are
missing some of the features we need. Since we don't currently have a
way of checking version of the 'oc' binary, we'll just never use 'oc'
binaries from $PATH, and assume the version snc cached is new enough.

This reverts most of f037c89dc6 as this introduced unneeded complexity
by using 2 different sources for oc binaries (either downloaded from
mirror.openshift.com, or extracted from the appropriate image). This
commit only downloads the oc binary from mirror.openshift.com as
extracting from the 'cli-artifacts' image was slower for me, and using
'oc' in order to get an 'oc' binary is not always possible (this assumes
we already have this binary available).

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -7,7 +7,7 @@ export LANG=C
 
 INSTALL_DIR=crc-tmp-install-data
 JQ=${JQ:-jq}
-OC=${OC:-oc}
+OC=${OC:-./oc}
 XMLLINT=${XMLLINT:-xmllint}
 YQ=${YQ:-yq}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
@@ -244,10 +244,8 @@ function delete_operator() {
 }
 
 # Download the oc binary if not present in current directory
-if ! which ${OC}; then
-    if [[ ! -e oc ]] ; then
-        curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-linux-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar zx oc
-    fi
+if [[ ! -e ${OC} ]] ; then
+    curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-linux-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar zx oc
     OC=./oc
 fi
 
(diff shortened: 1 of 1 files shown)


snc.sh
```diff
@@ -297,14 +295,6 @@ if test -z ${OPENSHIFT_INSTALL-}; then
...
 fi
 
-# Extract oc binary from the payload and use it for all following operations
-if ! test -f oc; then
-    echo "Extracting oc binary from OpenShift payload image"
-    oc_image=$(${OC} adm release -a ${OPENSHIFT_PULL_SECRET_PATH} info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=cli-artifacts)
-    ${OC} image -a ${OPENSHIFT_PULL_SECRET_PATH} extract ${oc_image} --confirm --path /usr/bin/oc:.
-    chmod +x oc
-    OC=./oc
-fi
```
praveenkumar, line 307: @cfergeau I don't want to remove this part because in the CI where you are testing the PR and not released/dev-preview bits then oc version should be same what is part of payload.

cfergeau:
Quick summary of a lengthy private discussion: The initial change came about because of errors happening when using oc 4.5.x with a 4.6 cluster (see https://github.com/code-ready/snc/commit/f037c89dc609209953039b3f95be402528c8ebc4 for a few more details).

The image CI uses has an `oc` binary in its path, and most likely does not reuse the same `snc` directory multiple times, so the new code triggers and extracts the correct `oc` binary. Note that one needs a functional `oc` binary in order to be able to get the latest `oc` binary.

In most cases, the same could be achieved by overriding `MIRROR` to https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp-dev-preview/ , but this is not good enough for CI which uses `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` and will most often test newer code than what is mirrored anyway.

The fix from commit f037c89dc will only work when you have an `oc` binary in your path, but no `oc` binary in the current directory. It also assumes that the `oc` binary in the path is new enough to have the `oc adm release` command (which is not the case when using `oc` from `origin-clients` on fedora).

With all that said, I'm not sure what the best way of handling this is, short of 
- downloading a 4.x `oc` binary once
- always extracting the `oc` binary from the `cli-artifacts` image (which would be annoying when testing something unrelated to this :-/ )

cfergeau:
> With all that said, I'm not sure what the best way of handling this is, short of
> 
>     * downloading a 4.x `oc` binary once
> 
>     * always extracting the `oc` binary from the `cli-artifacts` image (which would be annoying when testing something unrelated to this :-/ )

https://github.com/code-ready/snc/commit/fdfa567e  went with a variation on this since we were always downloading oc binaries matching the OpenShift release we are building for.

PR #222, opened 2020-08-06
Copy kubeconfig file to the VM

praveenkumar:
Right now in crc we copy the kubeconfig file from host to VM which
is used to execute oc commands from the VM, this commit will make
sure we do it as part of disk creation so that we don't need to do
it on crc side.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -315,6 +315,10 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
 # Remove the packages from VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/packages
 
+# SCP the kubeconfig file to VM
+${SCP} $1/auth/kubeconfig core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo mv /home/core/kubeconfig /opt/'
+
 # Shutdown and Start the VM after installing the hyperV daemon packages.
 # This is required to get the latest ostree layer which have those installed packages.
 sudo virsh shutdown ${VM_PREFIX}-master-0


guillaumerose:
Since this PR changes what is inside the VM, should we bump the version number here https://github.com/code-ready/snc/blob/master/crc-bundle-info.json.sample#L5 ?

If we do this we could add a `if bundle version > 1.1, don't copy` in this PR https://github.com/code-ready/crc/pull/1441/files#r465815467 and have  smooth transition. 

It also opens the door to support multiple version of the bundle in crc :)

createdisk.sh
```diff
@@ -315,6 +315,10 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
 # Remove the packages from VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/packages
 
+# SCP the kubeconfig file to VM
+${SCP} $1/auth/kubeconfig core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo mv /home/core/kubeconfig /opt/'
```
cfergeau, line 320: out of curiosity, why `/opt/` rather than another path? For example `/home/core/` where you put it initially?

praveenkumar: @cfergeau no reason, I just like to `opt` for option config file :)

praveenkumar:
> Something I was wondering about is if there are any security implications to this? Anyone who can access this file has admin access to the cluster, I don't know if there are cases when an 'unauthorized' person could get their hands on it?

@cfergeau yes, but to gain access to this file someone need to have vm or host access.

praveenkumar:
> It also opens the door to support multiple version of the bundle in crc :)

@guillaumerose that is still need to figure out, I am not in favor to support multiple bundle with crc right now, but @cfergeau might disagree with me.

createdisk.sh
cfergeau: Hmm this kind of collide with 'normal' use though https://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/opt.html

cfergeau:
> > It also opens the door to support multiple version of the bundle in crc :)
> 
> @guillaumerose that is still need to figure out, I am not in favor to support multiple bundle with crc right now, but @cfergeau might disagree with me.

You read my mind ;) I was disapointed to see that `crc start` with the 1.13 release followed by `crc stop`, and then a `crc start` attempt with git master (with a changed bundle version) did not work as it used to.

guillaumerose:
Instead of saying "we support bundles version 1.x", we could say "we support bundle 4.5.4. Older bundles may work, use them at your own risk" (with a big warning banner?).

praveenkumar:
@guillaumerose https://github.com/code-ready/crc/issues/697 have some discussion about supporting other bundle.

PR #225, opened 2020-08-25
Make console operator unmanaged

praveenkumar:
This will allow us scale console and download
pods to 1 to save some resources.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -232,3 +232,10 @@ retry ${OC} scale --replicas=1 deployment etcd-quorum-guard -n openshift-etcd
 
 # Set default route for registry CRD from false to true.
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
+
+# Unmanage openshift-console operator config
+${OC} patch console.operator.openshift.io cluster -p='{"spec": {"managementState": "Unmanaged"}}' --type merge
+
+# Scale console and download pods to 1
+${OC} scale --replicas=1 deployment downloads -n openshift-console
+${OC} scale --replicas=1 deployment console -n openshift-console


snc.sh
```diff
@@ -429,3 +429,10 @@ ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-operator
 
 # Set default route for registry CRD from false to true.
 ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
+
+# Unmanage openshift-console operator config
+${OC} patch console.operator.openshift.io cluster -p='{"spec": {"managementState": "Unmanaged"}}' --type merge
+
+# Scale console and download pods to 1
+${OC} scale --replicas=0 deployment downloads
+${OC} scale --replicas=0 deployment console
```
cfergeau, line 438: I'd expect `--replicas=1` rather than `=0` ?

cfergeau (commented):
Any idea on the amount of resources which are saved?

cfergeau:
> Any idea on the amount of resources which are saved?

Do you know about this? Also, what are the implications of marking the operator as unmanaged? I expect it has some undesirable side effects in some situations?

gorkem:
It should save 150 Mi Memory and 10m CPU.

praveenkumar:
> Also, what are the implications of marking the operator as unmanaged? I expect it has some undesirable side effects in some situations?

@cfergeau the only situation we need to test this out when proxy is place, other than that I don't think any undesirable state because before we moved to CVO managed all the operators were unmanaged and didn't hit any issue.

cfergeau:
> > Also, what are the implications of marking the operator as unmanaged? I expect it has some undesirable side effects in some situations?
> 
> @cfergeau the only situation we need to test this out when proxy is place, other than that I don't think any undesirable state because before we moved to CVO managed all the operators were unmanaged and didn't hit any issue.

There must be some tradeoffs, otherwise openshift would not have this managed/unmanaged distinction.

praveenkumar:
> There must be some tradeoffs, otherwise openshift would not have this managed/unmanaged distinction.

@cfergeau yes the trade offs are it is not managed by CVO so if crash happen or any change (like proxy) which CVO managed are not affected on this.

cfergeau:
> r any change (like proxy) which CVO managed are not affected on this.

Ah, this means we might have to add some console/proxy specific code to crc? :-/

guillaumerose:
What if we wrote the smallest amount of code that:
* takes a release image,
* patches manifests we want (downsize deployment, eventually Srikrishna's changes, etc.)
* re-upload the image
* use it in snc. 

It would be cleaner and it will open the way for what we need upstream with the single node cluster profile.

praveenkumar:
@guillaumerose When you re-upload the image it will be no longer the release payload and when you file a BZ in case of issue we will get a pushback not using the release artifact and instead making modification and tempering the payload.

guillaumerose:
Good point but maybe it will be different now. We are trying really hard to push the cluster profile, they can see our commitment. We are not trying to do something bad, just having a decent experience for the user. 

I don't want to put everything in this release image but just what we would do with the profile and for sure, no hacks. Only changes that could go upstream.

gorkem:
I think it needs to be different with multiple profiles already proposed and exists in the wild.

guillaumerose:
With cluster capabilities, this will be just a boolean to turn on. Meanwhile, we can merged this.

/lgtm

guillaumerose:
After looking more in depth, the CVO is propagating proxy settings to the console operator. Then the operator propagates to the console container. 
It uses the proxy for 2 purposes:
* reach the oauth server (might be useful when the server is outside of the cluster)
* download helm charts.

Given this, it's better to wait for 4.8.

PR #226, opened 2020-08-26
Kill the child process when main snc process exit

praveenkumar:
Currently our etcd hack function is always running if the snc script
exit before bootstrap, this patch will make sure that as soon as snc
script exit then it's child process also exit (in our case it's etcd hack)

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -5,6 +5,9 @@ set -exuo pipefail
 export LC_ALL=C
 export LANG=C
 
+# kill all the child processes for this script when it exits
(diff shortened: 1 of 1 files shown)


snc.sh
```diff
@@ -5,6 +5,9 @@ set -exuo pipefail
 export LC_ALL=C
 export LANG=C
 
+# kill all the child process for this script when it exit
```
cfergeau, line 8: "child processes", "when it exits"

cfergeau (approved):

cfergeau (approved):

PR #229, opened 2020-09-03
Add auth operator hack to enable single apiserver

praveenkumar:
Recently there is an issue with auth operator which was causing
https://bugzilla.redhat.com/show_bug.cgi?id=1874713 one, to avoid
it we need to add unsupported config patch.

- https://github.com/openshift/cluster-authentication-operator/pull/318


guillaumerose (approved):
thanks

praveenkumar:
retest

PR #231, opened 2020-09-04
Load hv_sock module

guillaumerose:
Add support for Hyper-V socket. It does nothing on Linux and macOS but load the module on Windows.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -315,6 +315,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
 # Remove the packages from VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/packages
 
+# Adding Hyper-V vsock support
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
+    echo hv_sock > /etc/modules-load.d/vsock.conf
+EOF
+
 # SCP the kubeconfig file to VM
 ${SCP} $1/auth/kubeconfig core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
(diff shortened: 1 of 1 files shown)


snc.sh
```diff
@@ -323,6 +323,10 @@ ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} destroy cluster ${OPENSHIFT_INSTALL_EX
 rm id_rsa_crc* || true
 ssh-keygen -N "" -f id_rsa_crc -C "core"
 
+# Adding Hyper-V socket support
```
cfergeau, line 326: s/socket/vsock ?

cfergeau (commented):
Log is slightly confusing, it does not "load the module on Windows", but "load the module when the VM si being run by a HyperV hypervisor"
create_disk.sh already download some hyper-v specific rpms, maybe this should go there as well?

snc.sh
guillaumerose: changed

guillaumerose:
I moved the code to createdisk.sh. Although I don't see why we don't put most of the changes in snc.sh and keep createdisk.sh as minimal as possible.

praveenkumar:
> I don't see why we don't put most of the changes in snc.sh and keep createdisk.sh as minimal as possible.

@guillaumerose the motivation is all the VM related changes are part of `createdisk.sh` and all the openshift cluster related changes are part of `snc.sh` to keep us bit sane :)

createdisk.sh
```diff
@@ -315,6 +315,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
 # Remove the packages from VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/packages
 
+# Adding Hyper-V vsock support
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
+    echo hv_sock > /etc/modules-load.d/vsock.conf
+EOF
```
praveenkumar, line 321: :+1:

praveenkumar (approved):

PR #233, opened 2020-09-08
Add 5 min timeout for csr request

praveenkumar:
If the installer fails in the bootstrap process then csr request
is never going to approve but since we are using a loop without timeout
it makes the script to run an infinite loop.


cfergeau:
Do we have any way of detecting bootstrap failed VS bootstrap in progress?

praveenkumar:
> Do we have any way of detecting bootstrap failed VS bootstrap in progress?

@cfergeau no, bootstrap and provision steps are part of single `openshift-install create cluster` process :( so it fails if bootstraps fails or the provision fails after bootstrap.

PR #234, opened 2020-09-11
Add gvisor-tap-vsock daemon

guillaumerose:
https://github.com/code-ready/gvisor-tap-vsock/tree/v3

container will not run if a network interface is present.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -368,6 +368,14 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
     echo hv_sock > /etc/modules-load.d/vsock.conf
 EOF
 
+# Add gvisor-tap-vsock service
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
+  podman run -d --name=gvisor-tap-vsock --privileged --net=host -it quay.io/crcont/gvisor-tap-vsock:v3
+  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  systemctl daemon-reload
+  systemctl enable gvisor-tap-vsock.service
+EOF
+
 # SCP the kubeconfig file to VM
 ${SCP} $1/auth/kubeconfig core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/home/core/
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo mv /home/core/kubeconfig /opt/'


createdisk.sh
```diff
@@ -320,6 +320,15 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
     echo hv_sock > /etc/modules-load.d/vsock.conf
 EOF
 
+# Add gvisor-tap-vsock service
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
+  podman run -d --name=gvisor-tap-vsock --privileged --net=host -it quay.io/crcont/gvisor-tap-vsock:v1
+  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
```
praveenkumar, line 326: @guillaumerose why `restart-policy` is set to `no`, I think in case of container crashed then shouldn't it need restart?

guillaumerose: I don't want to pollute too much regular installation of SNC. So, if after 100 tries (with a sleep of 1s), it still can't connect, the container die. But it will still retry at the next startup.

guillaumerose: I changed the behaviour: now if some particular network interfaces are present, the container will shut down.

createdisk.sh
```diff
@@ -340,6 +340,15 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
     echo hv_sock > /etc/modules-load.d/vsock.conf
 EOF
 
+# Add gvisor-tap-vsock service
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
+  podman run -d --name=gvisor-tap-vsock --privileged --net=host -it quay.io/crcont/gvisor-tap-vsock:v3
+  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  systemctl daemon-reload
+  systemctl start gvisor-tap-vsock.service
+  systemctl enable gvisor-tap-vsock.service
```
praveenkumar, line 349: @guillaumerose I think as part of disk image creation you just need to enable it and everytime that this image boot will automatic start this service, or you also need to start this service on the current running instance?

guillaumerose: Yes you're right. Removing the start line.

createdisk.sh
gbraad: if the process/pod crashes?

gbraad (approved):
just wondering if a restart would be needed; when the process crashes, there is no network traffic possible, and this has happened before during tests.

guillaumerose:
Waiting for a final test of the bundle but currently the mirror is down.

createdisk.sh
guillaumerose: yes if it crashes, it's over. No way to contact the VM. 
I would expect that it doesn't happen and code doesn't panic and properly returns an error and fallback in the retry loop.

guillaumerose:
Working well now with the CI fixed.

PR #239, opened 2020-09-22
[POC] Performance tuning measures for single node OpenShift

spaparaju:
**Goal**

1. Figure out the reasons behind huge resource requirements for CRC OpenShift components
2. Demonstrate users with around 4 GB and 4 VCPUs can start using CRC on their personal laptops. This enables majority of the developer community with 8 GB laptops

**Approach**
Without changing the core payload of OCP used for a single node, these performance tuning changes were enabled (while creating the single node OpenShift disk image). Here is the mechanism to enable these performance tuning changes. The disk image for single node OpenShift ‘**does not contain**’ any the additional components (v1/alpha api, Pod presets, admission controllers) to enable performance tuning changes while creating the disk image.

**[WHICH] Performance tuning options tried out to reduce footprint of single node OpenShift**

1. At Kernel level: 
Disable huge memory pages: RHCOS defaults to ‘huge memory pages’. Disabling ‘huge memory pages’ and switching to typical 4K pages helps GoLang to gather more idle memory (more and smaller pages can be ready for garbage collection).  Rebooting node with  Linux kernel parameter ‘transparent_hugepage=never’ enables the switch to smaller memory pages (4K). 

2. At GoLang level:
Enable GoLang applications to speed up returning idle memory back to the Operating system (thru GoLang specific env. variables) :
{"name":"GOGC","value":"10"} (a new collection is triggered before newly allocated data in comparison to the current live data reaches this %)
{"name":"GODEBUG","value":"madvdontneed=1"} for faster releases to RSS of OS

3. Update resource requests/limits:
For Kube control plane components launched as static pods
For pods across all the openshift namespaces

4. Kubelet level changes:
Reduce ‘system-reserved’ resources (i.e. increase ‘allocatable’ resources)
Reduce max. Container log size 

**[HOW] Performance tuning options are applied** 
(_crc-perf-tuning.sh_ is the entry point for all these below indicated changes)

1. At Kernel level:
With grubby update to include kernel parameter to disable ‘huge pages’. (_apply-kernel-parameters.sh_)

2. At GoLang level:
(a) GOGC, GODEBUG environment variables are injected into the pod spec of pods (across all the openshift namespaces) through ‘Pod Presets’, a pre-compiled admission controller. ‘Pod presets’ require enabling support for ‘v1/alpha1(settings)’ at Kube API level.  (_enable-alpha-api.sh, trigger-podpresets.sh_)
(b) GOGC, GODEBUG environment variables are included in the manifest files of Kube control plane components (launched as static pods by Kubelet) (_update-kube-controlplane.sh_)

3. Update resource requests/limits:
(a) Enabling pre-compiled admission controller ‘MutatingAdmissionWebhook’ to inject required resource requests/limits values for pods across all the openshift namespaces (_admission-webhook.yaml_)
(b) Resources requests/limits are updated in the manifest files of Kube control plane components (launched as static pods by Kubelet) (_update-kube-controlplane.sh_)

4. Kubelet level changes:
Reduce ‘system-reserved’ resources by updating configuration file of kubelet (_update-kube-controlplane.sh_)


praveenkumar:
https://github.com/code-ready/snc/pull/239/commits/548de6c916ff87c0b4f9591b28d52d5093d4560a and https://github.com/code-ready/snc/pull/239/commits/548de6c916ff87c0b4f9591b28d52d5093d4560a should be a single commit.

tuning-crc-openshift-cluster/update-kube-controlplane.sh
```diff
@@ -0,0 +1,104 @@
...
+    cat final_manifest.json | ${JQ} -c > unformatted_final_manifest.json
+
+    ${SCP} -r unformatted_final_manifest.json core@master:/home/core/updated-kube-apiserver-manifest.yaml
+    ${SSH} core@master -- sudo cp /home/core/updated-kube-apiserver-manifest.yaml $1
+    ## Remove temp. files created
+    rm current_manifest.json updated_wth_memory_manifest.json updated_wth_cpu_manifest.json updated_with_env_manifest.json final_manifest.json unformatted_final_manifest.json
+}
+
+ update_kube_controller_manifests() {
+    ${SSH} core@master -- sudo cat $1  > current_manifest.json
```
praveenkumar, line 26: Not sure why you are using `core@master` instead use the ` core@api.${CRC_VM_NAME}.${BASE_DOMAIN} ` which is used in available scripts.

tuning-crc-openshift-cluster/update-kube-controlplane.sh
```diff
@@ -1,4 +1,4 @@
-
+SSH_KEYS_OF_MASTER_NODE=../id_rsa_crc
```
praveenkumar, line 1: This should be part of initial commit where `SSH` variable is added.

tuning-crc-openshift-cluster/crc-perf-tuning.sh
```diff
@@ -18,7 +18,9 @@ echo 'Enable Kube V1/alpha API .....'
 echo 'Update Kube control plane manifest files ......'
 ./update-kube-controlplane.sh
 echo 'Wait for Kube API to be available after the restart (triggered from updating the manifest files) .....'
-sleep 180
+while ! ${OC} get etcds cluster >/dev/null 2>&1; do
+    sleep 3
+done
```
praveenkumar, line 23: This should be part of initial commit where sleep was added.

gbraad:
These changes are more likely to be 'fixed' by the OpenShift team and the installer, as we shouldn't run CRC-specific changes. This will most likely result in a backfire when an issue is detected.

spaparaju:
Goal of this work is figure out WHAT (at RHCOS, GoLang and Kube) parameters to tune. The HOW part can be standardised with the ongoing work around 'Single node profile'

gbraad:
>  'Single node profile'

pretty much as I suggested on Slack. However, can therefore the title and description be changed as `CRC` into `POC` or `Single Node` ? as this would suggest more this is a proof of concept instead of actual changes we apply here.

cfergeau:
The series as it is is quite hard to follow. I'd try to organize it in a set of independent commits roughly following the outline in the PR description:

>    added prerequisite for creating CRC disk image
>    apply GOGC (madvdontneed=1), GOGC across the all CRC OpenShift pods (thru Podpresets)
>    apply initial memory, CPU requests thru Kubernetes MutatingWebhook for all the CRC OpenShift pods
>    update all the Kube. control plane components (launched as static pods: Kube API server, controller-manager, Etcd, Scheduler and Kubelet) with ENV variables (GOGC (madvdontneed=1), GOGC) and initial resource requests (memory, CPU)
>    Reduced the default resources (memory, CPU) used for system-reserved space which can not be allocated for CRC OpenShift components

For each commit, you can then give details about what / how this is achieved, what the tradeoffs are, ... (is this going to use more CPU, could this possibly become unstable during high loads, will the cluster complain about the low memory limits, ...)

spaparaju:
Seems easier to close this PR, creating a new PR with fewer commits.

PR #240, opened 2020-09-24
Add CVO override for cluster-node-tuning

praveenkumar:
Node tuning operator is used to tune the node as per workload and
it need tune profile to work and when it updated node needs reboot,
which is not possible for CRC so better to remove the resources
for tihs operator.

### DIFF
--- a/cvo_override.yaml
+++ b/cvo_override.yaml
@@ -56,3 +56,8 @@
     name: kube-storage-version-migrator-operator
     namespace: openshift-kube-storage-version-migrator-operator
     unmanaged: true
+  - kind: Deployment
+    group: apps/v1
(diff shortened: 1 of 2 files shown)


cfergeau (approved):
"and it need tune profile to work" -> "and it needs tune profile to work"
I don't understand exactly what you mean with this?
You could stop the sentence right after this though
```
it need tune profile to work and when it updated node needs reboot,
which is not possible for CRC so better to remove the resources
for tihs operator.
```
becomes
```
it needs tune profile to work. After it is updated, the node needs reboot,
which is not possible for CRC so better to remove the resources
for tihs operator.
```
(also `tihs` -> `this`)

PR #243, opened 2020-10-05
Make crc-bundle-info.json the first element in the tar archive

guillaumerose:
Using --sort=name moves crc-bundle-info.json to the first place in the
crcbundle archive. It allows crc and other programs to read the first
bytes of the archive to get the json metadata.
Programs need to read the first ~3KB of the bundle to get the json.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -412,7 +412,7 @@ create_qemu_image $libvirtDestDir
 
 copy_additional_files $1 $libvirtDestDir
 
-tar cSf - $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix
+tar cSf - --sort=name $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix
 
 # HyperKit image generation
 # This must be done after the generation of libvirt image as it reuse some of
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -412,7 +412,7 @@ create_qemu_image $libvirtDestDir
 
 copy_additional_files $1 $libvirtDestDir
 
-tar cSf - $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix
+tar cSf - --sort=name $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix
```
cfergeau, line 415: a bit fragile as this breaks as soon as we add a file which comes before `crc-bundle-info.json` in alphabetical order. The alternatives would be a bit more verbose I think :-/

cfergeau (approved):

createdisk.sh
guillaumerose: Yes sadly it's not easy. A more general solution would be to use zip (so that there is an index of files) but the archive would be 700MB bigger.

cfergeau: `tar cSf - $libvirtDestDir/crc-bundle-info.json --exclude $libvirtDestDir/crc-bundle-info.json $libvirtDestDir | xz --threads=0 >$libvirtDestDir.$crcBundleSuffix` seems to do the trick, but it's on the ugly side :)

PR #247, opened 2020-10-08
Add entry in NetworkManager conf to use dnsmasq as dns

praveenkumar:


cfergeau (approved):
Would be nice to check if there's already a  file with that content before doing it, but the same comment applies to the preexisting code so...

praveenkumar:
> Would be nice to check if there's already a file with that content before doing it, but the same comment applies to the preexisting code so...

cfergeau (approved):

PR #248, opened 2020-10-12
RFC: Optionally add a default pull secret in the bundle

guillaumerose:
This pull secret can be used to configure the cluster later on. It would
be primarily used by OKD to distribute a good default pull secret.

It can be also used by a large company, where developers are not allowed
to register on cloud.redhat.com to get a pull secret, to internally
distribute the bundle with a pre-defined pull secret. 

cc @cgruver 

Related to https://github.com/code-ready/snc/issues/237

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -149,6 +149,19 @@ function update_json_description {
         >$destDir/crc-bundle-info.json
 }
 
+function eventually_add_pull_secret {
+    local destDir=$1
+
+    if [[ -f "$BUNDLED_PULL_SECRET_PATH" ]]
+    then
+      cat "$BUNDLED_PULL_SECRET_PATH" > "$destDir/default-pull-secret"
+      cat $destDir/crc-bundle-info.json \
+          | ${JQ} '.clusterInfo.openshiftPullSecret = "default-pull-secret"' \
+          >$destDir/crc-bundle-info.json.tmp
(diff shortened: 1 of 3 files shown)


createdisk.sh
```diff
@@ -149,6 +149,20 @@ function update_json_description {
...
+    local destDir=$1
+
+    if [[ -f "$BUNDLED_PULL_SECRET_PATH" ]]
+    then
+      cat "$BUNDLED_PULL_SECRET_PATH" > "$destDir/default-pull-secret"
+      mv $destDir/crc-bundle-info.json $destDir/crc-bundle-info.json.tmp
+      cat $destDir/crc-bundle-info.json.tmp \
+          | ${JQ} '.clusterInfo.openshiftPullSecret = "default-pull-secret"' \
+          >$destDir/crc-bundle-info.json
+      rm $destDir/crc-bundle-info.json.tmp
```
cfergeau, line 162:
```
cat $destDir/crc-bundle-info.json ... >$destDir/crc-bundle-info.json.tmp
mv $destDir/crc-bundle-info.json.tmp $destDir/crc-bundle-info.json
```
would save the need for the `rm`, but that's not really important

cfergeau (approved):

createdisk.sh
guillaumerose: done, thanks!

Issue #214, opened 2020-07-17
Include a brief message in README about Red Hat Support


mtarsel:
With a blog going on OpenShift website, make mention in this repo that this is not an officially supported Red Hat product and the code comes "as-is".

Example:
`Please note the SNC and CRC projects are “as-is” on the Github repository. At this time, it is not a Red Hat supported solution.`

PR #249, opened 2020-10-13
Address Red Hat Support in README

mtarsel:
Update README.md to include: 

"Please note the SNC project is “as-is” on this Github repository. At this time, it is not an offically supported Red Hat solution."

Close issue #214


gbraad (changes requested):
`is provided "as-is"`

However, we have a whole document that talks about this, but so far we never got a conclusive answer to the support status, as "there is some support"

Issue #250, opened 2020-10-13
[ppc64le] Include default libvirt network


mtarsel:
On ppc64le, if the default Libvirt network is not present then it will produce the following error:

```
+ ./oc get etcds cluster
ERROR
ERROR Error: virError(Code=38, Domain=7, Message='unable to connect to server at '192.168.122.1:16509': Connection timed out')
ERROR
ERROR   on ../../tmp/openshift-install-718900460/main.tf line 1, in provider "libvirt":
ERROR    1: provider "libvirt" {
ERROR
ERROR
ERROR Failed to read tfstate: open /tmp/openshift-install-718900460/terraform.tfstate: no such file or directory
FATAL failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change
```

In order to have the default.xml network available in /usr/share/libvirt/networks/ an additional package is required for power hardware:

`yum install libvirt-daemon-config-network`

For SNC, I think we could just include a simple check in  `run_preflight_checks()`  for default network:

```
if ! virsh -c qemu:///system net-list --all --name | grep "default" &> /dev/null; then
echo "default network not available"
fi
```

gbraad:
Please provide a PR as we do not maintain the PPC generation code

Issue #227, opened 2020-08-28
Installer exits with error message: ERROR Error: virError(Code=38, Domain=7, Message='unable to connect to server at '192.168.122.1:16509': Connection timed out')


anderotxoa:
Hi

Im trying to install snc in a POWER8 box with 8 cores and 128GB RAM. I have managed to install it with no issues in an x86 box but in the P8 box te installer stops with the following message:

DEBUG Initializing the backend...                  
DEBUG                                              
DEBUG Initializing provider plugins...             
DEBUG                                              
DEBUG Terraform has been successfully initialized! 
DEBUG                                              
DEBUG You may now begin working with Terraform. Try running "terraform plan" to see 
DEBUG any changes that are required for your infrastructure. All Terraform commands 
DEBUG should now work.                             
DEBUG                                              
DEBUG If you ever set or change modules or backend configuration for Terraform, 
DEBUG rerun this command to reinitialize your working directory. If you forget, other 
DEBUG commands will detect it and remind you to do so if necessary. 
+ ./oc get etcds cluster
+ sleep 3
+ ./oc get etcds cluster
+ sleep 3
+ ./oc get etcds cluster
ERROR                                              
ERROR Error: virError(Code=38, Domain=7, Message='unable to connect to server at '192.168.122.1:16509': Connection timed out') 
ERROR                                              
ERROR   on ../../tmp/openshift-install-554439829/main.tf line 1, in provider "libvirt": 
ERROR    1: provider "libvirt" {                   
ERROR                                              
ERROR                                              
ERROR Failed to read tfstate: open /tmp/openshift-install-554439829/terraform.tfstate: no such file or directory 
FATAL failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change 
+ echo 'failed to create the cluster, but that is expected.  We will block on a successful cluster via a future wait-for.'
failed to create the cluster, but that is expected.  We will block on a successful cluster via a future wait-for.
+ renew_certificates
++ ./oc adm release -a /root/snc/secret.txt info quay.io/openshift-release-dev/ocp-release@sha256:2ad9a328235f9c116e1c4319df2f0610253b40646b45afe846798ed4ffe6bb7d --image-for=cli
+ cli_image=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:cb95ba34205fd6d3bd84b5fb6d3c8fb44333781fabc519838094bc30ee252679
[...]

PR #251, opened 2020-10-13
Check if libvirt default network is present for ppc64le

mtarsel:
For ppc64le arch, the default network must be present in order to proceed with the snc install.

If default network is not present, the script will exit. The user should install libvirt-daemon-config-network package.

Closes issue #250

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -73,6 +73,18 @@ function run_preflight_checks() {
         if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
                 preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
         fi
+
+	if ! virsh -c ${LIBVIRT_URI} net-info default &> /dev/null; then
+		echo "Default libvirt network is not available. Exiting now!"
+		exit 1
+	fi
+	echo "default network is available"
+
+	#Check if default libvirt network is Active
+	if [[ $(virsh -c ${LIBVIRT_URI}  net-info default | awk '{print $2}' | sed '3q;d') == "no" ]]; then
+		echo "Default network is not active. Exiting now!"
+		exit 1
+	fi
 	
 	#Just warn if architecture is not supported
 	case $ARCH in


cfergeau:
What is different on ppc64le compared to all others architecture? Why only this arch require the `default` libvirt network?

cfergeau:
I believe all arches will try to connect to `192.168.122.1` as this is why we need to enable TCP support for libvirt. So rather than a ppc64le specific test, I'd go with what you initially suggested in #250. This can use `virsh -c qemu:///system net-info default` rather than `virsh net-list`

mtarsel:
> Why only this arch require the default libvirt network?

I figured it was only on power since it wasn't a check already in the script. It was related to #227 and I confirmed this on my ppc64le box as well.

I'll redo this commit with a better message. I'll also add a check to make sure the default network is not only present via `net-info` but also Active.

mtarsel:
I deleted the previous commit which looks like it closed this PR automatically. I cleaned up the commit and added more comments in this same branch so it's re-opened now. thanks

snc.sh
```diff
@@ -74,6 +74,20 @@ function run_preflight_checks() {
...
+	
+		#Check if default libvirt network is Active
+		if [[ $(virsh -c qemu:///system net-info default | awk '{print $2}' | sed '3q;d') == "no" ]]; then
+			echo "Default network is not active. Exiting now!"
+			exit 1
+		fi
+	else
+		echo "Default libvirt network is not available. Exiting now!"
+		exit 1
+	fi
```
cfergeau, line 90: Would be simpler as 
```
if ! virsh -c ${LIBVIRT_URI} net-info default &> /dev/null; then
	echo "Default libvirt network is not available. Exiting now!"
	exit 1
fi
echo "default network is available"

#Check if default libvirt network is Active
if [[ $(virsh -c qemu:///system net-info default | awk '{print $2}' | sed '3q;d') == "no" ]]; then
	echo "Default network is not active. Exiting now!"
	exit 1
fi
```
This saves some indentation levels, and the nominal (no error) case is easier to follow.
For the "active" check, I'd just use something like `virsh -c qemu:///system net-info default |grep '^Active: *yes`, this is easier to read, and feels more robust.

mtarsel:
rebased and pushed with edits from @cfergeau

cfergeau (approved):

PR #260, opened 2020-10-15
Add the script run by the CI

guillaumerose:
Extracted from https://github.com/openshift/release/blob/431aac76304a21b1c290811441d8b12b9252450c/ci-operator/step-registry/code-ready/snc/e2e/test/code-ready-snc-e2e-test-commands.sh

This will allow us to define the CI script here instead of defining it in openshift/release repo.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+./snc.sh
+if [[ $? -ne 0 ]]; then
+  exit 1
+fi
+# wait till the cluster is stable
+sleep 5m
+export KUBECONFIG=crc-tmp-install-data/auth/kubeconfig
+# Remove all the failed Pods
+oc delete pods --field-selector=status.phase=Failed -A
+# Wait till all the pods are either running or pending or completed or in terminating state
+while oc get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed | grep -v Terminating | grep -v Pending; do
+   sleep 2
+done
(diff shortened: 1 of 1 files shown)


ci.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+./snc.sh
+if [[ $? -ne 0 ]]; then
+  exit 1
+fi
+# wait till the cluster is stable
+sleep 5m
```
cfergeau, line 7: Not the right PR for mentioning this, if a sleep 5m is needed after `snc.sh` returns, I'd argue it's a snc bug.

ci.sh
```diff
@@ -0,0 +1,21 @@
...
+export KUBECONFIG=crc-tmp-install-data/auth/kubeconfig
+# Remove all the failed Pods
+oc delete pods --field-selector=status.phase=Failed -A
```
cfergeau, line 10: and this is using the wrong `oc`, should be the one `snc` downloaded.

cfergeau (approved):
couple of comments, but I guess preexisting ones which only become visible because the script is no longer buried somewhere in the CI.

guillaumerose:
Next step is https://github.com/openshift/release/pull/13485 then use ci.sh in https://github.com/openshift/release/blob/431aac76304a21b1c290811441d8b12b9252450c/ci-operator/step-registry/openshift/e2e/gcp/crc/test/openshift-e2e-gcp-crc-test-commands.sh

PR #263, opened 2020-10-22
Do not download mac/windows binary for ppc64le

mtarsel:
These binaries will not work on ppc64le. Only create the openshift-clients/ directories which are needed by the architecture. This also remove the need for the unzip package on ppc64le.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -22,7 +22,6 @@ INSTALL_DIR=crc-tmp-install-data
 JQ=${JQ:-jq}
 XMLLINT=${XMLLINT:-xmllint}
 YQ=${YQ:-yq}
-UNZIP=${UNZIP:-unzip}
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
With just this change, `createdisk.sh` will fail when running `generate_hyperv_directory`/`generate_hyperkit_directory` (which are also not needed on ppc64)

praveenkumar:
This should be handled as part of `createdish.sh`, CI passes because of x86_64 platform.

mtarsel:
rebase was a little messy for me, sorry! deleted branch and pushed cleanly so hopefully works out

snc.sh
```diff
@@ -266,16 +265,21 @@ function delete_operator() {
...
+	unzip -o -d openshift-clients/windows/ openshift-clients/windows/oc.zip
 fi
 
-# Download the oc binary for all platforms
-mkdir -p openshift-clients/linux openshift-clients/mac openshift-clients/windows
-curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-linux.tar.gz" | tar -zx -C openshift-clients/linux oc
-curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac.tar.gz" | tar -zx -C openshift-clients/mac oc
-curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-windows.zip" > openshift-clients/windows/oc.zip
-${UNZIP} -o -d openshift-clients/windows/ openshift-clients/windows/oc.zip
 OC=./openshift-clients/linux/oc
```
cfergeau, line 283: I'd keep that `OC` line grouped with the extraction of `openshift-client-linux.tar.gz`.
However, https://github.com/code-ready/snc/pull/263#pullrequestreview-514421267 still seems to be unaddressed? These mac/windows clients exist in `${MIRROR}/${OPENSHIFT_RELEASE_VERSION}` even for ppc64le/s390, and as your change currently stands, `create_disk.sh` will fail because it's expecting these windows/macos binaries to exist.
I agree it's suboptimal to download these binaries on ppc64le, but I also don't think it's breaking anything? So I'm reluctant to take a patch removing this, but breaking create_disk.sh.

mtarsel: ok sorry I misunderstood the review comment and will address this change in my next PR. I am thinking about just adding a case statement into createdisk.sh for different archs. Similar to what was added in `run_preflight_checks()` in snc.sh so we can build the bundles for the specific archs as needed. I'm not sure yet, but something similar to...


```
ARCH=$(uname -m)
case $ARCH in
    x86_64)
        generate_hyperKit
        generate_hyperV;;
    ppc64le|s390x)
        generate_ppc64le;;
    *)	
        echo "The host arch is ${ARCH}. This is not supported by SNC!" && exit 1;;
esac
```

I'll take a closer look at createdisk.sh and submit a more appropriate PR :)

cfergeau: You'll probably have a `generate_libvirt` both on x86_64 and ppc64le, but otherwise makes sense.

mtarsel:
Closing this. Use PR #305

PR #264, opened 2020-10-26
Use a constant name for node name crc-abcde

guillaumerose:
It allows crc and consumers of snc to have a static DNS configuration.
It doesnt depend anymore of a random name picked by the installer.

---

Only few files contains the random name, it is faily easy to change.

```
crc-tmp-install-data/manifests/cluster-infrastructure-02-config.yml
15:  infrastructureName: crc-abcde

crc-tmp-install-data/openshift/99_openshift-cluster-api_master-machines-0.yaml
6:    machine.openshift.io/cluster-api-cluster: crc-abcde
9:  name: crc-abcde-master-0
27:      networkInterfaceName: crc-abcde
31:        baseVolumeID: crc-abcde-base
32:        poolName: crc-abcde

crc-tmp-install-data/openshift/99_openshift-cluster-api_worker-machineset-0.yaml
6:    machine.openshift.io/cluster-api-cluster: crc-abcde
9:  name: crc-abcde-worker-0
15:      machine.openshift.io/cluster-api-cluster: crc-abcde
16:      machine.openshift.io/cluster-api-machineset: crc-abcde-worker-0
21:        machine.openshift.io/cluster-api-cluster: crc-abcde
24:        machine.openshift.io/cluster-api-machineset: crc-abcde-worker-0
41:          networkInterfaceName: crc-abcde
45:            baseVolumeID: crc-abcde-base
46:            poolName: crc-abcde

```

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -388,6 +388,11 @@ ${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml sshKey "$(cat id_rsa_cr
 # Create the manifests using the INSTALL_DIR
 ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests || exit 1
 
+# Use a fixed node name
+OLD_INFRA_NAME=$(${YQ} read ${INSTALL_DIR}/manifests/cluster-infrastructure-02-config.yml 'status.infrastructureName')
+NEW_INFRA_NAME=$(${YQ} read install-config.yaml metadata.name)-abcde
+find ${INSTALL_DIR} -type f -exec sed -i "s/${OLD_INFRA_NAME}/${NEW_INFRA_NAME}/g"  {} \;
+
 # Add custom domain to cluster-ingress
 ${YQ} write --inplace ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml spec[domain] apps-${CRC_VM_NAME}.${BASE_DOMAIN}
 # Add master memory to 12 GB and 6 cpus 


gbraad:
doesn't fail on mirror.openshift.com now, but instead: 


```
NWRAP_ERROR(103) - nwrap_files_cache_reload: Unable to open '/tmp/packer/passwd' readonly -1:No such file or directory
NWRAP_ERROR(103) - nwrap_files_getpwuid: Error loading passwd file
NWRAP_ERROR(118) - nwrap_files_cache_reload: Unable to open '/tmp/packer/passwd' readonly -1:No such file or directory
NWRAP_ERROR(118) - nwrap_files_getpwuid: Error loading passwd file
NWRAP_ERROR(122) - nwrap_files_cache_reload: Unable to open '/tmp/packer/passwd' readonly -1:No such file or directory
NWRAP_ERROR(122) - nwrap_files_getpwuid: Error loading passwd file
unknown user 1002320000
ERROR: (gcloud.compute.scp) [/bin/scp] exited with return code [255].
error: failed to execute wrapped command: exit status 255
```

snc.sh
```diff
@@ -388,6 +388,11 @@ ${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml sshKey "$(cat id_rsa_cr
 # Create the manifests using the INSTALL_DIR
 ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests || exit 1
 
+# Use a fixed node name
+OLD_INFRA_NAME=$(${YQ} read ${INSTALL_DIR}/manifests/cluster-infrastructure-02-config.yml 'status.infrastructureName')
+NEW_INFRA_NAME=$(${YQ} read install-config.yaml metadata.name)-abcde
```
praveenkumar, line 393: for `NEW_INFRA_NAME` can be just `${CRC_VM_NAME}-abcde` instead reading the install config again because that what we do as part install-config https://github.com/code-ready/snc/blob/master/snc.sh#L382

cfergeau:
At the moment, crc is fine with this dynamic name, it picks it from the bundle metadata, and adjusts the DNS configuration as needed. This change would remove the "pick it from the bundle metadata" part, we'd still need to setup the name resolution for crc-abcde somewhere.

This change would help with https://github.com/code-ready/crc/pull/1526 as at the moment it's not possible for `crc start` to dynamically update the DNS information when it knows which bundle it's  dealing with. My understanding is that the code in this PR is already hardcoding `crc.testing` and `apps-crc.testing` rather than getting them from the bundle information. So the bigger question is how strongly we want to tie crc with the bundles snc generates. Currently we try to go through the bundle metadata file for any bundle specific information, even if  some of these values never changes.
I'd prefer that we keep some flexibility, and rather than hardcoding things `crc` has no control over, we could have a way to feed the bundle metadata to the userland DNS implementation so that it can update its behaviour (or is this a ton of work?)

guillaumerose:
I hack a bit the userland netstack and now I use a regexp for this domain. https://github.com/guillaumerose/gvisor-tap-vsock/blob/master/cmd/host/main.go#L63

praveenkumar:
@guillaumerose If we don't need it anymore, can we close this PR or you are planning to have this to be part of bundle and not use the regex as part of vsock codebase?

guillaumerose:
Yes closing it.

PR #267, opened 2020-11-15
Add bundle name to metadata json

guillaumerose:
Currently the name used by crc is the directory name contained in the
crcbundle file. crc also guesses it from the crcbundle file, it takes
the name of the bundle without the extension.
Having a clear name property and the fact that the metadata file can be
read by only extracting the first bytes of the bundle, we can remove
this ambiguity.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -135,6 +135,7 @@ function update_json_description {
     diskSha256Sum=$(sha256sum $destDir/${CRC_VM_NAME}.qcow2 | awk '{print $1}')
 
     cat $srcDir/crc-bundle-info.json \
+        | ${JQ} ".name = \"${destDir}\"" \
         | ${JQ} '.clusterInfo.sshPrivateKeyFile = "id_rsa_crc"' \
         | ${JQ} '.clusterInfo.kubeConfig = "kubeconfig"' \
         | ${JQ} '.clusterInfo.kubeadminPasswordFile = "kubeadmin-password"' \
--- a/crc-bundle-info.json.sample
+++ b/crc-bundle-info.json.sample
@@ -6,6 +6,8 @@
   # Type of this bundle content
   # Currently the only valid type is 'snc' (which stands for 'single-node-cluster')
   "type": "snc",
+  # Name of the bundle
+  "name": "crc_libvirt_4.6.1",
   "buildInfo": {
     # Time this bundle was built
     "buildTime": "2019-04-23T14:55:32+00:00",


createdisk.sh
```diff
@@ -135,6 +135,7 @@ function update_json_description {
     diskSha256Sum=$(sha256sum $destDir/${CRC_VM_NAME}.qcow2 | awk '{print $1}')
 
     cat $srcDir/crc-bundle-info.json \
+        | ${JQ} ".name = \"${destDir}\"" \
```
praveenkumar, line 138: @guillaumerose I didn't understand the improvement using this patch, here we are putting the name against the `$destDir` which in the end use to create tarball for different platform, point I didn't get about ambiguity around doing same as part of  `crc` codebase.

guillaumerose: I would like when crc uses a bundle that it unpacks only the metadata to get the name, then create the directory, then unpack everything in without looking at the structure of the initial archive.

It will avoid bug when the crcbundle filename is different of the directory inside the crcbundle.

`crc start -p ~/pull-secret.txt -b ~/Downloads/crc_libvirt_2020-11-15\(5\).crcbundle --log-level debug`

I hope also to support both a path and a bundle name. `crc start -p ~/pull-secret.txt -b crc_libvirt_2020-11-15 --log-level debug`

Also, I would like to change the bundle name to `openshift-4.6.1`. So that I can have a collection of bundle like `openshift-x` and `okd-x`.

cfergeau:
> I would like when crc uses a bundle that it unpacks only the metadata to get the name, then create the directory, then unpack everything in without looking at the structure of the initial archive.
...
> Also, I would like to change the bundle name to openshift-4.6.1. So that I can have a collection of bundle like openshift-x and okd-x.

It seems you want to put crc in control of the target directory name it uses to unpack the bundle content. This name would depend on some metadata specific to the bundle (is it openshift or okd? which version?). This seems to be an argument against adding `"name": "crc_libvirt_4.6.1"` to the metadata though, as you'd then want crc to ignore this?

guillaumerose:
> This name would depend on some metadata specific to the bundle (is it openshift or okd? which version?). 

Yes

> This seems to be an argument against adding "name": "crc_libvirt_4.6.1" to the metadata though, as you'd then want crc to ignore this?

crc will not ignore it but use it as the bundle name instead of the filename, instead of a directory name. If we want to rename the bundle for crc, we would need to rename it here, in snc.

cfergeau:
> If we want to rename the bundle for crc, we would need to rename it here, in snc.

Is it useful to let snc mandate where the bundle should be extracted to? We could change crc extraction code to always extract foo.crcbundle to `crc_$hypervisor_$openshift_version` for now rather than using the default name used when creating the tarball. This also can all be inferred from the bundle metadata, and does not require any changes snc side. Is there some additional benefit I'm missing?

guillaumerose:
If we choose this pattern this will prevent us from running variations of an openshift bundle. For instance, we could have a bundle `openshift-4.6.1.libvirt`, `okd-4.6.1.libvirt` and a bundle `openshift-4.6.1.libvirt+srikrishna`

cfergeau:
I was merely describing how we can slightly change crc to keep exactly the same behaviour as currently, while moving toward what you describe. Once crc takes control of the naming of the directory the bundle is unpacked to, it can do `${product}-${product-version}.${hypervisor}` if that's what we want.
The `+srikrishna` example is interesting as this is not something which can be currently done nicely, which means a new field would be needed in the bundle metadata anyway. Your suggestion is to have a `bundle-directory-name`, but maybe an `extra-version` field would be more useful as this is also needed if we want to differentiate multiple bundles which are made for the same openshift version (such as what we did for 4.6.1)?

guillaumerose:
My suggestion is to give a clear name to each bundle, and not let the consumer guess it like today. Consumer can do what they want but the creator of the bundle gives it a name when building it and today, the name is crc_libvirt_4.6.1.

cfergeau:
If we add that name, and this name matches the name of the directory in the tarball, and if we make it mandatory to keep the directory structure and use that name, fine with me.
If we add that name, but also change crc to ignore this name, and make it use its own naming instead, I would say it's pointless to add a name field to snc.

guillaumerose:
> If we add that name, and this name matches the name of the directory in the tarball, and if we make it mandatory to keep the directory structure and use that name, fine with me.

This is exactly the first case!

cfergeau:
Thinking some more about this, I wonder if this really is required, or if things could be improved solely on crc side?
1. When crc gets a `.crcbundle` file and uncompresses it, it knows which dir the bundle metadata was uncompressed to, so it does not have to guess it from the `.crcbundle` filename (even if this is what it's doing at the moment).
2. When we have an already existing VM, and crc needs to get to the bundle metadata, it guesses the path from the bundle name recorded in ~/.crc/machines/crc/config.json. If we don't want to rely on guessing the path, we can fallback on 1., or we could record the path to the bundle metadata file in that config.json file. I'd favour the latter
3. I think the only remaining case is when the bundle is not embedded, and bundle name is different from the tarball directory structure. We don't support this very well at the moment, if at all, but I'm not sure adding a name to the bundle would make a big difference?

guillaumerose:
> I think the only remaining case is when the bundle is not embedded, and bundle name is different from the tarball directory structure. We don't support this very well at the moment, if at all, but I'm not sure adding a name to the bundle would make a big difference?

Yes you're right, it is only for us. It is just to ease our life and it removes one more ambiguity/bug.

praveenkumar:
After reading https://github.com/code-ready/snc/pull/267#issuecomment-730221705 (use case - 2) to get the metadata info from the bundle you have it extracted and right now in second run we don't run extraction because the cache exist and to get the info about that cache we make use of the `BundleName` which is part of ` ~/.crc/machines/crc/config.json` so are you suggesting that after this change we always run extraction code (even just extract the metadata bit) and then get the cache info?

guillaumerose:
> are you suggesting that after this change we always run extraction code (even just extract the metadata bit) and then get the cache info?

No. I would like to refactor completely the bundle package. Have a proper repository struct like:

```
type BundleRepository interface {
 Install(path string) *Bundle, error
 Get(name string) *Bundle, error
 Delete(name string) error
}
``` 

BundlePath should probably not be used outside of the cmd/crc package and all the pkg code should rely on the bunde name and use the repository if needed.

createdisk.sh
gbraad: > Also, I would like to change the bundle name to openshift-4.6.1. So that I can have a collection of bundle like openshift-x and okd-x.

This would make sense, but this would be `ocp` and `okd`, right? OpenShift is a very generic term in that case.

gbraad:
> Is it useful to let snc mandate where the bundle should be extracted to? 

it is however the responsibility of the snc bundle and it's metadata to identify and name. The use of a name seems a very valid approach.

However, are you suggestion a more complex:

```json
  Version: '4.6.2',
  Product: "ocp"
```

and then rely on `crc` to combine these elements? I do not see the need for CRC to maintai such logic. Also consider that the bundle can be consumed outside of crc, and in that case a 'name' seems a very valid identifier to extract to mean one thing. Details can be added where you see fit and even duplicated, as nothing prevents:

```json
  Name: "openshift-4.6.2",
  Version: "4.6.2',
  Product: 'ocp'
```

The idea of a BundleRepository is in line how we thought of the cache being used. a simple iteration over the directories (or getting the name from the metadata) in that case is sufficient allows flexibility outside our usecase, like a "custom-websphere-4.6.2" to identify an image maintained outside our team, and still being listed without too much logic. details can be added where people see fit.

gbraad (approved):

cfergeau:
> > Is it useful to let snc mandate where the bundle should be extracted to?
> 
> it is however the responsibility of the snc bundle and it's metadata to identify and name. The use of a name seems a very valid approach.
> 
> However, are you suggestion a more complex:
> 
> ```json
>   Version: '4.6.2',
>   Product: "ocp"
> ```
> 
> and then rely on `crc` to combine these elements?

Nope, this was asked before, the plan is to use the name as is in crc
https://github.com/code-ready/snc/pull/267#issuecomment-729050906

> The idea of a BundleRepository is in line how we thought of the cache being used. a simple iteration over the directories (or getting the name from the metadata) in that case is sufficient allows flexibility outside our usecase, like a "custom-websphere-4.6.2" to identify an image maintained outside our team, and still being listed without too much logic. details can be added where people see fit.

Just one thing to keep in mind, 'simple iteration over directories' is not free (though it's a lot faster with SSDs). This is not going to be a bottleneck with crc, so not worth a lengthy discussion :)

guillaumerose:
It doesn't engage us too much. We can still remove it easily.

PR #277, opened 2020-11-20
BUG 1899459: Delete prometheus validation admission webhook

praveenkumar:
Unsupported config is used to make etcd work during bootstrap but quorum is
still set to 3 nodes (replica) in the etcd-quorum-guard deployment. In case,
we want to make modification in the CVO override list after crc start then
reconciliation doesn't happen and it throw following error.

```
Error while reconciling 4.6.3: the workload openshift-etcd/etcd-quorum-guard has not yet successfully rolled out
```

With this patch etcd deployment is override to CVO and scale is set to 1 so that
there is no pending pod for quorum.

### DIFF
--- a/cvo_override.yaml
+++ b/cvo_override.yaml
@@ -56,3 +56,8 @@
     name: kube-storage-version-migrator-operator
     namespace: openshift-kube-storage-version-migrator-operator
     unmanaged: true
+  - kind: Deployment
+    group: apps/v1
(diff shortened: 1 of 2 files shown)


praveenkumar:
/retitle BUG 1899459: Delete prometheus validation admission webhook

cfergeau:
Successfully tested on a live cluster in https://github.com/code-ready/crc/issues/1665

PR #278, opened 2020-11-23
Remove system reserved config block from kubelet

praveenkumar:
Following is current setting for system reserved resource for kubelet
and since we are aiming to developers so removing this can free up some resource.

```
systemReserved:
  cpu: 500m
  memory: 1Gi
  ephemeral-storage: 1Gi
```

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -61,6 +61,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
 # Stop the kubelet service so it will not reprovision the pods
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
+# Remove system reserved block from kubelet config
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} cat /etc/kubernetes/kubelet.conf \
+  | ${YQ} delete - systemReserved \
+  | ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} sudo tee /etc/kubernetes/kubelet.conf
+
 # Unmask the chronyd service
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl unmask chronyd
 # Disable the chronyd service


praveenkumar:
We will first try to create the bundle using this PR and then run our crc integration test before merging it.

createdisk.sh
```diff
@@ -326,6 +326,10 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
 # Stop the kubelet service so it will not reprovision the pods
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
+# Remove system reserved block from kubelet config
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo curl -L https://github.com/mikefarah/yq/releases/download/3.3.0/yq_linux_${yq_ARCH} -o /usr/local/bin/yq && chmod +x /usr/local/bin/yq
```
guillaumerose, line 330: Delete yq after that? or perhaps do it on the host?

guillaumerose (approved):

cfergeau:
It would be nice to document a bit more the potential side effects in the commit log. What happens when the node is under pressure before/after that change? Can the kubelet start to misbehave, or be killed because it uses more memory than it requested?

praveenkumar:
> It would be nice to document a bit more the potential side effects in the commit log. What happens when the node is under pressure before/after that change? Can the kubelet start to misbehave, or be killed because it uses more memory than it requested?

@cfergeau yes, I first need to test this generated bundle and will update the commit log with more details.

/wip

guillaumerose:
Perhaps we can use this manifest to do it ? https://github.com/openshift/machine-config-operator/blob/master/examples/kubeletconfig.crd.yaml#L9

guillaumerose:
I tried to create a KubeletConfig object and it doesn't work. MCO can't restart the node and complains.

I also tried by putting the config directly in the release image.

guillaumerose:
^ test reports a failure but it's a success. Bundles are correctly generated.

guillaumerose:
Before:
```
Allocated resources:
  cpu                2275m (65%)   0 (0%)
  memory             6941Mi (93%)  512Mi (6%)
  ephemeral-storage  0 (0%)        0 (0%)
  hugepages-1Gi      0 (0%)        0 (0%)
  hugepages-2Mi      0 (0%)        0 (0%)

```

After:
```
Allocated resources:
  cpu                2275m (56%)   0 (0%)
  memory             6941Mi (82%)  512Mi (6%)
  ephemeral-storage  0 (0%)        0 (0%)
  hugepages-1Gi      0 (0%)        0 (0%)
  hugepages-2Mi      0 (0%)        0 (0%)
```

/lgtm

createdisk.sh
cfergeau: This can be done on the host with:
```
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} cat /etc/kubernetes/kubelet.conf | ${YQ} delete - systemReserved |  ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} sudo tee /etc/kubernetes/kubelet.conf
```
which imo is nicer than downloading the binary on the VM.

guillaumerose: Nice! I updated the PR.

PR #279, opened 2020-11-23
Add common bash script for tools

praveenkumar:
Remove code duplication from snc and createdisk script around different
tools installation.


tools.sh
```diff
@@ -66,3 +66,6 @@ fi
 if ! rpm -q libguestfs-xfs; then
     sudo yum install libguestfs-xfs
 fi
+
+# Restart the libvirt service after update
+sudo systemctl restart libvirtd
```
cfergeau, line 71: Could you run a `yum update`+`systemctl restart libvirtd` in the GCP job before invoking `snc.sh`? Or even update the CI image used? I don't think it's snc.sh job to fix outdated CI images.

cfergeau (commented):
Some minor comments on the first commit, and I'd drop the second one. Apart from this, looks good to me.

snc.sh
```diff
@@ -262,9 +262,10 @@ function delete_operator() {
         local namespace=$2
         local pod_selector=$3
 
+	retry ${OC} get pods
         pod=$(${OC} get pod -l ${pod_selector} -o jsonpath="{.items[0].metadata.name}" -n ${namespace})
```
cfergeau, line 266: indentation looks odd in github UI?

cfergeau (approved):
If it helps, I don't mind approving this, but `Restart the libvirt service after update` should be reverted as soon as the GCP images are fixed.

PR #281, opened 2020-11-25
Replace unconditionnal module loading with udev rule

cfergeau:
hv_sock is only loadable when ran under hyper-v. Unconditionnally
loading it causes a warning to be shown when ssh'ing into the VM.
This udev rule will only trigger on hyper-v guests and load hv_sock.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -346,8 +346,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install /home/
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- rm -fr /home/core/packages
 
 # Adding Hyper-V vsock support
+
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
-    echo hv_sock > /etc/modules-load.d/vsock.conf
+    echo 'CONST{virt}=="microsoft", RUN{builtin}+="kmod load hv_sock"' > /etc/udev/rules.d/90-crc-vsock.rules
 EOF
 
 # Add gvisor-tap-vsock service


guillaumerose:
Nice! Could this be added upstream with the condition ?

cfergeau:
> Nice! Could this be added upstream with the condition ?

Not quite clear where upstream would be ^^ systemd/udev does not have any hv_* related rules. I think the other `hv_*` modules are loaded automatically because they specify an alias which then triggers this rule: `ENV{MODALIAS}=="?*", RUN{builtin}+="kmod load $env{MODALIAS}"`
For `hv_sock`, there is a modalias  for the PF_VSOCK protocol. Other hv_* modules have a modalias based on the GUID they use, which is harder for hv_vsock since the vsock port used is part of the GUID. No idea if there can be wildcard matches.
In short, I would not know where to put this rule, nor if something more subtle than hardcoding could be done.

cfergeau:
it seems `*` in module aliases are possible, but not immediatly obvious how to add this in `drivers/hv/vmbus_drv.c` or `net/vmw_vsock/hyperv_transport.c` (if at all possible)

guillaumerose:
I was more thinking to put it in Red Hat CoreOs.

cfergeau:
> I was more thinking to put it in Red Hat CoreOs.

The path for getting something in Red Hat CoreOs is to fix it upstream (kernel, systemd, ...), get these changes in Fedora/RHEL packages, and finally it lands in RHCOS. In that case, systemd/udev does not seem to be the right place as they don't have any hyperv-specific rules, hyperv-daemons has a few udev rules which are fedora additions, but they seem mostly related to starting the right daemon when the right hv_xxx module is loaded. Which is why I ended up looking at the kernel for lack of a better place :)

cfergeau:
I've tested the windows bundle from CI, and `hv_sock` is loaded in a Windows VM.

PR #282, opened 2020-11-26
fix olm crash

praveenkumar:


guillaumerose:
Is it the same error as https://github.com/code-ready/crc/issues/1633 ?

praveenkumar:
> Is it the same error as code-ready/crc#1633 ?

@guillaumerose No, I didn't see this error yet because even with cert rotation the `openshift-install wait-for cluster` works as expected, CI is failing because of some apiserver reconciliation part, which will need to again debug and see if we need to update something more in this PR.

snc.sh
```diff
@@ -238,47 +238,43 @@ function create_pvs() {
...
+    # This random_string is created by installer and added to each resource type,
+    # in installer side also variable name is kept as `random_string`
+    # so to maintain consistancy, we are also using random_string here.
+    random_string=$(sudo virsh list --all | grep -oP "(?<=${CRC_VM_NAME}-).*(?=-master-0)")
+    if [ -z $random_string ]; then
+        echo "Could not find virtual machine created by snc.sh"
+        exit 1;
+    fi
+    VM_PREFIX=${CRC_VM_NAME}-${random_string}
+
```
cfergeau, line 268: Can this be extracted into a helper, especially since `createdisk.sh` has the same code?

snc.sh
```diff
@@ -238,47 +238,43 @@ function create_pvs() {
...
+    sudo virsh shutdown ${VM_PREFIX}-master-0
+    # Wait till instance started successfully
+    until sudo virsh domstate ${VM_PREFIX}-master-0 | grep shut; do
+        echo " ${VM_PREFIX}-master-0 still running"
+        sleep 3
+    done
+
```
cfergeau, line 275: Also duplicated with `create_disk.sh`

snc.sh
```diff
@@ -238,47 +238,43 @@ function create_pvs() {
...
-
-    # Force the rotation for secrets
-    ${OC} get secret -A -o json | ${JQ} -r '.items[] | select(.metadata.annotations."auth.openshift.io/certificate-not-after" | .!=null and fromdateiso8601<='$( date --date='+1year' +%s )') | "-n \(.metadata.namespace) \(.metadata.name)"' | \
-        xargs -n3 ${OC} patch secret -p='{"metadata": {"annotations": {"auth.openshift.io/certificate-not-after": null}}}'
+    sudo virsh start ${VM_PREFIX}-master-0
+    # Wait till ssh connection available
+    until ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "exit 0" >/dev/null 2>&1; do
+        echo " ${VM_PREFIX}-master-0 still booting"
+        sleep 2
+    done
```
cfergeau, line 285: No timeout?

snc.sh
```diff
@@ -238,47 +238,43 @@ function create_pvs() {
...
-    sleep 300
+    # After cluster starts kube-apiserver-client-kubelet signer need to be approved
+    timeout 300 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
+    ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
 
-    # Retry 4 times to make sure kubelet certs are rotated correctly.
+    # Retry 5 times to make sure kubelet certs are rotated correctly.
     i=0
-    while [ $i -lt 4 ]; do
+    while [ $i -lt 5 ]; do
```
cfergeau, line 293: Seems fairly arbitrary? Wondering if we could do better? Maybe loop a lot more, but don't sleep 60s each time?

snc.sh
```diff
@@ -56,6 +56,24 @@ else
     fi
 fi
 
+function retry {
+    local retries=5
```
cfergeau, line 60: Wondering if there should be a 'max wait time in seconds' parameter rather than hardcoding `5`?

cfergeau (commented):
Hopefully this new way of renewing certs will be less problematic than the previous ones, and allow us to stop waiting for 24h :)

snc.sh
praveenkumar: @cfergeau yes, I am planing to do that in tools PR.

snc.sh
praveenkumar: we don't have it in create disk also, will try to add it once extract it as part of common functions.

snc.sh
praveenkumar: I know, atm I don't know if we can have something better, I didn't change the 60s time but yes we can do it 30s and add more retry.

snc.sh
praveenkumar: @cfergeau yes we can change it when we really required that value, right now with this it is mostly working.

cfergeau (approved):

snc.sh
```diff
@@ -233,6 +233,10 @@ function renew_certificates() {
     cli_image=$(${OC} adm release -a ${OPENSHIFT_PULL_SECRET_PATH} info ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=cli)
     ${YQ} write kubelet-bootstrap-cred-manager-ds.yaml.in spec.template.spec.containers[0].image ${cli_image} >kubelet-bootstrap-cred-manager-ds.yaml
 
+    # Enable the network time sync and set the clock back to present on host
+    sudo date -s '1 day'
+    sudo timedatectl set-ntp on
+
```
cfergeau, line 238: Guillaume was not keen on changing the time on some of his test machines, I thought the agreement was to make cert renewal optional (for test purposes)?

snc.sh
```diff
@@ -229,35 +229,35 @@ function create_pvs() {
...
-
-    # Wait until bootstrap csr request is generated with 5 min timeout
+    # After cluster starts kube-apiserver-client-kubelet signer need to be approved
     timeout 300 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
     ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
 
-    delete_operator "daemonset/kubelet-bootstrap-cred-manager" "openshift-machine-config-operator" "k8s-app=kubelet-bootstrap-cred-manager"
+    # Retry 5 times to make sure kubelet certs are rotated correctly.
+    i=0
+    while [ $i -lt 5 ]; do
```
cfergeau, line 247: Maybe `for i in $(seq 1 5); do` instead of manual comparison/increment?

snc.sh
```diff
@@ -229,35 +229,35 @@ function create_pvs() {
...
+	    echo "Retry loop $i, wait for 60sec before starting next loop"
+            sleep 60
+	else
+            break
+        fi
+	i=$[$i+1]
+    done
+
+    if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem; then
+        preflight_failure "Certs are not yet rotated to have 30 days validity"
```
cfergeau, line 259: unrelated to any preflight check? `preflight_failure` does not seem very appropriate to call here.

cfergeau (commented):
Main remaining question is whether to do the host time change unconditionally or not.

snc.sh
praveenkumar: I just used `preflight_failure` just to avoid an echo and exit status but yes it doesn't make sense.

snc.sh
praveenkumar: Sure, I think ti was there before but will use for loop

snc.sh
praveenkumar: @cfergeau should I introduce a env variable which is set to true by default and if user don't want the cert rotation then set it to false?

tools.sh
```diff
@@ -66,7 +66,7 @@ if ! which ${UNZIP}; then
 fi
 
 function retry {
-    local retries=5
+    local retries=10
```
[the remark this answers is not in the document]
praveenkumar: yes it was recently added and needed to add more retry to cover the case where api server doesn't reconcile after 5 retry, The one way to make it bit more robust to add another function to check if required resource is available and then wait till it not available by a timeout but then it is more like same thing so I prefer to have it this way only.

snc.sh
cfergeau: yes, sounds good to me, @guillaumerose, would that fit your usecase?

snc.sh
```diff
@@ -26,6 +26,7 @@ BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_rsa_crc"
 MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+CERT_ROTATION=${CERT_ROTATION:-true}
```
cfergeau, line 29: for env vars we are expected to set externally, I prefer something namespaced, "SNC_DISABLE_CERT_ROTATION" for example. `CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}`

snc.sh
```diff
@@ -290,7 +291,10 @@ fi
 ${UNZIP} -o -d openshift-clients/windows/ openshift-clients/windows/oc.zip
 OC=./openshift-clients/linux/oc
 
-run_preflight_checks
+if [[ ${CERT_ROTATION} == "true" ]]
+then
+    run_preflight_checks
+fi
```
cfergeau, line 297: Why disable preflight checks if cert rotation is disabled? I'd expect `renew_certificates` to be disabled?

praveenkumar: Good catch, it should be `renew_certificates` :(

snc.sh
guillaumerose: yes, it would be awesome! I use my SNC build machine as my regular machine :) It will save my https connections!

PR #286, opened 2020-12-01
Bump version of crc-bundle-info.json file

guillaumerose:
The addition of the name property is a new API. Consumers can use the
version field to determine if a name is available or not in the
metadata.

---

Each time we break the compatibility with the previous bundle, we should bump this number.

### DIFF
--- a/crc-bundle-info.json.sample
+++ b/crc-bundle-info.json.sample
@@ -2,7 +2,7 @@
   # Version of the bundle, used to denote format changes
   # Major is only increased changes incompatible with previous versions
   # Minor is increased for backwards-compatible changes
-  "version": "1.0",
+  "version": "1.1",
   # Type of this bundle content
   # Currently the only valid type is 'snc' (which stands for 'single-node-cluster')
   "type": "snc",
--- a/snc.sh
+++ b/snc.sh
@@ -154,7 +154,7 @@ function apply_auth_hack() {
 function create_json_description {
     openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
     sncGitHash=$(git describe --abbrev=4 HEAD 2>/dev/null || git rev-parse --short=4 HEAD)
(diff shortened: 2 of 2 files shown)


cfergeau:
When the only change is new field additions, after parsing we can also just check whether the new field is set or not? We'd want to keep supporting bundle metadata version N-1 and N for a while anyway.

guillaumerose:
As our friends in OpenShift would say, I prefer explicit instead of implicit :)
It seems that this number never changed. I would like to start to change it more often. We forgot to change it when we added the oc client and moved the kubeconfig file inside the VM in /opt/kubeconfig.
Next bump will be when we will add podman-remote binary inside the bundle probably.

If we don't do it, we will keep longer code dealing with unnamed bundle.

cfergeau:
Ok, this goes with https://github.com/code-ready/crc/pull/1734/commits/844e48dd91eae1db105cec57f3ca5072c05fec0e
And the plan seems to be that we track more closely bundle version changes by changing the minor version number (which indicates that it's backwards compatible). With minor version changes, crc code will cope both with older versions and newer versions. But thanks to this fine-grained version tracking, we know better when a specific change was introduced (in particular the date), and when we decide to drop the code paths for the older versions, we will change the `^1.0` to `^1.2` so that users get an error message when using the older version.

This all makes sense to me, and it would have been useful to have the full explicit explanation from the start ;)

cfergeau:
And I hate to start a conversation on a minor detail which could go for a while, but should we use a 3rd field for minor additions (new fields)? `1.0.1` rather than `1.1`? With just 2 numbers, we could end up with version `1.74` soon, which may not be desirable?

guillaumerose:
> 1.0.1 rather than 1.1? With just 2 numbers, we could end up with version 1.74 soon, which may not be desirable?

As you wish, I don't have a strong opinion on this.

gbraad (approved):

PR #288, opened 2020-12-02
Customize the release image to fit our needs until it reaches OpenShift upstream

guillaumerose:
Before cluster profile is available, we can customize manifests in the release image. Once we validate a change, we can push it upstream. It allows us to validate quickly an idea.

Usage:

Image preparation:
``` 
$ ./prepare-image.sh 
STEP 8: RUN touch /release-manifests/image-references /release-manifests/release-metadata
STEP 9: COMMIT quay.io/username/ocp-release:4.6.8-4.2.8-154-gb998c3b
--> c7f24f0f582
c7f24f0f5824eaebd5cbbb9fca44dc8629ddddab2033b2c4a92c63def43fbbe4
$ podman push quay.io/username/ocp-release:4.6.8-4.2.8-154-gb998c3b
$ edit snc.sh to use this image
$ ./snc.sh
```

When making a PR:
* The contributor create the PR
* It is approved and lgtm'ed
* The maintainer checkout the PR, run ./prepare-image.sh and pushes the new image
* The maintainer asks the contributor to add a commit with the new image in snc.sh
* The PR is updated and merged when CI is green.

This process is a real burden but with only have a few PR in this repo. Hopefully it will be ok.

With this PR, the current release image is set in the stone and doesn't move automatically to the last OpenShift version. It is maybe a good thing and avoid an unwanted version bump.


gbraad:
this could also be done when the cluster is started from crc, right?

guillaumerose:
Updated and I think it can be merged like this.

/unhold

> this could also be done when the cluster is started from crc, right?

No. The point of this PR is to remove transient states when bootstrapping the cluster. 
Exemple: start with HA oauth and switch to non-HA.
It also removes the need of some cvo-overrides.
Ideally, we will remove all changes that are performed after the cluster is started.

praveenkumar:
After looking at the changes looks like it not once in while we need to create that custom image but also create this custom image for each minor/major release of openshift before even generating the bundles and then the real issue will come when we try to generate the bundles for nightly. 

Every single time this is pushed to `quay.io/crcont/ocp-release:<release-hash>` on before consuming it. Another issue is our CI on this repo will also going to use whatever one pushed to `quay.io/crcont` side.

guillaumerose:
Yes it won't help for nightly. We can think about an org to store these temporary release images but it's more work for sure.
For the crcont usage, yes, we will need to be careful with this organisation. 

Profile won't be available for 4.7. This is the (only?) way to get the same benefits !

guillaumerose:
Profile will be used to decide what is included in a release. How it runs is controlled by the cluster capabilities API. 
With 4.7, we can replicate cluster profile entirely with cvo overrides. 

Let's wait for 4.8 for all the single node improvements.

PR #289, opened 2020-12-05
Changes to support OKD 4.6 and change SSH key from RSA to ECDSA

cgruver:
io.podman.socket changed to podman.socket for FCOS 33
enable fedora and fedora-updates repos for hyperV driver in FCOS 33
change rsa key to ecdsa for SSH for both FCOS 33 and RHCOS

revolves https://github.com/code-ready/crc/issues/1747 for snc bundle
**Note:** this will break crc builds without: https://github.com/code-ready/crc/pull/1757

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -300,7 +300,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/co
 if [[ ${OKD_VERSION} != "none" ]]
 then
     # Install the hyperV and libvarlink-util rpms to VM
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
     ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install --allow-inactive hyperv-daemons libvarlink-util'
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora.repo'
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora-updates.repo'
 else
     # Download the hyperV daemons and libvarlink-util dependency on host
     mkdir $1/packages
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -282,8 +282,8 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl unmask chronyd
 # Disable the chronyd service
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable chronyd
 
-# Enable the io.podman.socket service
-    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable io.podman.socket
+# Enable the podman.socket service for API V2
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable podman.socket
 
 # Remove all the pods except openshift-sdn from the VM
 pods=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo crictl pods -o json | jq '.items[] | select(.metadata.namespace != "openshift-sdn")' | jq -r .id)
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -7,8 +7,8 @@ export LANG=C
 
 source tools.sh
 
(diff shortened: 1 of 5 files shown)


guillaumerose:
This breaks the format of the bundle. We shoud consider bumping it to a new major version.

createdisk.sh
```diff
@@ -283,8 +283,13 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl unmask chronyd
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable chronyd
 
 # Enable the io.podman.socket service
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable io.podman.socket
-
+if [[ ${OKD_VERSION} != "none" ]]
+then
+    # FCOS 33 drops the io. from podman.socket
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable podman.socket
+else
```
gbraad, line 290: how about coreos releases? @anjannath need to keep an eye on this.

gbraad (approved):

cfergeau (changes requested):
Can this be split in 3 separate commits?

createdisk.sh
```diff
@@ -283,8 +283,13 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl unmask chronyd
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable chronyd
 
 # Enable the io.podman.socket service
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable io.podman.socket
-
+if [[ ${OKD_VERSION} != "none" ]]
+then
+    # FCOS 33 drops the io. from podman.socket
```
cfergeau, line 288: this is misleading, `io.podman.socket` is for varlink support, `podman.socket` is  for apiv2 support. `podman-2.0.5-5.module+el8.3.0+8221+97165c3f.x86_64` on my RHEL machine has both for example:
```
$ rpm -ql podman |grep socket
/usr/lib/systemd/system/io.podman.socket
/usr/lib/systemd/system/podman.socket
/usr/lib/systemd/user/io.podman.socket
/usr/lib/systemd/user/podman.socket
```

varlink support was dropped from recent podman versions, so `io.podman.socket` is gone. As I understand it, the plan was to make use of the newer API, so RHCOS should be fine too.
crc 1.19 (OpenShift 4.6.3) has `podman-1.9.3-3.rhaos4.6.el8.x86_64` which also has both socket files. I don't think this needs to be FCOS specific, the commit can just be "stop using varlink, it's gone from recent podman versions, use apiv2 instead"

createdisk.sh
```diff
@@ -300,6 +305,8 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/co
 if [[ ${OKD_VERSION} != "none" ]]
 then
     # Install the hyperV and libvarlink-util rpms to VM
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
+    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
```
cfergeau, line 309: Should this be disabled again once the packages are installed?

snc.sh
```diff
@@ -332,8 +332,8 @@ fi
 ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} destroy cluster ${OPENSHIFT_INSTALL_EXTRA_ARGS} || echo "failed to destroy previous cluster.  Continuing anyway"
 # Generate a new ssh keypair for this cluster
 
-rm id_rsa_crc* || true
-ssh-keygen -N "" -f id_rsa_crc -C "core"
+rm id_ecdsa_crc* || true
+ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
```
cfergeau, line 336: out of curiosity, why -b 521? why ecdsa rather than ed25519? (could be in the commit log)

createdisk.sh
cgruver: Agreed.  That is clearer.  I'll remove the OKD switch on that.

createdisk.sh
cgruver: It probably should be.  It doesn't hurt anything leaving it there, except that folks can install packages on the VM that they shouldn't install that way on a real cluster.

I'll add that logic.

cgruver:
> Can this be split in 3 separate commits?

@cfergeau Do you mean?:

1. Drop varlink in favor of API v.2 for Podman.
1. Install HyperV driver in FCOS 33
1. Change from RSA to ECDSA for SSH

snc.sh
cgruver: No real reason for ecdsa over ed25519...  ECDSA has been around a little longer, but still considered solid.

Since 521 bits is the largest key size, but still not big enough to rally impact anything, I used 521 over 256 or 384.  Again, no real reason.  I'll add a comment in the script.

createdisk.sh
cgruver: Done.

createdisk.sh
cgruver: Done.

cgruver:
Split into 3 commits as requested.

cfergeau (approved):

guillaumerose:
Generated bundle works fine with current crc.
/lgtm

PR #291, opened 2020-12-07
Fix cURL argument

guillaumerose:
Before:
```
+ echo 'Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to quay.io/openshift/okd:4.6.0-0.okd-2020-11-27-200126'
```

After:
```
+ echo 'Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to quay.io/openshift/okd@sha256:a5392db432c990ee012ab189e9fa0d7b6a168f6e9b9e9da7fb10768b48752ca0'
```

This fixes the cURL for OKD.


snc.sh
```diff
@@ -296,7 +296,7 @@ elif [ ! -f ${OPENSHIFT_PULL_SECRET_PATH} ]; then
 fi
 
 if test -z "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE-}"; then
-    OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -l "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/release.txt" | sed -n 's/^Pull From: //p')"
+    OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/release.txt" | sed -n 's/^Pull From: //p')"
```
cfergeau, line 298: use explicitly `--location`, as `-l` seems to have been a typo for `-L`?

cfergeau (commented):
Not clear from the log why ' Use the same mechanism for both okd and ocp to get the release image ' is an option now? Is this a consequence of curl now following redirections?

snc.sh
```diff
@@ -17,7 +17,6 @@ if [[ ${OKD_VERSION} != "none" ]]
 then
     OPENSHIFT_VERSION=${OKD_VERSION}
     MIRROR=${MIRROR:-https://github.com/openshift/okd/releases/download}
-    OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE:-quay.io/openshift/okd:${OPENSHIFT_VERSION}}
```
praveenkumar, line 20: If you remove this one then https://github.com/code-ready/snc/blob/master/snc.sh#L305 will be set as the image override, not something we want in case of OKD.

praveenkumar (changes requested):

snc.sh
praveenkumar: Okay so looks like now https://github.com/openshift/okd/releases/download/4.6.0-0.okd-2020-11-27-200126/release.txt it have release txt which we can extract release info also

snc.sh
praveenkumar: Looks like the typo.

guillaumerose: `curl -L` is a lot of place. Should I use `${CURL}` instead as an alias ?

cfergeau: Ah could be useful, or I'm fine with the change as it is now after your comment.

guillaumerose: let's start with this then 👍

PR #292, opened 2020-12-09
Don't install libvarlink-util on the VM

cfergeau:
This is a follow-up to 634b37a and to the removal of varlink support
in upstream podman.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -83,21 +83,21 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/co
 
 if [[ ${OKD_VERSION} != "none" ]]
 then
-    # Install the hyperV and libvarlink-util rpms to VM
+    # Install the hyperV rpms to VM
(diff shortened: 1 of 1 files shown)


praveenkumar:
We should merge this as soon as RHCOS have latest version of podman for which it is not required.

/hold
/lgtm

gbraad (approved):

gbraad:
I agree with Praveen as this is better to coincide with the Podman v2 inclusion.

/hold
/lgtm

cfergeau:
> We should merge this as soon as RHCOS have latest version of podman for which it is not required.

This is already the case, see https://github.com/code-ready/snc/pull/289#discussion_r537401536

> io.podman.socket is for varlink support, podman.socket is for apiv2 support. podman-2.0.5-5.module+el8.3.0+8221+97165c3f.x86_64 on my RHEL machine has both for example:
> [...]
> varlink support was dropped from recent podman versions, so io.podman.socket is gone.
> [...]
> crc 1.19 (OpenShift 4.6.3) has podman-1.9.3-3.rhaos4.6.el8.x86_64 which also has both socket files.

praveenkumar:
Checked from 4.6 side looks like it is present
/unhold
/lgtm

PR #294, opened 2020-12-10
Add monitoring operator in the CVO overrides before the cluster starts

guillaumerose:
The operator will never run and won't create what we used to delete.
It is cleaner and speed up a little bit cluster creation.

A patch release in 4.6 branch extracted the CRDs from the bindata of
prometheus operator. All required CRDs are now present by default even
if the monitoring operator never ran.

The procedure to activate monitoring is to remove the ClusterOperator
and the Deployment from overrides.

This method can be reused for other operators. At the end, all overrides
should added before the cluster starts.

### DIFF
--- a/cvo-overrides-after-first-run.yaml
+++ b/cvo-overrides-after-first-run.yaml
@@ -82,10 +82,25 @@
     namespace: ""
     unmanaged: true
   # only used in bootstrap phase
+  - kind: ClusterOperator
+    group: config.openshift.io/v1
+    name: machine-api
+    namespace: ""
+    unmanaged: true
+  - kind: ClusterOperator
+    group: config.openshift.io/v1
+    name: machine-config
+    namespace: ""
+    unmanaged: true
   - kind: Deployment
     group: apps/v1
     name: machine-config-operator
     namespace: openshift-machine-config-operator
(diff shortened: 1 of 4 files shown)


guillaumerose:
It requires https://github.com/openshift/cluster-version-operator/pull/488

cvo-overrides-after-first-run.yaml
```diff
@@ -3,54 +3,91 @@
...
-    name: machine-config-operator
-    namespace: openshift-machine-config-operator
+    name: cluster-monitoring-operator
+    namespace: openshift-monitoring
     unmanaged: true
-  - kind: Deployment
-    group: apps/v1
-    name: etcd-quorum-guard
-    namespace: openshift-machine-config-operator
+  - kind: ClusterOperator
```
praveenkumar, line 9: @guillaumerose do we need those `ClusterOperators` as `kind` in the override list also, doesn't setting up the deployment as unmanagement  works?

guillaumerose: yes it is needed as the installer is waiting for all cluster operators to be ready.

guillaumerose:
It's in master now.

PR #296, opened 2020-12-15
Remove /var/log/*/*.log after all components are stopped

guillaumerose:
Fixes https://github.com/code-ready/snc/issues/295

It will remove around 300MB of logs uncompressed.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -81,6 +81,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl 
 # Remove pull secret from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
 
+# Remove audit logs
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
+
 if [[ ${OKD_VERSION} != "none" ]]
 then
     # Install the hyperV and libvarlink-util rpms to VM


createdisk.sh
```diff
@@ -81,6 +81,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl
 # Remove pull secret from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
 
+# Remove audit logs
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/log/kube-apiserver/*'
```
praveenkumar, line 85: @guillaumerose what if we just delete everything from `/var/log` does those directory needs to be present for some container mount?

guillaumerose: good point, will try!

guillaumerose:
It works fine if I remove everything. PTAL.

createdisk.sh
```diff
@@ -81,6 +81,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl
 # Remove pull secret from the VM
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
 
+# Remove audit logs
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
```
praveenkumar, line 85: 👍🏼

praveenkumar (approved):

createdisk.sh
cfergeau: Looks like the comment/commit log were not updated when this PR changed from removing audit logs to removing /var/log?

PR #297, opened 2020-12-17
Use double quotes for variables which are passed to functions

praveenkumar:
If a variable's value contain the space then only first part of the
value taken by function, better to use double quotes to avoid unwanted
issues.


cfergeau (commented):
what about the other shell scripts? Do they need this too?

praveenkumar:
@cfergeau yes they might be needing it, right now I just added for some of the functions at then end of create disk but I think a clean solution would have a linter which can part of CI and block the PR if veriable doesn't follow the rules. something like 
https://github.com/koalaman/shellcheck one.

guillaumerose (approved):

PR #299, opened 2020-12-17
Implement 2 suggestions from rhbz#1868262

cfergeau:
This patch series implements 2 improvements which were suggested
by Miciah in https://bugzilla.redhat.com/show_bug.cgi?id=1868262#c5
There should be no behaviour changes, but snc.sh should be a bit more robust
after these.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+
+# Force yq download
+export YQ=./yq
+
 ./snc.sh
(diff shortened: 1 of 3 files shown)


cfergeau:
CI fails because its `yq` binary is too old:
```
 + replace_pull_secret crc-tmp-install-data/install-config.yaml
+ set +x
Error: unknown flag: --style
Usage:
  yq write [yaml_file] [path] [value] [flags] 
```

PR #302, opened 2020-12-21
Prepare 4.7 branch

guillaumerose:
* Disable LUKS
* Use latest preview

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -25,8 +25,9 @@ CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
+OPENSHIFT_VERSION=pre-release
 
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 # Otherwise use the tagged version if available
--- a/snc.sh
+++ b/snc.sh
@@ -70,7 +70,6 @@ fi
 
 if test -z "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE-}"; then
     OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/release.txt" | sed -n 's/^Pull From: //p')"
-    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE
 elif test -n "${OPENSHIFT_VERSION-}"; then
     echo "Both OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE and OPENSHIFT_VERSION are set, OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE will take precedence"
     echo "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE: $OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE"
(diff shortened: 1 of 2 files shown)


snc.sh
```diff
@@ -25,7 +25,7 @@ CRC_VM_NAME=${CRC_VM_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/pre-release}
```
praveenkumar, line 28: I would go with `https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview`

snc.sh
```diff
@@ -34,7 +34,7 @@ if test -n "${OPENSHIFT_VERSION-}"; then
     OPENSHIFT_RELEASE_VERSION=${OPENSHIFT_VERSION}
     echo "Using release ${OPENSHIFT_RELEASE_VERSION} from OPENSHIFT_VERSION"
 else
-    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/candidate-4.6/release.txt | sed -n 's/^ *Version: *//p')"
+    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/release.txt | sed -n 's/^ *Version: *//p')"
```
praveenkumar, line 37: Then this should be `$(curl -L "${MIRROR}"/pre-release/release.txt`

snc.sh
```diff
@@ -47,13 +47,13 @@ fi
 mkdir -p openshift-clients/linux openshift-clients/mac openshift-clients/windows
 if [[ ${OKD_VERSION} != "none" ]]
 then
-    curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-linux-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/linux oc
-    curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/mac oc
-    curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-windows-${OPENSHIFT_RELEASE_VERSION}.zip" > openshift-clients/windows/oc.zip
+    curl -L "${MIRROR}/openshift-client-linux-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/linux oc
```
praveenkumar, line 50: This should be `curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-linux.tar.gz"` , and we can remove this `OKD` check here because now we have the tarball without release version stuff on ocp side.

snc.sh
```diff
@@ -69,8 +69,7 @@ elif [ ! -f ${OPENSHIFT_PULL_SECRET_PATH} ]; then
 fi
 
 if test -z "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE-}"; then
-    OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/release.txt" | sed -n 's/^Pull From: //p')"
-    export OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE
+    OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$(curl -L "${MIRROR}/release.txt" | sed -n 's/^Pull From: //p')"
```
praveenkumar, line 72: This shouldn't be changed at all.

praveenkumar (changes requested):

snc.sh
guillaumerose: done!

snc.sh
guillaumerose: done!

snc.sh
guillaumerose: yes

snc.sh
guillaumerose: I set OPENSHIFT_VERSION directly to pre-release. The change is smaller with that.

guillaumerose:
Argh, there is no CI on branch 4.7! Maybe we should add it for all our branches?

guillaumerose:
Let's stick on master for now. I will reopen with what can be accepted.

PR #307, opened 2020-12-30
Extract baremetal installer using a one-liner

noseka1:
This way of extracting the baremetal installer is also used in the
official docs:
https://docs.openshift.com/container-platform/4.6/installing/installing_bare_metal_ipi/ipi-install-installation-workflow.html#extracting-the-openshift-installer_ipi-install-installation-workflow


guillaumerose:
/ok-to-test

Thanks for your PR!

PR #309, opened 2021-01-19
Clean up child processes more reliably

cfergeau:
The current code didn't seem to clean up children on exit. Also,
if there were no children to kill, the following error message
showed up on the console:
./snc.sh: line 1: jobs: unbound variable

This fixes both things by:
- replacing TERM with EXIT
- adding a check that 'jobs' is defined before trying to use it

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -9,7 +9,7 @@ source tools.sh
 source snc-library.sh
 
 # kill all the child processes for this script when it exits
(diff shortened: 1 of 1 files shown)


cfergeau:
This is a different attempt at fixing the issue described in https://github.com/code-ready/snc/pull/306

noseka1:
I tested the fix and it works for me. If this pull request is merged, the original pull request https://github.com/code-ready/snc/pull/306 can be closed.

guillaumerose:
Thanks for your fix and feedback @noseka1 !
/lgtm

PR #310, opened 2021-01-20
Rebased and updated logic for latest 4.7 release bits

praveenkumar:
This is just an update/rebase of #294

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -19,10 +19,9 @@ if [[ ${OKD_VERSION} != "none" ]]
 then
     OPENSHIFT_VERSION=${OKD_VERSION}
     BASE_OS=fedora-coreos
-    USE_LUKS=false
 fi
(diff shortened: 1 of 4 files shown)


cfergeau:
What is the plan? Merge this in a 4.7 branch? Or keep it in a PR until 4.7 is merged?

guillaumerose:
Maybe we can support both with a serie of `if version starts with 4.6 ...`. We can branch now also as I would expect little changes in 4.6 now.

praveenkumar:
4.6 changes part is already part of 4.6 branch and master is ready to take 4.7 stuff the only thing is we do need to update this once the actual rc  happen and artifacts available on `<MIRROR>/ocp` instead  `<MIRROR>/ocp-dev-preview`. I don't see any requirement to keep 4.6 and 4.7 both should build from master. If there is some PR which goes to master and required for 4.6 we will just cherrypick it to 4.6 branch.

PR #311, opened 2021-01-20
Change clusterNetwork CIDR once more

cfergeau:
This was initially changed because it was causing issues when running
crc in an OpenShift cluster.
However the new one causes collisions in a different environment.

Let's try a 3rd different value and see if we are lucky this time :)

I picked:
- an odd number
- bigger than 200

in the hope that sysadmins are less likely to see this as a good value
to use for internal networks.

Note: no problem dropping this if we feel there is no added value :)

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -19,13 +19,13 @@ metadata:
   name: crc
 networking:
   clusterNetwork:
-  - cidr: 10.116.0.0/14
+  - cidr: 10.217.0.0/22
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
   networkType: OpenShiftSDN
   serviceNetwork:
-  - 172.25.0.0/16
+  - 10.217.4.0/23
 platform:
   libvirt:
     URI: qemu+tcp://192.168.122.1/system


praveenkumar:
@cfergeau Looks like even installer not happy with this change `./openshift-baremetal-install --dir crc-tmp-install-data create manifests
level=fatal msg="failed to fetch Master Machines: failed to load asset \"Install Config\": invalid \"install-config.yaml\" file: networking.clusterNetwork[0].cidr: Invalid value: \"10.217.0.0/14\": invalid network address. got 10.217.0.0/14, expecting 10.216.0.0/14"`

guillaumerose:
`10.217.0.0/14` is not the "good" name of the network. The starting address of this subnet is 10.216.0.0.
We should not use a /14, it's way too big. /24 is already enough IP for all our pods.

cfergeau:
> `10.217.0.0/14` is not the "good" name of the network. The starting address of this subnet is 10.216.0.0.
> We should not use a /14, it's way too big. /24 is already enough IP for all our pods.
> 
> The service network can also be used in a /24. It could be the next one after the pod subnet.
> With this, we could say 10.216.0.0/2**3** is reserved for CRC.

Fwiw these parameters are described in https://github.com/openshift/installer/blob/master/docs/user/customization.md#platform-customization
By default, OpenShift allocates a /23 per node from `clusterNetwork` for pod IPs, I'd prefer to keep this unchanged. Since we have a single node, I agree we don't need the whole /14. Let's try to have `clusterNetwork` as `10.217.0.0/23` and `serviceNetwork` as `10.217.2.0/23`, which would mean crc reserves `10.217.0.0/22`?

cfergeau:
> By default, OpenShift allocates a /23 per node from `clusterNetwork` for pod IPs, I'd prefer to keep this unchanged. Since we have a single node, I agree we don't need the whole /14. Let's try to have `clusterNetwork` as `10.217.0.0/23` and `serviceNetwork` as `10.217.2.0/23`, which would mean crc reserves `10.217.0.0/22`?

CI failed with this, trying a less constrained IP address space.

guillaumerose (approved):

PR #314, opened 2021-01-27
Remove SNC_VALIDATE_CERT env from CI

praveenkumar:
This was helpful because initially we didn't have the way to do
cert rotation instead waiting for 24 hour, which is changed recently
and now it is not required.

 commit-id 9155345

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -24,7 +24,6 @@ done
 
 set -exuo pipefail
 # Run createdisk script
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
It's more than "not required", snc no longer uses that variable at all, so it's unused, and we forgot to remove the last reference to it from ci.sh

PR #315, opened 2021-01-27
Set kubelet node-ip with an environment file

guillaumerose:
OpenShift 4.6 added the flag --node-ip in the kubelet systemd unit.
SNC was already adding it at the very beginning. It works most of the
time but I suspect it is just luck and it might fail if the dummy
interface is not ready or so.

This change forces the node-ip to the value we want. The suggested env
file in the kubelet configuration is not working as expected.
I use here the same workaround as the bugzilla 1872632.

Related to https://github.com/code-ready/crc/issues/1888

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -153,7 +153,11 @@ ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${BASE_OS}-${ostree_h
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo nmcli conn add type dummy ifname eth10 con-name internalEtcd ip4 ${INTERNAL_IP}/24  && sudo nmcli conn up internalEtcd"
 
 # Add internalIP as node IP for kubelet systemd unit file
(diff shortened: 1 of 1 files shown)


cfergeau (approved):

guillaumerose:
Just took the bundle created by the CI and:
```
$ ps aux
root        2680  4.0  1.1 1865612 97284 ?       Ssl  19:17   0:00 kubelet --config=/etc/kubernetes/kubelet.conf --bootstrap-kubeconfig=/etc/kubernetes/kubeconfig --kubeconfig=/var/lib/kubelet/kubeconfig --container-runtime=remote --container-runtime-endpoint=/var/run/crio/crio.sock --runtime-cgroups=/system.slice/crio.service --node-labels=node-role.kubernetes.io/master,node.openshift.io/os_id=rhcos --node-ip=192.168.126.11 --minimum-container-ttl-duration=6m0s --cloud-provider= --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec --register-with-taints=node-role.kubernetes.io/master=:NoSchedule --pod-infra-container-image=quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:3f502978600415a084a7c99950a0573339104169b8cbc0b68dcc565f15a38854 --v=2
```

looks good!

PR #317, opened 2021-01-29
Use built-in crictl image pruning feature

guillaumerose:


cfergeau (commented):
Is the behaviour exactly the same as the previous method, are there going to be some small differences? Can you indicate this in the commit log?

guillaumerose:
yes done. No differences. The flag was not present at the time we added that.

https://github.com/code-ready/snc/pull/143

> crictl version used in the RHCOS is 0.1.0 which doesn't have fix
> for kubernetes-sigs/cri-tools#399 one yet so
> using kubernetes-sigs/cri-tools#399 (comment) as workaround.

cfergeau (approved):

gbraad (approved):

PR #320, opened 2021-02-02
createdisk: Compress bundle using zstd rather than xz

cfergeau:
It's both more efficient space-wise and faster to decompress:

   When compressed with `zstd --ultra -22 -T0`, the 4.6.9 bundle is 2.2G in 
   size, and uncompresses in about 1 minute. The xz-compressed bundle is 2.6G
   in size, and takes 3 minutes to uncompress.

   This requires crc-side support: https://github.com/code-ready/crc/pull/1905


guillaumerose:
With this bundle, on my machine: 1min36
With 4.6.9 bundle: 4min20.

guillaumerose (approved):

Issue #321, opened 2021-02-02
Bundle name is incorrect in json for hyperkit and hyper-v bundles


guillaumerose:
```
$ cat ~/.crc/cache/crc_hyperkit_4.6.9/crc-bundle-info.json 
{
  "version": "1.1",
  "type": "snc",
..
  "name": "crc_libvirt_4.6.9",
..
  "driverInfo": {
    "name": "hyperkit"
}
```

The name should be `crc_hyperkit_4.6.9`.

praveenkumar:
resolved by #322

PR #322, opened 2021-02-02
#321 set the correct bundle name for non-libvirt bundles

guillaumerose:
Hyperkit and hyper-v directories are generated based on libvirt
directory.
The JSON name was set once, when creating libvirt directory.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -189,6 +189,7 @@ function generate_hyperkit_directory {
 
     # Update the bundle metadata info
     cat $srcDir/crc-bundle-info.json \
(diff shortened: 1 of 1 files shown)


createdisk-library.sh
```diff
@@ -189,6 +189,7 @@ function generate_hyperkit_directory {
 
     # Update the bundle metadata info
     cat $srcDir/crc-bundle-info.json \
+        | ${JQ} ".name = \"${destDir}\"" \
```
praveenkumar, line 192: 👍🏼

praveenkumar (approved):

PR #305, opened 2020-12-23
Optimize ppc64le (request for comments)

mtarsel:
Since creating tools and library files, this commit attempts to address the multi-arch optimizations for ppc64le, aka POWER hardware. 

- Create download_oc_binary() in snc-library to obtain only the binaries needed for the specific arch. Saves some time in snc.sh
- Creates prepare_hyperV() in createdisk-library to optimize crc bundle creation for ppc64le. Even tho these hyperV changes are not going to happen on ppc64le, the master VM will still be restarted.
- Made a TODO in shellcheck.sh since there is not a stable release of shellcheck for ppc64le. Will be needed for ci in the future.
- Added comment about unzip package since only needed for windows.

Addresses PR #263 

I'm hoping this PR can start the discussion about how to address multi-arch support going forward with the 4.7 release utilizing the power box at osu. I'm available to chat :)

PR #323, opened 2021-02-02
Optimize ppc64le v2

mtarsel:
Addresses the original PR in #263 and the edits made in the PR #305.

Creates prepare_hyperV() and download_oc() to help speed up execution on ppc64le.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -172,7 +172,7 @@ function copy_additional_files {
 
 function prepare_hyperV() {
 
-    if [ "${ARCH}" != "ppc64le" ]; then
+    if [ "${ARCH}" == "ppc64le" ]; then
         echo "hyperv is not available on ppc64le"
         return
     fi
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -170,6 +170,45 @@ function copy_additional_files {
     eventually_add_pull_secret $destDir
 }
 
+function prepare_hyperV() {
+
+    if [ "${ARCH}" != "ppc64le" ]; then
+        echo "hyperv is not available on ppc64le"
+        return
+    fi
+
+    if [[ ${OKD_VERSION} != "none" ]]; then
+        # Install the hyperV and libvarlink-util rpms to VM
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install --allow-inactive hyperv-daemons libvarlink-util'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora-updates.repo'
+    else
+        # Download the hyperV daemons and libvarlink-util dependency on host
+        mkdir $1/packages
+        sudo yum install -y --downloadonly --downloaddir $1/packages hyperv-daemons libvarlink-util
+
+        # SCP the downloaded rpms to VM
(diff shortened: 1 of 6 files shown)


snc-library.sh
```diff
@@ -10,6 +10,30 @@ function preflight_failure() {
         fi
 }
 
+function download_oc() {
+    if [[ ${OKD_VERSION} != "none" ]]; then
```
praveenkumar, line 14: we don't need `OKD_VERSION` check here because it is same for OCP and OKD, you just need to stop downloading the oc for mac and windows side.

createdisk-library.sh
```diff
@@ -172,7 +172,7 @@ function copy_additional_files {
 
 function prepare_hyperV() {
 
-    if [ "${ARCH}" != "ppc64le" ]; then
+    if [ "${ARCH}" == "ppc64le" ]; then
```
praveenkumar, line 175: shouldn't this be part of first commit where this logic is added?

mtarsel: yea i could squash this if needed - i messed this up

snc-library.sh
mtarsel: oh ok - i was following comment from previous PR here: https://github.com/code-ready/snc/pull/305#discussion_r551803581

cfergeau: Since this comment, Praveen unified the 2 codepaths in commit 3ce9669d8a7

createdisk-library.sh
```diff
@@ -170,6 +170,45 @@ function copy_additional_files {
...
+        # Install the hyperV and libvarlink-util rpms to VM
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install --allow-inactive hyperv-daemons libvarlink-util'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora-updates.repo'
+    else
+        # Download the hyperV daemons and libvarlink-util dependency on host
+        mkdir $1/packages
+        sudo yum install -y --downloadonly --downloaddir $1/packages hyperv-daemons libvarlink-util
```
cfergeau, line 190: This is reverting the changes from https://github.com/code-ready/snc/commit/24f2971a964d9961bdb7fc646d56bb93419bdb79

cfergeau (commented):
This seems to have been based off a slightly older version of snc, so some parts need to be updated. Instead of adding some ppc64le checks in several places, I think it makes more sense to define  `GENERATE_MACOS_BUNDLE` and `GENERATE_WINDOWS_BUNDLE` variables which are set to 1/0 depending on the arch, and to use these in the rest of the scripts.
This will make it easier to adjust the script for s390 (which is != ppc64le I believe?), for aarch64, ... I can do that in a follow-up after this PR has been merged though.

mtarsel:
Oh geez, sorry about pushing outdated code. I like the idea of creating some variable in tools.sh for MacOS, Windows, and POWER hardware. tools.sh seems to be the best place for this since its source'd in everything else. This is turning into a bigger "commit" than I thought but I think separating the archs and OS environments would be beneficial for future development. I will be careful about how I separate this work for the sake of the commit history.

One idea is to create the `env_bit` variable which is passed to functions as needed. This way we can pass the same variable to functions and in that function we have a case or if statement to execute code for that arch and OS.

Alternatively, yes we have different variables defined in tools.sh and if that variable is 1 or 0 we run that code. I'm ok with this approach too. Here's another idea:

```
tools.sh:


ARCH=$(uname -m)

case $ARCH in
    ppc64le)
        set env_bit=ppc64le;
    x86_64)
        check_windows_or_mac()
        if windows = true
            set env_bit=windows
        else:
            set env_bit=macos; 
    *)
 	echo "The host arch is ${ARCH}. This is not supported by SNC!";;
esac

```

so then elsewhere we can just pass the $env_bit variable and do the following:

```
if $env_bit = windows :
   prepare_hyperV

if $env_bit = macos:
   prepare_hyperkit
```

With this MacOS, Windows, ppc64le approach it raises a question that I don't have much experience with. I have only been focused on differences in architecture but is there a clean easy way (or a preferred way for you guys) to determine if it is either MacOS or Windows? 

I can continue to work on this branch and revert these commits or just start a new branch. I like making new branches so it's cleaner on my box but that seems to open a new PR so I think I'll just revert and approach this differently if thats ok. Thanks a lot for the feedback on this. Other things have settled down for 4.7 so I'm focusing on this for now.

PR #326, opened 2021-02-04
Add CRC_ZSTD_EXTRA_FLAGS env variable to change zstd options for CI

praveenkumar:
This option will help us to provide different flag for zstd when running
in the CI to generate bundles fast.


cfergeau (approved):
Wondering if we should default to something fast for casual testing, and only force expensive compression for release bundles? But fine with me this way.

PR #328, opened 2021-02-09
Create $ENV_TYPE and function download_oc() 

mtarsel:
The ENV_TYPE variable is initialized based off the type of kernel running in order to determine if this is Linux, Windows, or Mac OS.

This is not the same value as the global env $OSTYPE.

download_oc() downloads the specific oc binary for the OS that is running. $OC is still set in snc.sh with the Linux oc file

Since UNZIP is only needed to extract the windows oc binary, first check if the environment is WINDOWS in tools.sh

Sorry I had to make another branch - my git tree was messed up. This PR is to address the discussion here: https://github.com/code-ready/snc/pull/323#pullrequestreview-582275789

### DIFF
--- a/tools.sh
+++ b/tools.sh
@@ -16,6 +16,19 @@ CRC_ZSTD_EXTRA_FLAGS=${CRC_ZSTD_EXTRA_FLAGS:-"--ultra -22"}
 
 ARCH=$(uname -m)
 
+# Set the environment type based off the kernel name. 
+# NOTE: the kernel type is not the same as $OSTYPE
+case "$(uname -s)" in
+    Darwin*|darwin*)
+        ENV_TYPE="MACOS";;
+    Linux)
+        ENV_TYPE="LINUX";;
+    CYGWIN*|MINGW32*|MSYS*|MINGW*)
+        ENV_TYPE="WINDOWS";;
+    *)
+        echo "SNC runs on Linux, Windows, or Mac OS."
+        return 1;;
+esac
 
 yq_ARCH=${ARCH}
 # yq and install_config.yaml use amd64 as arch for x86_64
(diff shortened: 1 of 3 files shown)


tools.sh
```diff
@@ -16,6 +16,19 @@ CRC_ZSTD_EXTRA_FLAGS=${CRC_ZSTD_EXTRA_FLAGS:-"--ultra -22"}
...
+    Darwin*|darwin*)
+        ENV_TYPE="MACOS";;
+    Linux)
+        ENV_TYPE="LINUX";;
+    CYGWIN*|MINGW32*|MSYS*|MINGW*)
+        ENV_TYPE="WINDOWS";;
+    *)
+        echo "SNC runs on Linux, Windows, or Mac OS."
+        return 1;;
+esac
```
praveenkumar, line 31: This is going to break our current CI and also downstream CI because we are generating all the bundles from linux host and with this we are not able to have mac and windows bits. I thought we are going to add env `CRC_MAC_BUNDLE` and `CRC_WINDOWS_BUNDLE` to check if those are required and have those value true by default.

cfergeau: Yup, there seems to have been a misunderstanding here, we never run `snc.sh` on macos/windows, always on linux. It's really depending on the arch that we are going to be able to generate macos/windows bundles or not.

cfergeau: I've done what I had in mind in https://github.com/code-ready/snc/pull/329 Only lightly tested, especially the `ARCH != x86_64` case.

mtarsel: ok - i understand now. Thanks for pushing that other PR. I really wasn't aware of your CI environment but that makes sense. I'll test that PR today and post comments there. I appreciate this feedback, thanks guys.

cfergeau: Closing this in favour of #329

PR #329, opened 2021-02-09
Add SNC_GENERATE_{MACOS,WINDOWS}_BUNDLE

cfergeau:
On most arches, it makes no sense to generate macos and/or Windows bundles.
This commit adds SNC_GENERATE_{MACOS,WINDOWS}_BUNDLE variables which will
only be set x86_64.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -170,6 +170,38 @@ function copy_additional_files {
     eventually_add_pull_secret $destDir
 }
 
+function prepare_hyperV() {
+    if [[ ${OKD_VERSION} != "none" ]]; then
+        # Install the hyperV and libvarlink-util rpms to VM
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rpm-ostree install --allow-inactive hyperv-daemons libvarlink-util'
+        ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora.repo'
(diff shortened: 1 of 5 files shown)


cfergeau:
/hold
At this point it's a proof of concept I'd like to run through CI/show to @mtarsel

createdisk-library.sh
```diff
@@ -177,6 +177,7 @@ function generate_hyperkit_directory {
     local kernel_release=$4
     local kernel_cmd_line=$5
 
+    mkdir "$destDir"
```
praveenkumar, line 180: Shouldn't the function also renamed to `generate_hyperkit_bundle` instead `generate_hyperkit_directory` here?

cfergeau: ah thanks, forgot to change it :)

mtarsel:
snc.sh runs successfully on ppc64le with this patch. Thanks for sharing, this logic is cleaner :) 

When executing the createdisk.sh I got this error...

```
+ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc core@api.crc.testing -- 'sudo rm -f /var/lib/kubelet/config.json'
+ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc core@api.crc.testing -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
+ [[ none != \n\o\n\e ]]
+ mkdir crc-tmp-install-data//packages
+ sudo yum install -y --downloadonly --downloaddir crc-tmp-install-data//packages hyperv-daemons
Last metadata expiration check: -1 day, 2:31:51 ago on Wed Feb 10 13:47:58 2021.
No match for argument: hyperv-daemons
Error: Unable to find a match: hyperv-daemons
[root@pkvmci824 snc]#
```

which I pushed in a sloppy PR previously here: https://github.com/code-ready/snc/pull/323/commits/869a2f44135860dfe12aa93acdac0ca23c6ad861


So something like this should be added

```
if [ -n "${SNC_GENERATE_WINDOWS_BUNDLE}" ]; then
    prepare_hyperV()

```
which will handle that OKD_VERSION check and the hyperV v-sock support.

cfergeau:
> So something like this should be added
> 
> ```
> if [ -n "${SNC_GENERATE_WINDOWS_BUNDLE}" ]; then
>     prepare_hyperV()
> ```
> 
> which will handle that OKD_VERSION check and the hyperV v-sock support.

Indeed, thanks, I've taken your preexisting patch for this, and slightly tweaked it to give proper credit for this change. I can squash it in the initial commit if you prefer.

mtarsel:
Great - thank you very much for pushing that. This patch works on ppc64le

mtarsel:
oh and thanks a lot for giving me credit in the patch too. I don't have a preference if it's squashed or not. Going forward, I'll use this approach for ppc64le support so I'm sure there will be more patches :)

PR #330, opened 2021-02-09
Make explicit in the console that this login method is for developer

guillaumerose:
Screenshot from 2021-02-09 20-05-18

vs 

Screenshot from 2021-02-09 14-33-49

"developer" instead of htpasswd_provider

### DIFF
--- a/htpasswd_cr.yaml
+++ b/htpasswd_cr.yaml
@@ -4,7 +4,7 @@ metadata:
   name: cluster
 spec:
   identityProviders:
-  - name: htpasswd_provider 
+  - name: developer
     mappingMethod: claim 
     type: HTPasswd
     htpasswd:


gbraad (approved):

gbraad:
is this used anywhere else as a fixed value? or this is just 'informative' ?
anyway, that makes sense

gbraad:
this might have been described in the documentation

htpasswd_cr.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: cluster
 spec:
   identityProviders:
-  - name: htpasswd_provider 
+  - name: developer only
```
praveenkumar, line 7: I never thought it would be that simple :)

praveenkumar (approved):

htpasswd_cr.yaml
cfergeau: does this mean I should only click on this if I'm a crc/openshift developer? Or should I click on it if I want to login as the `developer` user?

guillaumerose: Yes naming is hard. Only if you want to login with the developer account.

cfergeau: `login as 'developer'` or `'developer' login`?

guillaumerose: There is already the prefix "Log in with". Maybe just say "developer" and nothing else?

cfergeau: Sounds good, or `'developer' user`

guillaumerose:
There is a mention of this button in the doc in crc also. Thanks @gbraad 

```
docs/source/topics/proc_accessing-the-openshift-web-console.adoc
16:. Choose the **htpasswd_provider** option in the OpenShift web console.
```

gbraad:
Create an issue for this on CRC and link this pr, so we know when to merge and for @robin-owen to track

PR #332, opened 2021-02-10
Issue 331 Extend token validity to infinite 

praveenkumar:
Since crc cluster is for development purpose, better to not expire
the token when it generated.


guillaumerose:
Does that mean we can avoid in crc to call authentication service to get a token and use one generated at snc build time ?

praveenkumar:
> Does that mean we can avoid in crc to call authentication service to get a token and use one generated at snc build time ?

@guillaumerose we are not generating as part of snc (because we are using `system:admin` end to end)  but even it is generated the auth service need to be running to make use of it.

guillaumerose:
Yes sure but I would like to avoid crc to vendor part of openshift/oc code to do the login. 
If I can take 2 predefined tokens from the bundle, it would help a lot.

praveenkumar:
> Yes sure but I would like to avoid crc to vendor part of openshift/oc code to do the login.
> If I can take 2 predefined tokens from the bundle, it would help a lot.

What we can do in that case create a new `kubeconfig_user` which have token for `kube:admin` and `developer` user and ship it as part of the bundle?

cfergeau:
> Yes sure but I would like to avoid crc to vendor part of openshift/oc code to do the login.
> If I can take 2 predefined tokens from the bundle, it would help a lot.

The kubeadmin password is the same for all clusters, it's not regenerated each time we recreate a cluster, is it? Reason I'm asking is that we were asked to recreate ssh keys each time we recreate a new cluster, rather than using one created at bundle generation time. I was about to write that the tokens should be handled in a similar way, but then I realized the kubeadmin password is the same for all instances?
Next question would be if that's a good thing or not :)

guillaumerose:
First, this PR is I think a good thing for the user. Let's fix the security issue in a follow-up.
/lgtm

New issue to discuss: https://github.com/code-ready/snc/issues/333

gbraad:
> The kubeadmin password is the same for all clusters, it's not regenerated each time we recreate a cluster, is it?

This is from an earlier discussion again... Let's follow up on #333 for this. This was also part of a security analysis that concluded this wasn't ideal.

cfergeau:
> First, this PR is I think a good thing for the user. Let's fix the security issue in a follow-up.

Yup, I was reacting to your suggestion to move all of this to `snc`, which might not be a good thing. But this PR does not make things worse  from that perspective.

gbraad (approved):

PR #336, opened 2021-02-16
Update yq version to 4.5.1

praveenkumar:


snc-library.sh
```diff
@@ -98,7 +98,7 @@ function replace_pull_secret() {
         # get leaked in CI logs
         set +x
         local filename=$1
-        ${YQ} write --inplace $filename --style literal pullSecret "$(< $OPENSHIFT_PULL_SECRET_PATH)"
+	pull_secret="$(< $OPENSHIFT_PULL_SECRET_PATH)" ${YQ} eval --inplace '.pullSecret = strenv(pull_secret) | .pullSecret style="literal"' $filename
```
cfergeau, line 101: not sure why there is a `|  .pullSecret style="literal"`? what does it mean/do?

snc.sh
```diff
@@ -120,30 +120,30 @@ fi
 
 # Create the INSTALL_DIR for the installer and copy the install-config
 rm -fr ${INSTALL_DIR} && mkdir ${INSTALL_DIR} && cp install-config.yaml ${INSTALL_DIR}
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml compute[0].architecture ${yq_ARCH}
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml controlPlane.architecture ${yq_ARCH}
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml baseDomain ${BASE_DOMAIN}
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml metadata.name ${CRC_VM_NAME}
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml compute[0].replicas 0
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret "@HIDDEN_PULL_SECRET@"
+yq_ARCH=${yq_ARCH} ${YQ} eval --inplace '.compute[0].architecture = strenv(yq_ARCH)' ${INSTALL_DIR}/install-config.yaml
```
cfergeau, line 123: It's no longer possible at all to directly use variables in the `yq` command line? Something like `${YQ} eval --inplace ".compute[0].architecture = $yq_ARCH" ${INSTALL_DIR}/install-config.yaml`
Imo this deserves a mention in the commit log

snc.sh
```diff
@@ -120,30 +120,30 @@ fi
...
+yq_ARCH=${yq_ARCH} ${YQ} eval --inplace '.controlPlane.architecture = strenv(yq_ARCH)' ${INSTALL_DIR}/install-config.yaml
+base_domain=${BASE_DOMAIN} ${YQ} eval --inplace '.baseDomain = strenv(base_domain)' ${INSTALL_DIR}/install-config.yaml
+crc_vm_name=${CRC_VM_NAME} ${YQ} eval --inplace '.metadata.name = strenv(crc_vm_name)' ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace '.compute[0].replicas = 0' ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace '.pullSecret = "@HIDDEN_PULL_SECRET@"' ${INSTALL_DIR}/install-config.yaml
 replace_pull_secret ${INSTALL_DIR}/install-config.yaml
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml sshKey "$(cat id_ecdsa_crc.pub)"
+sshKey=$(cat id_ecdsa_crc.pub) ${YQ} eval '.sshKey = strenv(sshKey)' --inplace ${INSTALL_DIR}/install-config.yaml
```
cfergeau, line 130: Does this need `""` around `$(cat id_ecdsa_crc.pub)`?

tools.sh
```diff
@@ -32,7 +32,7 @@ esac
 # Download yq/jq for manipulating in place yaml configs
 if ! "${YQ}" -V; then
     if [[ ! -e yq ]]; then
-        curl -L https://github.com/mikefarah/yq/releases/download/3.3.0/yq_linux_${yq_ARCH} -o yq
+        curl -L https://github.com/mikefarah/yq/releases/download/v4.5.1/yq_linux_${yq_ARCH} -o yq
```
cfergeau, line 35: What about users who already have yq installed? Is the new syntax you are using backwards compatible with 3.x? Or does `yq` have to be replaced before snc.sh can run?

snc-library.sh
cfergeau: ah `|` is a pipe, not a 'or'. Still not sure why this is needed :)

snc.sh
```diff
@@ -120,30 +120,30 @@ fi
...
 
 # Create the manifests using the INSTALL_DIR
 OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests || exit 1
 
 # Add CVO overrides before first start of the cluster. Objects declared in this file won't be created.
-${YQ} merge -i ${INSTALL_DIR}/manifests/cvo-overrides.yaml cvo-overrides.yaml
+${YQ} eval-all --inplace 'select(fileIndex == 0) * select(filename == "cvo-overrides.yaml")' ${INSTALL_DIR}/manifests/cvo-overrides.yaml cvo-overrides.yaml
```
cfergeau, line 136: Some references to the new documentation in the commit log would be useful, https://mikefarah.gitbook.io/yq/v/v4.x/upgrading-from-v3#merging-documents for example.

cfergeau (commented):
yq v4 brings significant syntax changes, we must have some very good reason for doing this change, it would be good to have this documented in the commit log.

tools.sh
praveenkumar: @cfergeau I will try to change it something we are doing for `openshift-install` command, always download if not specified.

snc.sh
praveenkumar: Ah yes, good idea since the output might have space in between.

snc.sh
praveenkumar: It is not unfortunately, I can mention the docs around env variable this in the commit log.

snc-library.sh
praveenkumar: This is required because of using different style then default one `style="literal"`

snc.sh
cfergeau: Note that in `${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml metadata.name ${CRC_VM_NAME}`, it is the shell which replaces `$CRC_VM_NAME` with its value before running `${YQ}`, so from `YQ` perspective, there is no variable involved. Are you sure this was not some escaping-related issues that you had?
```
export SSH_KEY=foo
./yq_linux_amd64 eval ".sshKey = \"$SSH_KEY\"" ./install-config.yaml
```
adds `.sshKey=foo` to install-config.yaml

cfergeau: k, the doc link that you gave says
```
While environment variables can, of course, be passed in via your CLI with string interpolation, this often comes with complex quote escaping and can be tricky to write and read. 
```

It's the fact that we need to add `yq_ARCH=${yq_ARCH}`  before the yq call which bothers me (it's because the variable is not exported that we need to do that). Maybe in our case it's good enough to use shell variable interpoalation rather than `strenv`?

snc.sh
```diff
@@ -49,14 +57,6 @@ OC=./openshift-clients/linux/oc
 
 run_preflight_checks
```
cfergeau, line 58: same could be said about the preflight checks, maybe the pull secret check should be moved there, and the preflights moved before oc downloading?

createdisk.sh
```diff
@@ -73,15 +73,6 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable chronyd
...
 
-# Remove all the pods except openshift-sdn from the VM
-pods=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo crictl pods -o json | jq '.items[] | select(.metadata.namespace != "openshift-sdn")' | jq -r .id)
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo crictl stopp "${pods}" || true"
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "for i in {1..3}; do sudo crictl rmp "${pods}" && break || sleep 2; done || true"
-
-# Remove openshift-sdn pods also from the VM
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl stopp $(sudo crictl pods -q) || true'
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmp $(sudo crictl pods -q) || true'
-
```
cfergeau, line 84: ` because now there is crio-wipe service` 

What does "now" mean? Is this 4.7 specific or has this been in OpenShift releases for a while?

praveenkumar: @cfergeau It is there for a while now, iirc from the 4.5 side.

snc.sh
praveenkumar: @cfergeau shell variable interpolation you mean using quote escaping?

snc.sh
praveenkumar: Sure I can try to do that, and yes pref light should run before.

snc.sh
cfergeau: The shell replacing `$SSH_KEY` on the command line with the value of `$SSH_KEY` is what is called 'shell interpolation' in `yq` documentation (I think :). In order to use this instead of `strenv`, some quote escaping has to be done yes.

snc-library.sh
cfergeau, line 109: `pull_secret="$(< $OPENSHIFT_PULL_SECRET_PATH)" ` might be slightly easier to read as `pull_secret="$(cat $OPENSHIFT_PULL_SECRET_PATH)" `

snc.sh
```diff
@@ -120,30 +120,30 @@ fi
...
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml pullSecret "@HIDDEN_PULL_SECRET@"
+${YQ} eval --inplace ".compute[0].architecture = \"${yq_ARCH}\"" ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace ".controlPlane.architecture = \"${yq_ARCH}\"" ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace ".baseDomain = \"${BASE_DOMAIN}\"" ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace ".metadata.name = \"${CRC_VM_NAME}\"" ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace '.compute[0].replicas = 0' ${INSTALL_DIR}/install-config.yaml
+${YQ} eval --inplace '.pullSecret = "@HIDDEN_PULL_SECRET@"' ${INSTALL_DIR}/install-config.yaml
 replace_pull_secret ${INSTALL_DIR}/install-config.yaml
-${YQ} write --inplace ${INSTALL_DIR}/install-config.yaml sshKey "$(cat id_ecdsa_crc.pub)"
+${YQ} eval ".sshKey = \"$(cat id_ecdsa_crc.pub)\"" --inplace ${INSTALL_DIR}/install-config.yaml
```
cfergeau, line 121: `eval --inplace` for consistentcy with the other lines?

cfergeau (approved):

PR #337, opened 2021-02-17
Use installer wait-for functionality to wait till cluster become stable

praveenkumar:
Lot of changes and patches are applied to the cluster after it
provisioned. This patch will make sure to cluster again become stable
after all those patches/changes.

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -246,10 +246,62 @@ function delete_operator() {
         local namespace=$2
         local pod_selector=$3
 
-	retry ${OC} get pods
+        retry ${OC} get pods
         pod=$(${OC} get pod -l ${pod_selector} -o jsonpath="{.items[0].metadata.name}" -n ${namespace})
 
         retry ${OC} delete ${delete_object} -n ${namespace}
         # Wait until the operator pod is deleted before trying to delete the resources it manages
         ${OC} wait --for=delete pod/${pod} --timeout=120s -n ${namespace} || ${OC} delete pod/${pod} --grace-period=0 --force -n ${namespace} || true
 }
+
+function all_operators_available() {
+    # Check the cluster operator output, status for available should be true for all operators
+    # The single line oc get co output should not contain any occurrences of `False`
+    ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Available")].status}' | grep -v False
+}
+
+function no_operators_progressing() {
+    # Check the cluster operator output, status for progressing should be false for all operators
+    # The single line oc get co output should not contain any occurrences of `True`
+    ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Progressing")].status}' | grep -v True
+}
+
+function no_operators_degraded() {
+    # Check the cluster operator output, status for available should be false for all operators
+    # The single line oc get co output should not contain any occurrences of `True`
+    ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Degraded")].status}' | grep -v True
+}
+
+function wait_till_cluster_stable() {
+    sleep 1m
+
+    local retryCount=30
+    local numConsecutive=3
+    local count=0
+
+    # Remove all the failed Pods
+    ${OC} delete pods --field-selector=status.phase=Failed -A || true
+
+    local a=0
+    while [ $a -lt $retryCount ]; do
+       if all_operators_available && no_operators_progressing && no_operators_degraded; then
+           echo "All operators are available. Ensuring stability ..."
+           count=$((count + 1))
+       else
+           echo "Some operators are still not available"
+           count=0
+       fi
+       if [ $count -eq $numConsecutive ]; then
+           echo "Cluster has stabilized"
+           break
+       fi
+       sleep 30s
+       a=$((a + 1))
+    done
+
+    # Wait till all the pods are either running or complete state
+    while oc get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed; do
+       sleep 2
+    done
+}
+
--- a/snc.sh
+++ b/snc.sh
@@ -27,6 +27,7 @@ CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
+DEVELOPER_USER_PASS='developer:$2y$05$paX6Xc9AiLa6VT7qr2VvB.Qi.GJsaqS80TR3Kb78FEIlIL0YyBuyS'
 
 run_preflight_checks
 
@@ -196,6 +197,17 @@ retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"
 
 # Delete the pods which are there in Complete state
 retry ${OC} delete pod --field-selector=status.phase==Succeeded --all-namespaces
+
+# Add a user developer:developer with htpasswd identity provider and give it sudoer role
+retry ${OC} create secret generic htpass-secret --from-literal=htpasswd=${DEVELOPER_USER_PASS} -n openshift-config
+retry ${OC} apply -f oauth_cr.yaml
+retry ${OC} create clusterrolebinding developer --clusterrole=sudoer --user=developer
+
+# Replace pull secret with a null json string '{}'
+retry ${OC} replace -f pull-secret.yaml
+
+# Remove the Cluster ID with a empty string.
+retry ${OC} patch clusterversion version -p '{"spec":{"clusterID":""}}' --type merge
 
 # Wait for the cluster again to become stable because of all the patches/changes
 wait_till_cluster_stable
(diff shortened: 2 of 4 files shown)


guillaumerose (approved):

snc-library.sh
```diff
@@ -253,3 +253,52 @@ function delete_operator() {
...
 }
+
+function operators_available() {
+        # Check the cluster operator output, status for available should be true for all operators
+        ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Available")].status}' | grep -v False
+}
+
+function operators_progressing() {
+        # Check the cluster operator output, status for progressing should be false for all operators
+        ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Progressing")].status}' | grep -v True
```
cfergeau, line 264: Sorry, won't have time for a proper review right now, but the logic seems off in these `operators_*()` functions. They check that that at least one operator status is not `False` or `True` depending on the function. I tried with the following test case, hopefully I did not make mistakes
```
function test_grep_false() {
	grep -v False <<EOF
1: False
2: False
3: True
EOF
}

function test_grep_true() {
	grep -v True <<EOF
1: False
2: False
3: True
EOF
}

if test_grep_false; then 
	echo "if (test_grep_false) succeeded"
else 
	echo "if (test_grep_false) failed"
fi

echo ""

if test_grep_true; then 
	# the test fails if all operators have a "True" status
	# the test succeeds if at least one operator is "False"
	echo "if (test_grep_true) succeeded"
else 
	echo "if (test_grep_true) failed"
fi
```

cfergeau: Regarding naming of the functions, when the comment is `# Check the cluster operator output, status for progressing should be false for all operators ` , the function could be named "no_operators_progressing"?

snc-library.sh
```diff
@@ -253,3 +253,52 @@ function delete_operator() {
...
+function operators_degraded() {
+        # Check the cluster operator output, status for available should be false for all operators
+        ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Degraded")].status}' | grep -v True
+}
+
+function wait_till_cluster_stable() {
+	sleep 1m
+
+        local retryCount=30
+        local numConsecutive=3
```
cfergeau, line 276: indentation is not the same for `sleep` and `local`? Why start with a sleep when we have more sophisticated logic for waiting?

snc-library.sh
```diff
@@ -253,3 +253,52 @@ function delete_operator() {
...
+           else
+               echo "Still some operators are not available"
+               count=0
+           fi
+           if [ $count -eq $numConsecutive ]; then
+               echo "Cluster is stabilize"
+               break
+           fi
+           sleep 30s
+	   a=$((a + 1))
```
cfergeau, line 296: indentation?

snc-library.sh
```diff
@@ -253,3 +253,52 @@ function delete_operator() {
...
+        while [ $a -lt $retryCount ]; do
+           if operators_available && operators_progressing && operators_degraded; then
+               echo "All operators are available. Ensuring stability ..."
+               count=$((count + 1))
+           else
+               echo "Still some operators are not available"
+               count=0
+           fi
+           if [ $count -eq $numConsecutive ]; then
+               echo "Cluster is stabilize"
```
cfergeau, line 292: `is stable` / `has stabilized`

snc-library.sh
```diff
@@ -253,3 +253,52 @@ function delete_operator() {
...
+        # Remove all the failed Pods
+        ${OC} delete pods --field-selector=status.phase=Failed -A || true
+
+        local a=0
+        while [ $a -lt $retryCount ]; do
+           if operators_available && operators_progressing && operators_degraded; then
+               echo "All operators are available. Ensuring stability ..."
+               count=$((count + 1))
+           else
+               echo "Still some operators are not available"
```
cfergeau, line 288: `Some operators are still not available`

snc.sh
```diff
@@ -197,5 +198,16 @@ retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"
...
+retry ${OC} create clusterrolebinding developer --clusterrole=sudoer --user=developer
+
+# Replace pull secret with a null json string '{}'
+retry ${OC} replace -f pull-secret.yaml
+
+# Remove the Cluster ID with a empty string.
+retry ${OC} patch clusterversion version -p '{"spec":{"clusterID":""}}' --type merge
+
 # Wait for the cluster again to become stable because of all the patches/changes
 wait_till_cluster_stable
```
cfergeau, line 210: I would move `${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet` right after `wait_till_cluster_stable`, this guarantees there won't be more cluster changes after this point.

cfergeau (commented):
Logic of the helpers does not seem correct in the first patch, othterwise looks fine.

snc-library.sh
praveenkumar: @cfergeau I agree about the naming and after experimenting lot of different stuff looks like following is provide us correct result. Here you can change any value and see if the cluster status. If you have better suggestion how to achieve in bash then I would like to explore that. (I am not very good in bash)

```bash
#!/bin/bash

set -exuo pipefail

export LC_ALL=C.UTF-8
export LANG=C.UTF-8

function all_available() {
        grep -v True >/dev/null <<EOF
True True True True
EOF
}

function none_progressing() {
        grep -v False >/dev/null <<EOF
False False False True
EOF
}

function none_degrading() {
        grep -v False >/dev/null <<EOF
True False False False
EOF
}

function check_cluster() {
        all_available
        a=$?
        none_progressing
        b=$?
        none_degrading
        c=$?
        echo $((a + b + c))
}

if [[ $(check_cluster) -ne 3 ]]; then
        echo "cluster unstable"
else
        echo "cluster stable"
fi
```

cfergeau: Ah ok, you get a single line from `oc`, not multiple lines as in my test case. So my test case is indeed incorrect :)
For `none_progressing`/`none_degrading`, we want to check that the line we are processing does not contain any `True` occurrences. `grep True` will tell us if it does contain `True`, and `grep -v True` will tell us if there are no occurrences of `True`, so I think`none_progressing`/`none_degrading` should be doing this rather than `grep -v False`.

Similarly, for `all_available`, we want to make sure there are no `False` occurrences on the line, so `grep -v False`.

In short, your initial code seems correct apart from the naming, and my comments were misleading as I was not testing with the correct data

```bash
function all_operators_available() {
        # Check the cluster operator output, status for available should be true for all operators
        ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Available")].status}' | grep -v False
}

function no_operators_progressing() {
        # Check the cluster operator output, status for progressing should be false for all operators
        ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Progressing")].status}' | grep -v True
}

function no_operators_degraded() {
        # Check the cluster operator output, status for available should be false for all operators
        ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Degraded")].status}' | grep -v True
}
```

snc.sh
praveenkumar: @cfergeau I am not moving the kubelet stop part in here because what if someone just run snc and try out the cluster?

snc-library.sh
```diff
@@ -242,14 +242,63 @@ function renew_certificates() {
...
-        pod=$(${OC} get pod -l ${pod_selector} -o jsonpath="{.items[0].metadata.name}" -n ${namespace})
+    retry ${OC} get pods
+    pod=$(${OC} get pod -l ${pod_selector} -o jsonpath="{.items[0].metadata.name}" -n ${namespace})
 
-        retry ${OC} delete ${delete_object} -n ${namespace}
-        # Wait until the operator pod is deleted before trying to delete the resources it manages
-        ${OC} wait --for=delete pod/${pod} --timeout=120s -n ${namespace} || ${OC} delete pod/${pod} --grace-period=0 --force -n ${namespace} || true
+    retry ${OC} delete ${delete_object} -n ${namespace}
+    # Wait until the operator pod is deleted before trying to delete the resources it manages
+    ${OC} wait --for=delete pod/${pod} --timeout=120s -n ${namespace} || ${OC} delete pod/${pod} --grace-period=0 --force -n ${namespace} || true
```
cfergeau, line 254: Does not belong in this commit

snc-library.sh
```diff
@@ -242,14 +242,63 @@ function renew_certificates() {
...
 }
+
+function all_operators_available() {
+    # Check the cluster operator output, status for available should be true for all operators
```
cfergeau, line 258: might be easier to understand that code in the future if you add after your comment "so the single line `oc get co` output should not contain any occurrences of `False"?

snc-library.sh
```diff
@@ -242,14 +242,63 @@ function renew_certificates() {
...
+    # Remove all the failed Pods
+    ${OC} delete pods --field-selector=status.phase=Failed -A || true
+
+    local a=0
+    while [ $a -lt $retryCount ]; do
+       if all_operators_available && no_operators_progressing && no_operators_degraded; then
+           echo "All operators are available. Ensuring stability ..."
+           count=$((count + 1))
+       else
+           echo "Still some operators are still not available"
```
cfergeau, line 288: One extra 'Still"

snc.sh
cfergeau: can go at the very beginning of `createdisk.sh` then, with a big comment saying cluster changes must not be done as part of createdisk? (maybe it's already this way, did not check). Can be done in a followup though.

cfergeau (commented):
Minor comments left on the first commit which would be nice to address, apart from this looks good to me.

cfergeau (approved):

PR #348, opened 2021-03-02
Add kubeadmin user as part of htpasswd

praveenkumar:
To remove the temp kube:admin user we need to have another user
which have cluster admin role as per documents[0]. This patch is
going to add kubeadmin user and also update the authfile for same.
It will also remove the temp kube:admin user.

[0] https://docs.openshift.com/container-platform/latest/authentication/remove-kubeadmin.html

### DIFF
--- a/tools.sh
+++ b/tools.sh
@@ -13,6 +13,8 @@ UNZIP=${UNZIP:-unzip}
 ZSTD=${ZSTD:-zstd}
 CRC_ZSTD_EXTRA_FLAGS=${CRC_ZSTD_EXTRA_FLAGS:-"--ultra -22"}
 
+HTPASSWD=${HTPASSWD:-htpasswd}
+
 ARCH=$(uname -m)
 
 case "${ARCH}" in
@@ -77,6 +79,10 @@ if ! which ${ZSTD}; then
     sudo yum -y install /usr/bin/zstd
 fi
 
+if ! which ${HTPASSWD}; then
+    sudo yum -y install /usr/bin/htpasswd
+fi
+
 function retry {
     local retries=10
     local count=0
@@ -126,6 +132,14 @@ function start_vm {
         echo " ${vm_prefix}-master-0 still booting"
         sleep 2
     done
+}
+
+function generate_htpasswd_file {
+   local auth_file_dir=$1
+   local pass_file=$2
+   random_password=$(cat $1/auth/kubeadmin-password)
+   ${HTPASSWD} -c -B -b ${pass_file} developer developer
+   ${HTPASSWD} -B -b ${pass_file} kubeadmin ${random_password}
 }
 
 # Restart the libvirt service after update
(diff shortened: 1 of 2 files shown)


guillaumerose:
Works great for me! It removes the choice for auth. method in the console. It's really a good improvement!

/lgtm

guillaumerose:
I have an error with crc: `WARN Cannot update kubeconfig: challenger chose not to retry the request`. I am looking at it.

tools.sh
```diff
@@ -128,5 +139,16 @@ function start_vm {
...
 
+function generate_htpasswd_file {
+   local auth_file_dir=$1
+   local pass_file=$2
+   random_password=$(${OPENSSL} rand -base64 14)
+   ${HTPASSWD} -c -B -b ${pass_file} developer developer
+   ${HTPASSWD} -B -b ${pass_file} kubeadmin ${random_password}
+
+   # Update kubeadmin-password file
+   echo ${random_password} > $1/auth/kubeadmin-password
```
guillaumerose, line 150: The newline at the end of the password is unexpected. crc takes the \n has a character of the password. 

I guess you need to use `echo -n`.

guillaumerose: I opened this https://github.com/code-ready/crc/pull/2046, it also fixes the problem.

praveenkumar: @guillaumerose sure we can control the way of putting the password in the file.

tools.sh
```diff
@@ -128,5 +139,16 @@ function start_vm {
     done
 }
 
+function generate_htpasswd_file {
+   local auth_file_dir=$1
+   local pass_file=$2
+   random_password=$(${OPENSSL} rand -base64 14)
```
cfergeau, line 145: Could be `$ head -c14 /dev/random |base64` to avoid the need for openssl

cfergeau (approved):

tools.sh
```diff
@@ -128,5 +134,13 @@ function start_vm {
     done
 }
 
+function generate_htpasswd_file {
+   local auth_file_dir=$1
+   local pass_file=$2
+   random_password=$(cat $1/auth/kubeadmin-password)
```
cfergeau, line 140: Can you call this `kubeadmin_password` or such? `random_password` when we don't generate it was confusing to me.

tools.sh
```diff
@@ -128,5 +134,13 @@ function start_vm {
...
+   ${HTPASSWD} -c -B -b ${pass_file} developer developer
+   ${HTPASSWD} -B -b ${pass_file} kubeadmin ${random_password}
```
cfergeau, line 142: Or you could remove `${random_password}` and use something like:
```
cat  $1/auth/kubeadmin-password) | ${HTPASSWD} -i -B -b ${pass_file} kubeadmin
```

cfergeau (approved):

PR #349, opened 2021-03-02
Use systemd unit file to create dummy network

praveenkumar:
In 4.8 nightly builds, it is observed that the changes for dummy
network not presisted during the reboot. Looking into the networkManager
logs doesn't provide any suspious activity. This patch will create
a systemd unit file which suppose to create the dummy network for
every start.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -100,7 +100,22 @@ kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cm
 ${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${BASE_OS}-${ostree_hash}/* $1
 
 # Add a dummy network interface with internalIP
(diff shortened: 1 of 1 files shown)


cfergeau:
Has there been a definitive answer explaining why the previous approach is no longer working with 4.8? Without knowing why this is broken, this feels more like a workaround for an undetermined issue than a fix.

praveenkumar:
> Has there been a definitive answer explaining why the previous approach is no longer working with 4.8? Without knowing why this is broken, this feels more like a workaround for an undetermined issue than a fix.

@cfergeau no definitive answer, I tried whatever conversation we had on slack but doesn't find the reason why it is happening.

cfergeau (commented):
I'd change `which suppose to create the dummy network for
every start.` to `which will invoke nmcli recreate the dummy interface each time the system boots`

Let's go with this to unblock CI, but I really hope we can find the root cause for this problem so that we can make an informed decision how to fix this.

cfergeau (approved):

Issue #335, opened 2021-02-16
Update yq to version 4


praveenkumar:
Latest version of yq allow us to first check the key if exist before updating/removing it. We observered that in one of the CI run the kubelet.conf file was empty but then also createdisk process didn't exit.

cfergeau:
Isn't it a bug that kubelet.conf is empty/createdisk does not exit?

PR #351, opened 2021-03-04
Add a check for systemReserved key in kubelet

praveenkumar:
It is Observered that in one of the CI run the kubelet.conf file
was empty but then also createdisk process didn't exit because `yq`
doesn't check the key exist during the delete.

fixes: #335


createdisk.sh
```diff
@@ -47,6 +47,10 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo podman pull quay.io/crcont
 # Stop the kubelet service so it will not reprovision the pods
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl stop kubelet
 
+# Check if system reserved block exist in kubelet config
```
cfergeau, line 50: It seems this is used to detect some buggy situation when the kubelet config file was empty? Would be useful to mention this in the comment rather than repeating what the line below already says :)

PR #360, opened 2021-03-11
Use machineconfig resource to create dummy network service

praveenkumar:


99-openshift-machineconfig-master-dummy-networks.yaml
```diff
@@ -0,0 +1,31 @@
...
+      units:
+      - contents: |
+          [Unit]
+          Description=Create dummy network
+          After=NetworkManager.service
+          
+          [Service]
+          Type=oneshot
+          RemainAfterExit=yes
+          ExecStart=/bin/nmcli conn add type dummy ifname eth10 autoconnect yes save yes con-name internalEtcd ip4 192.168.126.11/24
```
cfergeau, line 21: I was expecting this would be used to create a persistent NM configuration file in `/etc/NetworkManager/...`?

praveenkumar: @cfergeau I was trying to do follow but that was not working.

```
# cat  <<EOF > /etc/NetworkManager/system-connections/test.nmconnection
[connection]
id=test
type=dummy
interface-name=eth11
permissions=

[dummy]

[ipv4]
address1=192.168.124.11/24
dns-search=
method=manual

[ipv6]
addr-gen-mode=stable-privacy
dns-search=
method=auto

[proxy]
EOF

# systemctl reload NetworkManager
#  ifconfig eth11
eth11: error fetching interface information: Device not found
```

Even https://docs.fedoraproject.org/en-US/fedora-coreos/static-ip-config/ does provide such options :(

cfergeau: Make sure the file permissions are 0400. I needed that before NetworkManager used the file.

PR #373, opened 2021-03-19
Remove OPENSHIFT_INSTALL_EXPERIMENTAL_CLUSTER_PROFILE env for installer

praveenkumar:
single-node-developer profile is not part of some of the operators and
we recently noticed that it is causing the issue for dns operator where
openshift/cluster-dns-operator#216 is still not
merged and because of that the dns operator manifest doesn't part of
CVO.

This caused issue for upstream CI and really became hard to debug so
this patch will make sure we don't use any profile and go with default.

- BZ https://bugzilla.redhat.com/show_bug.cgi?id=1940309


praveenkumar:
/hold

wait till the CI passes

PR #380, opened 2021-03-31
Add 'additionalFiles' field to crc-bundle-info.json

cfergeau:
Rather than hardcoding in crc that the bundle contains the oc binary, this
adds a section in the crc-bundle-info.json file to indicate its presence.
This should help when a recent crc binary tries to load an older bundle,
see https://github.com/code-ready/crc/issues/2145


createdisk-library.sh
```diff
@@ -116,6 +119,10 @@ function update_json_description {
         | ${JQ} '.storage.diskImages[0].format = "qcow2"' \
         | ${JQ} ".storage.diskImages[0].size = \"${diskSize}\"" \
         | ${JQ} ".storage.diskImages[0].sha256sum = \"${diskSha256Sum}\"" \
+        | ${JQ} ".storage.additionalFiles[0].name = \"oc\"" \
+        | ${JQ} '.storage.additionalFiles[0].type = "oc-executable"' \
```
praveenkumar, line 123: @cfergeau do we really need this `oc-executable` here if in the template it is same for all platform?

test-metadata-generation.sh
```diff
@@ -0,0 +1,78 @@
...
+# If the user set OKD_VERSION in the environment, then use it to override OPENSHIFT_VERSION, set BASE_OS, and set USE_LUKS
+# Unless, those variables are explicitly set as well.
+OKD_VERSION=${OKD_VERSION:-none}
+if [[ ${OKD_VERSION} != "none" ]]
+then
+    OPENSHIFT_VERSION=${OKD_VERSION}
+    BASE_OS=fedora-coreos
+fi
+BASE_OS=${BASE_OS:-rhcos}
+USE_LUKS=${USE_LUKS:-false}
```
praveenkumar, line 18: We can remove the `USE_LUKS` variable now since it is now false for RHCOS and also for FCOS.

test-metadata-generation.sh
```diff
@@ -0,0 +1,78 @@
...
+OKD_VERSION=${OKD_VERSION:-none}
+if [[ ${OKD_VERSION} != "none" ]]
+then
+    OPENSHIFT_VERSION=${OKD_VERSION}
+    MIRROR=${MIRROR:-https://github.com/openshift/okd/releases/download}
+fi
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+OPENSHIFT_RELEASE_VERSION="${OPENSHIFT_VERSION-4.7.0}"
+
+# CRC_VM_NAME: short VM name to use in crc_libvirt.sh
```
praveenkumar, line 31: `short VM name to use in crc_libvirt.sh` can be now just `short VM name` ?

test-metadata-generation.sh
```diff
@@ -0,0 +1,78 @@
...
+    | ${JQ} ".buildInfo.openshiftInstallerVersion = \"0.0.0\"" \
+    | ${JQ} ".buildInfo.sncVersion = \"xxx\"" \
+    | ${JQ} ".clusterInfo.openshiftVersion = \"${OPENSHIFT_RELEASE_VERSION}\"" \
+    | ${JQ} ".clusterInfo.clusterName = \"${CRC_VM_NAME}\"" \
+    | ${JQ} ".clusterInfo.baseDomain = \"${BASE_DOMAIN}\"" \
+    | ${JQ} ".clusterInfo.appsDomain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" >${srcDir}/crc-bundle-info.json
+
+mkdir -p "$destDir/linux"
+#mkdir -p "$destDir/macos"
+#mkdir -p "$destDir/windows"
```
praveenkumar, line 69: Should this be un-commented ?

crc-bundle-info.json.sample
```diff
@@ -70,6 +70,16 @@
 	"size": "9129426944"
 	"sha256sum": "49766122a0834d62c1a24fb4e0de30cd7a39b8112083aa5e01fc26f16c15aed3"
       }
+    ],
+    "additionalFiles": [
```
guillaumerose, line 74: would it be possible to put all files here instead of just additional ? Maybe as a follow-up.  I understand in shell, it will be real pain to write :/

guillaumerose:
Should the format version be bumped ?

createdisk-library.sh
cfergeau: I can see this also being used for `podman`, and the name is platform-dependant (because of the `.exe` extension on Windows), so this 'type' gives us a generic identifier which will be the same on all platforms. Without it we'll have to make some guesses.

test-metadata-generation.sh
cfergeau: Ah I'll look at adding an additional patch for this.

test-metadata-generation.sh
cfergeau: Ah no, it should be removed, `generate_{hyperkit,hyperv}_bundle` start with `mkdir "$destDir"`

crc-bundle-info.json.sample
cfergeau: The other files are already referenced in other places so I felt less need for it
```
    # Name of a file containing an SSH private key which can be used to connect to
    # the cluster nodes
    "sshPrivateKeyFile": "id_ecdsa_crc",
    # Name of the kubeconfig file stored in the bundle
    "kubeConfig": "kubeconfig",
    # Name of the file containing the kubeadmin password for use in the
    # openshift console
    "kubeadminPasswordFile": "kubeadmin-password"
    # pull secret that can be used to fetch OpenShift container images (optional)
    # "openshiftPullSecret": "default-pull-secret"
[...]
      # details about its format
      "diskImage": "crc.qcow2"
```

Adding them to 'additionalFiles' would be redundant, but it would also allow us to give checksums and various metadata about these files , which can be a good thing. We actually already have the same redundancy with the qcow2 image, `"diskImage": "crc.qcow2"` is a key into the `"diskImages"` array which gives more details about the actual image.

So yeah, we could add more files in there, but I'll keep that as a future improvement :)
Maybe I should use a different name than `additionalFiles` though if we want to use it more widely.

cfergeau:
> Should the format version be bumped ?

Hmm that's a good question. the key being present or not is a good indication of a new enough format. This would allow to improve the check in https://github.com/code-ready/crc/pull/2174/files though:
```

func (bundle *CrcBundleInfo) createSymlinkOrCopyOpenShiftClient(ocBinDir string) error {
	ocInBundle := bundle.resolvePath(constants.OcExecutableName)
	ocInBundle := bundle.GetOcPath()
	// For backward-compatibility with bundles which shipped an oc binary in the bundle
	// before additionalFiles was added to crc-bundle-info.json
	if ocInBundle == "" {
		ocInBundle = bundle.resolvePath(constants.OcExecutableName)
		if !crcos.FileExists(ocInBundle) {
			// This is not an error, bundles for OpenShift 4.5 and older did not contain oc
			return nil
		}
	}
	ocInBinDir := filepath.Join(ocBinDir, constants.OcExecutableName)
```

I'll bump it for once, good opportunity to try this.

crc-bundle-info.json.sample
cfergeau: I renamed the key to the more generic `fileList`

PR #384, opened 2021-04-06
Minor fixes

praveenkumar:
- Remove 'exit 1' for 'or' bash operator
 snc.sh script run with `set -exuo pipefail` which automatic exit as
soon as any command fails so no need to have specific exit for some
commands.

- Update ignition version for chronyd machine config resource

### DIFF
--- a/99_master-chronyd-mask.yaml
+++ b/99_master-chronyd-mask.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   config:
     ignition:
-      version: 3.1.0
+      version: 3.2.0
     systemd:
       units:
        - name: chronyd.service
--- a/snc.sh
+++ b/snc.sh
@@ -123,7 +123,7 @@ replace_pull_secret ${INSTALL_DIR}/install-config.yaml
 ${YQ} eval ".sshKey = \"$(cat id_ecdsa_crc.pub)\"" --inplace ${INSTALL_DIR}/install-config.yaml
 
 # Create the manifests using the INSTALL_DIR
-${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests || exit 1
+${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests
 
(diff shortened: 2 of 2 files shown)


cfergeau (approved):
Why is it needed to change the ignition version?

praveenkumar:
> Why is it needed to change the ignition version?

@cfergeau it is just to sync with what ignition version currently 4.8 is using as part of MCO.
```
$ oc get clusterversion
NAME      VERSION                             AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.8.0-0.nightly-2021-04-05-174735   True   

$ oc get mc
NAME                                               GENERATEDBYCONTROLLER                      IGNITIONVERSION   AGE
00-master                                          86270f3375f894ff1dc21eee74247f04790dd0e1   3.2.0             3h24m
00-worker                                          86270f3375f894ff1dc21eee74247f04790dd0e1   3.2.0             3h24m
01-master-container-runtime                        86270f3375f894ff1dc21eee74247f04790dd0e1   3.2.0             3h24m
01-master-kubelet                                  86270f3375f894ff1dc21eee74247f04790dd0e1   3.2.0             3h24m
01-worker-container-runtime                        86270f3375f894ff1dc21eee74247f04790dd0e1   3.2.0             3h24m

```

cfergeau:
> > Why is it needed to change the ignition version?
> 
> @cfergeau it is just to sync with what ignition version currently 4.8 is using as part of MCO.
> 

If it's just to sync the version number with what 4.8 is using, but both ignition versions work the same way, I'd mention this in the commit log. It's useful for example to know it should not be backported to the 4.7 branch.

praveenkumar:
> If it's just to sync the version number with what 4.8 is using, but both ignition versions work the same way, I'd mention this in the commit log. It's useful for example to know it should not be backported to the 4.7 branch.

@cfergeau I think even 4.7 uses `3.2.0`, it's 4.6 which had `3.1.0` and it is backward compatible.

Issue #383, opened 2021-04-06
Use second in the date timestamp when creating the directory


praveenkumar:
Since we are using a same GCP bucket for all the bundles produced by snc/installer PR and right now the bundles directory structure depend on the https://github.com/code-ready/snc/blob/master/createdisk-library.sh#L14 date which means new bundles override old one if produced on same day so better to include seconds field as part of directory creation.

cfergeau:
Maybe it's possible to add the PR number in the directory structure? Only keeping the latest bundles for a PR should be good enough, and will save some space.

praveenkumar:
@cfergeau yes that also a possible solution, will  try that.

PR #385, opened 2021-04-06
Use PR number for bundle destination directory

praveenkumar:
'PULL_NUMBER' is part of prow job env variables and can be used for
step registry in openshift CI. It will help us to have only latest bundle
for each PR.

- https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables

fixes: #383

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -7,6 +7,7 @@ function get_dest_dir {
     if [ "${OPENSHIFT_VERSION-}" != "" ]; then
         DEST_DIR=$OPENSHIFT_VERSION
     else
+        DEST_DIR=${PULL_NUMBER}
         set +e
         DEST_DIR=$(git describe --exact-match --tags HEAD)
         set -e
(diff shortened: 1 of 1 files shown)


createdisk-library.sh
```diff
@@ -7,6 +7,7 @@ function get_dest_dir {
     if [ "${OPENSHIFT_VERSION-}" != "" ]; then
         DEST_DIR=$OPENSHIFT_VERSION
     else
+        DEST_DIR=${PULL_NUMBER}
         set +e
         DEST_DIR=$(git describe --exact-match --tags HEAD)
```
cfergeau, line 12: This overwrites the DEST_DIR you just set?

praveenkumar: @cfergeau Ah, anyway `DEST_DIR=$(git describe --exact-match --tags HEAD)` is not used anymore so better to remove it all together.

cfergeau (approved):

gbraad (approved):

Issue #393, opened 2021-04-14
Figure out if Marketplace update can be turned off


praveenkumar:
Looks like marketplace release is adhoc with openshift release and even we do latest z stream release and try that after an new marketplace release then it start pulling images from the internet and which is around ~500Mb data. We need to figure out if there is a way to turn off that auto update on marketplace operator side so it will not update automatic and use whatever we have at time of bundle creation.

praveenkumar:
After digging in the maketplace pod side looks like it using `imagePullPolicy: Always` which force to go to network and pull the metadata or images.
```
      imagePullPolicy: Always
      volumeMounts:
        - name: certified-operators-token-64gg9
          readOnly: true
          mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      terminationMessagePolicy: File
      image: 'registry.redhat.io/redhat/certified-operator-index:v4.7'
```

PR #395, opened 2021-04-19
Update registryPoll interval for catalog source

praveenkumar:
We are obervering that as soon as user start crc it pulls the
images for catalog sources which is part of openshift-maketplace.

This patch will change the registryPoll interval from 10m to 720h (30d)
to avoid this pull even image is updated on the registry.

fixes: #393

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -214,6 +214,12 @@ retry ${OC} patch clusterversion version -p '{"spec":{"clusterID":""}}' --type m
 retry ${OC} delete machineconfigs --all
 retry ${OC} delete machineconfigpools --all
 
+# Update the catalog source registryPoll interval
+retry ${OC} patch catalogsources -n openshift-marketplace certified-operators -p '{"spec":{"updateStrategy":{"registryPoll":{"interval":"7200h"}}}}' --type merge
+retry ${OC} patch catalogsources -n openshift-marketplace community-operators -p '{"spec":{"updateStrategy":{"registryPoll":{"interval":"7200h"}}}}' --type merge
+retry ${OC} patch catalogsources -n openshift-marketplace redhat-marketplace -p '{"spec":{"updateStrategy":{"registryPoll":{"interval":"7200h"}}}}' --type merge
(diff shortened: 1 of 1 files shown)


praveenkumar:
/hold

Need to clarify with openshift marketplace team once.

cfergeau:
If I want the latest version of an operator from the marketplace, will the catalog be downloaded on demand when needed? or will I always get an outdated version until I ran some command to update it?

praveenkumar:
>  If I want the latest version of an operator from the marketplace, will the catalog be downloaded on demand when needed? or will I always get an outdated version until I ran some command to update it?

@cfergeau you will suppose to get the same version which was during the bundle creation, if you need an recent one then you do need to update the interval again but I do need to verify it with marketplace team if this is correct assumption and fixes what @guillaumerose faces.

gbraad:
We know from feedback that people prefer a static image with a predictable setup. So if a user wants a new image this is OK to be a user's choice to deal with; so documentation is needed to describe the 'limitation'.

gbraad (approved):

gbraad:
Note: we might not want marketplace to be a standard part, but rather move to this the enable as addon like the telemetry/metrics option as we do with prometheus.

guillaumerose:
I am wondering if we should not just remove this update block. It is weird to trigger the update only after 30d. Why not 90d ?

gbraad:
> I am wondering if we should not just remove this update block. It is weird to trigger the update only after 30d. Why not 90d ?

I actually felt the same. Completely disabling or an extremely long interval, as 30 days will be reached for sure

guillaumerose:
What do we do with that @praveenkumar ? Do you think we can take this solution for the next release ?

praveenkumar:
> What do we do with that @praveenkumar ? Do you think we can take this solution for the next release ?

@guillaumerose I will need to update it as per suggestion but as per slack conversation with operator team looks like it only make sense to first disable default catalog sources and then add with modified resource which have modified version of registryPoll.

praveenkumar:
Closing this because there is no way we change the pod `imagePullPolicy` which is default set to `Always` so even if we change the interval the fresh start of pod going to registry and check for that image.

PR #398, opened 2021-04-21
Preload routes controller image

guillaumerose:
The feature was added in crc by https://github.com/code-ready/crc/pull/1793
The image digest needs to be sync between both projects.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -83,6 +83,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
   systemctl enable gvisor-tap-vsock.service
 EOF
 
+# Preload routes controller
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl pull quay.io/crcont/routes-controller:latest'
+
 # Shutdown and Start the VM after installing the hyperV daemon packages.
 # This is required to get the latest ostree layer which have those installed packages.
 shutdown_vm ${VM_PREFIX}


createdisk.sh
```diff
@@ -83,6 +83,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
   systemctl enable gvisor-tap-vsock.service
 EOF
 
+# Preload routes controller
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl pull quay.io/crcont/routes-controller:6effc2f0304d3a11c1e22c278b10630e834d0220'
```
gbraad, line 87: no version?

gbraad (changes requested):
Can we version this?

createdisk.sh
praveenkumar: We are using sha so do we still need version?

praveenkumar:
This image is part of pod so that means if anything change on the image we need to update it snc and also on crc side, would it be possible to have that info in metadata and on crc side we just consume it from metadata instead hardcoding there? That way it would be updated on single place.

guillaumerose:
I could use a generic tag inside the VM like `routes-controller:current`.
Or, in crc, list all images with `crictl images` and run the first one with routes-controller name. 
WDYT? I found that changing the json is a little bit painful each time.

praveenkumar:
> I could use a generic tag inside the VM like `routes-controller:current`.
> Or, in crc, list all images with `crictl images` and run the first one with routes-controller name.
> WDYT? I found that changing the json is a little bit painful each time.

@guillaumerose I would go with generic tag then because using `crictl images` and get the image ID with `route-controller` looks bit hacky. Generic tag looks like a clean way to avoid some situations.

guillaumerose:
I tried with :current tag. It works but I am worry that the user might delete the image from the bundle by mistake.
If the user does something like podman rmi $(podman images -q), the pod won't be able to start anymore as the tag is only local.

I propose to keep the duplication of version for now.. If the preloading is not right when we develop, it's not a big deal, just few MB to download. We just need to ensure at release time, versions are the same.

WDYT?

(cc from https://github.com/code-ready/crc/pull/2273#issuecomment-827384565)

createdisk.sh
guillaumerose: release created, v0.0.1

praveenkumar:
> I tried with :current tag. It works but I am worry that the user might delete the image from the bundle by mistake.
> If the user does something like podman rmi $(podman images -q), the pod won't be able to start anymore as the tag is only local.
> 

The way I thought it to have `latest` or `current` tag https://quay.io/repository/crcont/routes-controller?tab=tags which always points to latest release of the image (I created latest one which on the quay UI show it comes from v0.0.1 tag) . The benefit which we are going to have is we no longer worry about changing tag on `snc` side or on `crc` side for each router controller release.

guillaumerose:
If we change the behavior of the router and the user deletes the image, it will fetch it again and it might not work. 
I guess it is unlikely to happen but still. WDYT?

praveenkumar:
> If we change the behavior of the router and the user deletes the image, it will fetch it again and it might not work.
> I guess it is unlikely to happen but still. WDYT?

@guillaumerose  Yes and it is really unlikely since we are doing same with `dnsmasq` https://github.com/code-ready/snc/blob/master/createdisk.sh#L44 with latest and didn't get any issue from users. But we can argue that dnsmasq is not a suitable example here since we only care about dnsmasq package but here we are actually put router-controller code which developed by us. Still I am thinking to use latest is better than pointing version here because it will avoid us some of code maintainability because of this version saga on crc and snc repo :)
 
@code-ready/crc-team wdyt?

guillaumerose:
OK I change it to latest :+1:

createdisk.sh
```diff
@@ -83,6 +83,9 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
   systemctl enable gvisor-tap-vsock.service
 EOF
 
+# Preload routes controller
+${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl pull quay.io/crcont/routes-controller:latest'
```
praveenkumar, line 87: 👍🏼

praveenkumar (approved):

PR #400, opened 2021-04-26
Remove machineconfigs and machineconfigspools

praveenkumar:
We are already override the machine config operator so removing those
configs shouldn't affect the cluster and those are regenerated with
updated content as long as user enable it.


cfergeau (approved):
Maybe `as long as` -> `when` in the commit log.

Issue #403, opened 2021-04-29
Add exposed registry route CA to VM


praveenkumar:
We are exposing registry route `default-route-openshift-image-registry.apps-crc.testing` for the user to push or pull the images from openshift internal registry and use it to deploy their application. 

Current flow is user login to the registry and push the image to specific namespace. Then to use it with the application either they need to use internal registry URl `image-registry.openshift-image-registry.svc:5000` or enable the `image-lookup` as mentioned in the doc https://code-ready.github.io/crc/#accessing-the-internal-openshift-registry_gsg 

What we can do is use the registry route CA and add it to VM so it will not failed with following error.
```
Failed to pull image "default-route-openshift-image-registry.apps-crc.testing/zen/cpd-operator-init:v1.0.1-x86_64": rpc error: code = Unknown desc = error pinging docker registry default-route-openshift-image-registry.apps-crc.testing: Get https://default-route-openshift-image-registry.apps-crc.testing/v2/: x509: certificate signed by unknown authority
```

- https://github.com/code-ready/crc/issues/1376

PR #404, opened 2021-04-29
Add CA cert of exposed registry route to VM

praveenkumar:
We are exposing registry route `default-route-openshift-image-registry.apps-crc.testing`
for the user to push or pull the images from openshift internal registry
and use it to deploy their application.

Current flow is user login to the registry and push the image to specific namespace.
Then to use it with the application either they need to use internal registry URl
`image-registry.openshift-image-registry.svc:5000` or enable the image-lookup as
mentioned in the doc https://code-ready.github.io/crc/#accessing-the-internal-openshift-registry_gsg

This patch is use the registry route CA and add it to VM so it will not failed with following error in
case of application deployed using this registry.
```
Failed to pull image "default-route-openshift-image-registry.apps-crc.testing/zen/cpd-operator-init:v1.0.1-x86_64": rpc error: code = Unknown desc = error pinging docker registry default-route-openshift-image-registry.apps-crc.testing: Get https://default-route-openshift-image-registry.apps-crc.testing/v2/: x509: certificate signed by unknown authority
```

- https://github.com/code-ready/crc/issues/1376

fixes: #403

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -236,6 +236,11 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "/memory: /d" /opt/re
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "/memory: /d" /opt/release-manifests/*operator.yaml'
 retry ${OC} -n openshift-cluster-version patch deploy cluster-version-operator --type=json -p=$(cat custom-release.json | jq -c .)
 
+# Add exposed registry CA to VM
+retry ${OC} extract secret/router-ca --keys=tls.crt -n openshift-ingress-operator --confirm
+retry ${OC} create configmap registry-certs --from-file=default-route-openshift-image-registry.apps-crc.testing=tls.crt -n openshift-config
+retry ${OC} patch image.config.openshift.io cluster -p '{"spec": {"additionalTrustedCA": {"name": "registry-certs"}}}' --type merge
+
 # Wait for the cluster again to become stable because of all the patches/changes
 wait_till_cluster_stable
 


cfergeau:
Is `default-route-openshift-image-registry.apps-crc.testing` something created by default by `openshift-install`? I assume you add the certs manually with `ssh` because on a real cluster this is all managed by the MCO, but we can't use it  here?

praveenkumar:
> Is `default-route-openshift-image-registry.apps-crc.testing` something created by default by `openshift-install`?

This is auto created as soon as we expose the image registry https://github.com/code-ready/snc/blob/master/snc.sh#L193 here.

> I assume you add the certs manually with `ssh` because on a real cluster this is all managed by the MCO, but we can't use it here?

Yes, so in case of MCO we can add this registry as part of allowed registry with the CA and then MCO will take care of adding that CA at proper location, here we have to do manually.

praveenkumar:
/hold

Need to test generated bundle if that works as expected and if it does then I would like to have it as part of this release bundle.

cfergeau:
> > Is `default-route-openshift-image-registry.apps-crc.testing` something created by default by `openshift-install`?
> 
> This is auto created as soon as we expose the image registry https://github.com/code-ready/snc/blob/master/snc.sh#L193 here.
> 
> > I assume you add the certs manually with `ssh` because on a real cluster this is all managed by the MCO, but we can't use it here?
> 
> Yes, so in case of MCO we can add this registry as part of allowed registry with the CA and then MCO will take care of adding that CA at proper location, here we have to do manually.

Could we add it the right way before we remove/disable the MCO? I'm wondering if adding that default-route is the right thing to do, or if it would be better to have a dedicated crc command instead to mess with registries (as this significantly depart from documented openshift procedures).

praveenkumar:
> Could we add it the right way before we remove/disable the MCO?

@cfergeau we could but that will only works with 4.8 (need testing though) (since it have single node support), 4.7 it is not going to work.

praveenkumar:
> > > Could we add it the right way before we remove/disable the MCO?
> > 
> > 
> > @cfergeau we could but that will only works with 4.8 (need testing though) (since it have single node support), 4.7 it is not going to work.
> 
> Since there is already some MCO/4.8 work committed to master, shouldn't this PR also make use of the MCO to setup the registry in git master, with a different version in the 4.7 branch to account for the lack of MCO?

@cfergeau current changes are going to work with both 4.8 and 4.7. Looks like adding registry cert is not depend on MCO instead adding insecure registry does.

cfergeau (approved):

PR #406, opened 2021-05-03
Enable MCO from cluster operator

praveenkumar:


createdisk.sh
```diff
@@ -65,9 +65,6 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl disable chronyd
 # Enable the podman.socket service for API V2
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo systemctl enable podman.socket
 
-# Remove pull secret from the VM
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo rm -f /var/lib/kubelet/config.json'
-
```
cfergeau, line 70: I would expect to see the change to remove config.json with the MCO to be part of this commit?

cfergeau (commented):
Good to be able to enable one more 'core' operator :)

createdisk.sh
praveenkumar: @cfergeau you mean this should be part of first commit about removing MCO from override list?

praveenkumar:
This is now tested and only change we require on crc side to add the generated ssh key to the MCO side mostly patching the `99-master-ssh` machineconfig resource.

createdisk.sh
cfergeau: No no, I'm talking about ` Pull secret on the disk should be removed by MCO `, this removes the `sudo rm -f /var/lib/kubelet/config.json` through ssh, but this does not add some equivalent operation going through the MCO. I expect some kind of yaml file will need to be applied to the cluster in order to remove this file through the MCO?

praveenkumar: @cfergeau ah no. So we are replacing the pull secret https://github.com/code-ready/snc/blob/master/snc.sh#L211 here and as soon as you do it, mco automatic update `/var/lib/kubelet/config.json` file on the VM with empty content so we don't need to have any extra yaml file explict remove it.

cfergeau: Ah ok, this is the kind of things which are nice to explain in commit logs :)

PR #409, opened 2021-05-03
Add a tip in login page for getting credentials

guillaumerose:
When starting the cluster from the tray or when restarting a cluster
after a long time, the user might not remember the credentials.
This message will help him get into the cluster.

Screenshot from 2021-05-03 10-12-56

(wording, is it secure, etc.. to be discussed :) !)

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -192,6 +192,11 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 # Set default route for registry CRD from false to true.
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
+# Add a tip in the login page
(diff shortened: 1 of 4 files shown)


snc.sh
```diff
@@ -192,6 +192,11 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 # Set default route for registry CRD from false to true.
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
+# Add a tip in the login page
+retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
+patch login.html < login.html.patch
```
praveenkumar, line 197: @guillaumerose CI failure is happening  because `patch` command not found on the system. I think  you need to use `yum install -y /usr/bin/patch` in the tools script to install it before consume.

guillaumerose:
Failed in conformance tests

/retest

Issue #414, opened 2021-05-10
Remove kubeadmin-password file from bundle


praveenkumar:
As part of CRC we are already deleting the `kube-admin` user which is created by installer so the password file of that user doesn't make sense to have it in the bundle.

PR #415, opened 2021-05-10
Remove kubeadmin-passwd file from the bundle

praveenkumar:
Since new `kubeadmin` user is created using `htpasswd` provider and
password is updated for each time when `crc start` happen.

fixes: #414


cfergeau:
This is related to https://github.com/code-ready/crc/pull/2327/commits , better to get the crc PR in first, as otherwise it won't be able to load bundle generated by newer snc.

guillaumerose (approved):

PR #418, opened 2021-05-19
Externalization VM settings: memory, vCPUs and disk size.

danielmenezesbr:
Externalization VM settings through environment variables:
CRC_VM_MEMORY
CRC_VM_CPUS
CRC_VM_DISK_SIZE

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -22,6 +22,9 @@ fi
 
 INSTALL_DIR=crc-tmp-install-data
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
+SNC_CLUSTER_MEMORY=${SNC_CLUSTER_MEMORY:-14336}
+SNC_CLUSTER_CPUS=${SNC_CLUSTER_CPUS:-6}
+CRC_VM_DISK_SIZE=${CRC_VM_DISK_SIZE:-33285996544}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
@@ -130,13 +133,16 @@ ${YQ} eval-all --inplace 'select(fileIndex == 0) * select(filename == "cvo-overr
 
 # Add custom domain to cluster-ingress
 ${YQ} eval --inplace ".spec.domain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml
-# Add master memory to 12 GB and 6 cpus 
+# Set master memory and cpus
 # This is only valid for openshift 4.3 onwards
-${YQ} eval --inplace '.spec.providerSpec.value.domainMemory = 14336' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
-${YQ} eval --inplace '.spec.providerSpec.value.domainVcpu = 6' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
-# Add master disk size to 31 GiB
+echo "Master memory: $SNC_CLUSTER_MEMORY"
+${YQ} eval --inplace ".spec.providerSpec.value.domainMemory = $SNC_CLUSTER_MEMORY" ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
+echo "Master CPUS: $SNC_CLUSTER_CPUS"
+${YQ} eval --inplace ".spec.providerSpec.value.domainVcpu = $SNC_CLUSTER_CPUS" ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
+# Set master disk size
 # This is only valid for openshift 4.5 onwards
-${YQ} eval --inplace '.spec.providerSpec.value.volume.volumeSize = 33285996544' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
+echo "Master disk size: $CRC_VM_DISK_SIZE"
+${YQ} eval --inplace ".spec.providerSpec.value.volume.volumeSize = $CRC_VM_DISK_SIZE" ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
 # Add network resource to lower the mtu for CNV
 cp cluster-network-03-config.yaml ${INSTALL_DIR}/manifests/
 # Add patch to mask the chronyd service on master


snc.sh
```diff
@@ -132,11 +135,11 @@ ${YQ} eval-all --inplace 'select(fileIndex == 0) * select(filename == "cvo-overr
 ${YQ} eval --inplace ".spec.domain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml
 # Add master memory to 12 GB and 6 cpus
```
cfergeau, line 136: Can you update this comment as well?

snc.sh
```diff
@@ -132,11 +135,11 @@ ${YQ} eval-all --inplace 'select(fileIndex == 0) * select(filename == "cvo-overr
 ${YQ} eval --inplace ".spec.domain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml
 # Add master memory to 12 GB and 6 cpus 
 # This is only valid for openshift 4.3 onwards
-${YQ} eval --inplace '.spec.providerSpec.value.domainMemory = 14336' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
-${YQ} eval --inplace '.spec.providerSpec.value.domainVcpu = 6' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
+${YQ} eval --inplace ".spec.providerSpec.value.domainMemory = $CRC_VM_MEMORY" ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
+${YQ} eval --inplace ".spec.providerSpec.value.domainVcpu = $CRC_VM_CPUS" ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
 # Add master disk size to 31 GiB
```
cfergeau, line 140: And this one? Just `Set master disk size` should do

cfergeau (commented):
Kind of curious about your snc use case. People making use of this functionality need to be aware that they should not pick values smaller than the default. But the change by itself is fine with me.

snc.sh
cfergeau: Or even better, add a `echo` printing the actual amount of memory and CPU which will be used as now they are no longer hardcoded.

danielmenezesbr: I updated the comments and added echo for each variable: CRC_VM_MEMORY, CRC_VM_CPUS and CRC_VM_DISK_SIZE

cfergeau (commented):
Out of curiosity, can you provide more details as to what you are tryingto do with `snc`?

danielmenezesbr:
Hi @cfergeau!

Yes, sure! In my use case CRC has some limitations:
  * Sometimes a specific version of OpenShift is not available. For example 4.6.18 (the closest version available on the CRC is 4.6.15).
  * It is not possible to use a domain other than * .testing. For example, the use of services like xip.io could facilitate access to a remote CRC through an SSH tunnel as it is not necessary to change the host file on the client.

SNC allows overcoming these limitations.

snc.sh
```diff
@@ -130,13 +133,16 @@ ${YQ} eval-all --inplace 'select(fileIndex == 0) * select(filename == "cvo-overr
...
 ${YQ} eval --inplace ".spec.domain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" ${INSTALL_DIR}/manifests/cluster-ingress-02-config.yml
-# Add master memory to 12 GB and 6 cpus 
+# Set master memory and cpus
 # This is only valid for openshift 4.3 onwards
-${YQ} eval --inplace '.spec.providerSpec.value.domainMemory = 14336' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
-${YQ} eval --inplace '.spec.providerSpec.value.domainVcpu = 6' ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
-# Add master disk size to 31 GiB
+echo "Master memory: $CRC_VM_MEMORY"
+${YQ} eval --inplace ".spec.providerSpec.value.domainMemory = $CRC_VM_MEMORY" ${INSTALL_DIR}/openshift/99_openshift-cluster-api_master-machines-0.yaml
+echo "Master CPUS: $CRC_VM_MEMORY"
```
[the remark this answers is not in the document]
danielmenezesbr: Done! Please, check again.

cfergeau:
> SNC allows overcoming these limitations.

Ah you definitely should be using snc for this kind of things, this was the memory/disk size/... changes I was wondering about :) I guess since you are tweaking the bundle used by crc for the reasons you mentioned, you want to adjust memory/cpu/disk to your needs to get an image which fully fits your needs?

cfergeau (commented):
One thing that just occurred to me is that `CRC_VM_CPUS`/`CRC_VM_MEMORY` are a bit misleading, these values are only used by snc during the bundle generation. crc will use its own default values for CPU/memory: https://github.com/code-ready/crc/blob/fbe06cc081acf3b1e86b40dbf3fd7143430dce8d/pkg/crc/constants/constants.go#L15-L16
(I think disk size will be fine). Dunno if that's a problem for you?
Apart from this interrogation, this looks good to me :)

danielmenezesbr:
> Ah you definitely should be using snc for this kind of things, this was the memory/disk size/... changes I was wondering about :) I guess since you are tweaking the bundle used by crc for the reasons you mentioned, you want to adjust memory/cpu/disk to your needs to get an image which fully fits your needs?

Yes. In my use case, we use Red Hat OpenShift Serverless and OpenShift Service Mesh. These components demand large amounts of resources. The default values (CPUs/memory/disk size) are not sufficient and therefore need to be adjusted.

danielmenezesbr:
> One thing that just occurred to me is that `CRC_VM_CPUS`/`CRC_VM_MEMORY` are a bit misleading, these values are only used by snc during the bundle generation. crc will use its own default values for CPU/memory: https://github.com/code-ready/crc/blob/fbe06cc081acf3b1e86b40dbf3fd7143430dce8d/pkg/crc/constants/constants.go#L15-L16
> (I think disk size will be fine). Dunno if that's a problem for you?
> Apart from this interrogation, this looks good to me :)

No problem! :)
Do you suggest changing the name `CRC_VM_CPUS` /` CRC_VM_MEMORY`?

praveenkumar:
@danielmenezesbr So you are using snc to create cluster but not using to generate bundle (disk images) to use with crc and that's the reason you want to increase mem/cpu here itself?

> Do you suggest changing the name CRC_VM_CPUS / CRC_VM_MEMORY?

I would suggest `SNC_CLUSTER_CPUS/SNC_CLUSTER_MEMORY` because this is part of snc repo and here we are trying to have env with `SNC` name.

danielmenezesbr:
> So you are using snc to create cluster but not using to generate bundle (disk images) to use with crc and that's the reason you want to increase mem/cpu here itself?

@praveenkumar Yes. I use only SNC without CRC.

> would suggest SNC_CLUSTER_CPUS/SNC_CLUSTER_MEMORY because this is part of snc repo and here we are trying to have env with SNC name.

Done.

cfergeau (approved):
It's the very last steps of `snc.sh` which fail in CI, which means the whole install was successful, and the installed cluster was working well enough for most of what snc tries to do. In other words, I don't think the failure is related to this PR

PR #419, opened 2021-05-20
Force kill any existing dnf process

praveenkumar:
We are seeing lot of failure on the CI because of existing
dnf process.

```
+ sudo yum install -y podman make golang rsync
[...]
Waiting for process with pid 1281 to finish.
```

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -2,6 +2,7 @@
 
 set -exuo pipefail
 
+sudo pkill -9 dnf
 sudo yum install -y podman make golang rsync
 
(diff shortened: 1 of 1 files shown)


cfergeau:
But why is the CI environment in that state? I'd rather fix the root cause rather than papering over it, and hoping this does not break anything important.

cfergeau:
The background dnf process is `dnf-automatic`. It might be trying to update the image to 8.4 in the background, which is taking more time than usual dnf-automatic runs, and ends up getting of the way of our CI scripts.

PR #426, opened 2021-05-27
Disable the chronyd service instead masking it

praveenkumar:
As of now, we are unmasking the chronyd unit file directly using
ssh command so in case of MCO enabled it modify it back again to
masked one so first start of bundle succeed but it fails as soon
as we do stop => start since the crc codebase not able to start
a masked unit.

```
$ systemctl status chronyd.service
● chronyd.service
   Loaded: masked (Reason: Unit chronyd.service is masked.)
   Active: inactive (dead)
$ sudo timedatectl set-ntp on
Failed to set ntp: Failed to start/stop NTP unit
```

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -220,9 +220,6 @@ retry ${OC} delete secrets kubeadmin -n kube-system
 # Add security message on the web console
 retry ${OC} create -f security-notice.yaml
 
-# Replace pull secret with a null json string '{}'
-retry ${OC} replace -f pull-secret.yaml
-
 # Remove the Cluster ID with a empty string.
 retry ${OC} patch clusterversion version -p '{"spec":{"clusterID":""}}' --type merge
 
@@ -251,6 +248,12 @@ retry ${OC} patch image.config.openshift.io cluster -p '{"spec": {"additionalTru
 retry ${OC} delete mc chronyd-mask
 
 # Wait for the cluster again to become stable because of all the patches/changes
+wait_till_cluster_stable
+
+# Replace pull secret with a null json string '{}'
+retry ${OC} replace -f pull-secret.yaml
+
+# Wait for the cluster again to become stable because of remove of pull secret
 wait_till_cluster_stable
 
 # Delete the pods which are there in Complete state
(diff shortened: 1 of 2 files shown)


praveenkumar:
/hold

Need to check the bundle manually because this was not possible older version of ignition so trying again if it is working with new one.

cfergeau:
The disabled chronyd service is useful for 2 things:
- in snc, for `renew_certificates` (we create the cluster with a date set 1 day in the past with NTP disabled, then we jump back to the current time, which will trigger regeneration of the certificates)
- in crc so that the e2e tests can test certificate expiry/renewal by setting `CRC_DEBUG_ENABLE_STOP_NTP`

and I think that's not used for more than that. I'd favour just reenabling NTP in the VM after `renew_certificates` is done, and stop it again for e2e tests when crc starts with `CRC_DEBUG_ENABLE_STOP_NTP` set. I don't think the e2e tests need to reboot after disabling NTP, so hopefully they would not have to fight with the MCO?

praveenkumar:
> and I think that's not used for more than that. I'd favour just reenabling NTP in the VM after renew_certificates is done, and stop it again for e2e tests when crc starts with CRC_DEBUG_ENABLE_STOP_NTP set. I don't think the e2e tests need to reboot after disabling NTP, so hopefully they would not have to fight with the MCO?

@cfergeau yes, this is the  plan so we can have service enabled/active before create the disk image and in e2e test side disable manually as soon as VM starts.

snc.sh
```diff
@@ -253,6 +250,12 @@ retry ${OC} delete mc chronyd-mask
 # Wait for the cluster again to become stable because of all the patches/changes
 wait_till_cluster_stable
 
+# Replace pull secret with a null json string '{}'
+retry ${OC} replace -f pull-secret.yaml
```
cfergeau, line 254: Will this help with crc downloading a lot of  data right after initial start? Or is this unrelated?

cfergeau (approved):

snc.sh
praveenkumar: @cfergeau it is unrelated, downloading the data still happen if any image on marketplace is updated.

PR #432, opened 2021-06-02
Don't disable the chronyd service

praveenkumar:
This patch will make sure the generated bundle have same kind of
property for 4.8 (0beeb25) and 4.7 which is running chronyd service
as default.


cfergeau (approved):
Sure, though I think this will break the cert renewal e2e test until it's adjusted?

praveenkumar:
> Sure, though I think this will break the cert renewal e2e test until it's adjusted?

Yes this should be adjusted in the crc side.

PR #433, opened 2021-06-03
Don't delete MCO resource for 4.8

praveenkumar:
As part of ae6029b we removed the MCO from override list but missed
this one..


cfergeau (approved):
Ack, but also wondering if we have a way of blocking upgrade attempts?

praveenkumar:
> Ack, but also wondering if we have a way of blocking upgrade attempts?

@cfergeau don't know yet but  will figure this out.

PR #434, opened 2021-06-08
Use dnf download plugin to download packages

praveenkumar:


snc-library.sh
```diff
@@ -188,7 +188,8 @@ function create_pvs() {
 # This follows https://blog.openshift.com/enabling-openshift-4-clusters-to-stop-and-resume-cluster-vms/
 # in order to trigger regeneration of the initial 24h certs the installer created on the cluster
 function renew_certificates() {
-    local vm_prefix=$(get_vm_prefix ${CRC_VM_NAME})
+    local vm_prefix
+    vm_prefix=$(get_vm_prefix ${CRC_VM_NAME})
```
cfergeau, line 192: Oh nice, I thought the fix would be much more complicated :)

createdisk-library.sh
```diff
@@ -165,7 +165,7 @@ function prepare_hyperV() {
     else
         # Download the hyperV daemons and libvarlink-util dependency on host
         mkdir $1/packages
-        sudo yum install -y --downloadonly --downloaddir $1/packages hyperv-daemons
+        sudo yum download --downloadonly --downloaddir $1/packages hyperv-daemons --resolve
```
cfergeau, line 168: This is part of `dnf-utils`, dunno if this gets installed by default? For what it's worth, `--resolve` will only pick up dependencies if they are not already installed on the system, so this has the same issue as what you are trying to solve, but only for dependencies (ie `hyperv-daemons` will always be downloaded).

praveenkumar: > This is part of dnf-utils, dunno if this gets installed by default?

It is by default installed.

>  For what it's worth, --resolve will only pick up dependencies if they are not already installed on the system, so this has the same issue as what you are trying to solve, but only for dependencies (ie hyperv-daemons will always be downloaded).

@cfergeau Ah yes looks like I need to use `alldep` also.

```
--alldeps
              When used with --resolve, download all dependencies (do not skip already installed ones).
```

praveenkumar: Looks like `alldep` is not working as I thought it suppose to work :( https://unix.stackexchange.com/questions/50642/download-all-dependencies-with-yumdownloader-even-if-already-installed does contain some of the info perform this but those all looks cryptic so living it as it is with `--resolve`

PR #437, opened 2021-06-14
Use create instead run for podman 

praveenkumar:


createdisk.sh
```diff
@@ -59,7 +59,7 @@ fi
 # Add gvisor-tap-vsock service
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
   podman create --name=gvisor-tap-vsock --privileged --net=host -it quay.io/crcont/gvisor-tap-vsock:v4
-  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  podman generate systemd --new --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
```
cfergeau, line 62: Is this PR fixing a bug? The systemd unit file will be tied to the crc VM, in other words, to a single machine. And before the first commit in this PR, the container/pod was already running before `podman generate` was run. `--new` documentation also says: 
> The unit files are created best effort and may need to be further edited; please review the generated files carefully before using them in production

Some context is missing to correctly assess/review this PR :)

praveenkumar:
/hold

need testing.

createdisk.sh
praveenkumar: > Some context is missing to correctly assess/review this PR :)

@cfergeau without `new` it is bound to crc VM and also to the container ID which created during first run, since podman 3.0.2 (which is right now as part of 4.8) doesn't provide `RequiresMountsFor` so meanwhile `--new` option is only best bet to make sure unit file doesn't tied with container ID.

cfergeau: I don't understand how it is an issue that it's tied to the container ID? This ID does not seem to change on reboots here? I'm not against this `--new` change, but I'd like to understand what concrete benefit we are going to get out of it, or if it's just cosmetic.
Do I read "since podman 3.0.2 doesn't provide RequiresMountsFor" correctly, and even after this PR `RequiresMountsFor` will still be missing from the generated unit file?

praveenkumar: > Do I read "since podman 3.0.2 doesn't provide RequiresMountsFor" correctly, and even after this PR `RequiresMountsFor` will still be missing from the generated unit file?

@cfergeau yes, also `--new` change not making any good during testing.

cfergeau:
I don't think this PR had been approved.

PR #438, opened 2021-06-15
Revert MCO changes

praveenkumar:
- https://bugzilla.redhat.com/show_bug.cgi?id=1965992


guillaumerose:
Maybe we can keep a branch with the MCO ? Like `master-mco` ?

praveenkumar:
> Maybe we can keep a branch with the MCO ? Like `master-mco` ?

@guillaumerose yes that was the plan, have branch locally `mco_4.8` but didn't push it.

cfergeau:
From this bugzilla link, I only see "slow shutdown". I guess there are bigger issues caused by this (failing snc CI? failing crc CI? something else?). Can this be mentioned explicitly in the commit log so that we know what to test for when the bug is fixed and we want to revert this?

praveenkumar:
> From this bugzilla link, I only see "slow shutdown". I guess there are bigger issues caused by this (failing snc CI? failing crc CI? something else?). Can this be mentioned explicitly in the commit log so that we know what to test for when the bug is fixed and we want to revert this?

@cfergeau It is slow shutdown which happen because of MCO so if we forcefully poweroff the VM then the data about container storage gets corrupted and that was the reason we have gvisor-tap container failing to start. I think bugzilla link have enough context and once that is fixed we can create the bundle again to check if it is fixed.

cfergeau:
> I think bugzilla link have enough context 

It has far too much context, which is why a few lines summary is useful to know exactly what is important there. And the gvisor-tap-vsock failure is not mentioned in bugzilla. If the gvisor-tap-vsock failure is addressed differently, are there more issues which would force us to revert right now? I understand there is disk corruption, but as long as we don't ship this to our users, this might be acceptable?

cfergeau:
https://bugzilla.redhat.com/show_bug.cgi?id=1965992#c10 has a workaround so hopefully we can do without that revert.

praveenkumar:
> https://bugzilla.redhat.com/show_bug.cgi?id=1965992#c10 has a workaround so hopefully we can do without that revert.

Yes #439 should have the workaround but need to testing, will keep this open meanwhile and close it once the workaround actually works for us.

cfergeau:
I think this can be closed as this will be handled in `crc` instead? Or do we also need this in `snc` when shutting down our cluster during install?

praveenkumar:
Closing this since it handled by crc now.

PR #439, opened 2021-06-16
Remove MCO daemon and server pods forcefully to avoid update lock

praveenkumar:
"ignore SIGTERM" handler doesn't get removed in the case of rebootless
updates (so that "fortunately rebootless" update may not have been so fortunate).
An MCO-triggered reboot takes a different code path and removes the lock
itself then manually reboot. As a workaround we need to forcefully
delete the MCO pod.

- https://bugzilla.redhat.com/show_bug.cgi?id=1965992

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -240,6 +240,13 @@ retry ${OC} delete mc chronyd-mask
 # Wait for the cluster again to become stable because of all the patches/changes
 wait_till_cluster_stable
 
+# Delete the pods from MCO forcefully as workaround of
+# https://bugzilla.redhat.com/show_bug.cgi?id=1965992
+retry ${OC} delete pods --all -n openshift-machine-config-operator --grace-period=0
+
+# Wait for the cluster to become stable again after removing the MCO pods
+wait_till_cluster_stable
+
 # Replace pull secret with a null json string '{}'
 retry ${OC} replace -f pull-secret.yaml
 


praveenkumar:
/hold

need to test the generated bundle.

snc.sh
```diff
@@ -246,6 +246,13 @@ retry ${OC} replace -f pull-secret.yaml
 # Wait for the cluster again to become stable because of remove of pull secret
 wait_till_cluster_stable
 
+# Delete the pods from MCO forcefully as workaround of
+# https://bugzilla.redhat.com/show_bug.cgi?id=1965992
+retry ${OC} delete pods --all -n openshift-machine-config-operator --grace-period=0
+
+# Wait for the cluster again to become stable because removing pods from MCO
```
cfergeau, line 253: `Wait for the cluster to become stable again after removing the MCO pods` reads better I think

snc.sh
```diff
@@ -246,6 +246,13 @@ retry ${OC} replace -f pull-secret.yaml
 # Wait for the cluster again to become stable because of remove of pull secret
 wait_till_cluster_stable
 
+# Delete the pods from MCO forcefully as workaround of
+# https://bugzilla.redhat.com/show_bug.cgi?id=1965992
+retry ${OC} delete pods --all -n openshift-machine-config-operator --grace-period=0
```
cfergeau, line 251: A `retry` should not be needed as we just waited for the cluster to be stable?

cfergeau (approved):

snc.sh
praveenkumar: @cfergeau yes it is just for a safer side, even we waited for cluster to be stable it doesn't provide any guarantee that it will no more go to reconcile  state.

praveenkumar:
During testing looks like even we kill it here it doesn't work so we have to kill those pods as part of crc codebase before stop command.


/close

PR #443, opened 2021-06-22
Update MIRROR uri for 4.8 branch and use latest 4.8 candidate version

praveenkumar:


guillaumerose (approved):

snc.sh
```diff
@@ -41,7 +41,7 @@ if test -n "${OPENSHIFT_VERSION-}"; then
     OPENSHIFT_RELEASE_VERSION=${OPENSHIFT_VERSION}
     echo "Using release ${OPENSHIFT_RELEASE_VERSION} from OPENSHIFT_VERSION"
 else
-    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/latest-4.8/release.txt | sed -n 's/^ *Version: *//p')"
+    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/candidate-4.8/release.txt | sed -n 's/^ *Version: *//p')"
```
cfergeau, line 44: Why switch from `latest` to `candidate`? `latest` is more in line with what we'll be releasing with?

cfergeau (approved):

snc.sh
praveenkumar: yes but for CI we always test what is going to be the latest so it's good to have `latest` here.

PR #446, opened 2021-07-01
Remove .swp file and add it to gitignore

praveenkumar:
It's added by mistake as part of https://github.com/code-ready/snc/pull/442


.gitignore
```diff
@@ -8,3 +8,4 @@ oc
 yq
 openshift-clients/
 podman-remote/
+.swp
```
cfergeau, line 11: Might be worth matching `.sw?` https://unix.stackexchange.com/questions/326707/vim-what-are-all-the-possible-swapfile-extensions

cfergeau (approved):

PR #447, opened 2021-07-01
Use ocp location instead ocp-dev-preview for mirror

praveenkumar:
Since 4.8 have rc bits available it is good time to switch to
latest 4.8 candidate directory then using ocp-dev-preview location.


cfergeau:
This is the same as #443 isn't it?

PR #448, opened 2021-07-02
Remove the system:openshift:openshift-authenticator csr

praveenkumar:
Looks like after the cert rotation
`system:openshift:openshift-authenticator` csr is not recreated and
causing issue for auth operator. As a workaround we need to manually
delete it so new csr is generated with updated cert info.

- https://bugzilla.redhat.com/show_bug.cgi?id=1978193

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -169,6 +169,10 @@ fi
 # Wait for install to complete, this provide another 30 mins to make resources (apis) stable
 ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} wait-for install-complete ${OPENSHIFT_INSTALL_EXTRA_ARGS}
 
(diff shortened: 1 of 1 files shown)


praveenkumar:
/hold

Need testing

snc.sh
```diff
@@ -169,6 +169,10 @@ fi
 # Wait for install to complete, this provide another 30 mins to make resources (apis) stable
 ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} wait-for install-complete ${OPENSHIFT_INSTALL_EXTRA_ARGS}
 
+# Remove the openshift-authenticator
```
cfergeau, line 172: `Remove the openshift-authenticator CSR to force a certificate regeneration`?

cfergeau (approved):

PR #454, opened 2021-07-14
support for arm64

Prashanth684:
with the developer preview of ARM available for 4.9, adding support in it for SNC so we can work toward having crc support for macs

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -76,17 +76,20 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo chown core.core ~/.ssh/au
 shutdown_vm ${VM_PREFIX}
 start_vm ${VM_PREFIX}
 
-# Get the rhcos ostree Hash ID
-ostree_hash=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "cat /proc/cmdline | grep -oP \"(?<=${BASE_OS}-).*(?=/vmlinuz)\"")
+# Only used for hyperkit bundle generation
+if [ -n "${SNC_GENERATE_MACOS_BUNDLE}" ]; then
+    # Get the rhcos ostree Hash ID
+    ostree_hash=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "cat /proc/cmdline | grep -oP \"(?<=${BASE_OS}-).*(?=/vmlinuz)\"")
 
-# Get the rhcos kernel release
-kernel_release=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'uname -r')
+    # Get the rhcos kernel release
+    kernel_release=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'uname -r')
 
-# Get the kernel command line arguments
-kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline')
+    # Get the kernel command line arguments
+    kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline')
 
-# SCP the vmlinuz/initramfs from VM to Host in provided folder.
-${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${BASE_OS}-${ostree_hash}/* $1
+    # SCP the vmlinuz/initramfs from VM to Host in provided folder.
+    ${SCP} -r core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${BASE_OS}-${ostree_hash}/* $1
+fi
 
 # Add internalIP as node IP for kubelet systemd unit file
 # More details at https://bugzilla.redhat.com/show_bug.cgi?id=1872632
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -58,7 +58,7 @@ function run_preflight_checks() {
 
 	#Just warn if architecture is not supported
 	case $ARCH in
-		x86_64|ppc64le|s390x)
+		x86_64|ppc64le|s390x|aarch64)
 			echo "The host arch is ${ARCH}.";;
 		*)
  			echo "The host arch is ${ARCH}. This is not supported by SNC!";;
--- a/snc.sh
+++ b/snc.sh
@@ -66,7 +66,7 @@ echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RE
 # Extract openshift-install binary if not present in current directory
 if test -z ${OPENSHIFT_INSTALL-}; then
     echo "Extracting OpenShift baremetal installer binary"
-    ${OC} adm release -a ${OPENSHIFT_PULL_SECRET_PATH} extract ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --command=openshift-baremetal-install --to .
+    ${OC} adm release extract -a ${OPENSHIFT_PULL_SECRET_PATH} ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --command=openshift-baremetal-install --to .
     OPENSHIFT_INSTALL=./openshift-baremetal-install
 fi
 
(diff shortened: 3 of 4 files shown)


gbraad:
We do not have the equipment to confirm this works. Can you provide tests, or a means to ensure consistency/stability for follow-up changes we make to the general SNC codebase.

Prashanth684:
> We do not have the equipment to confirm this works. Can you provide tests, or a means to ensure consistency/stability for follow-up changes we make to the general SNC codebase.

so "we" as in the multi-arch team at RedHat use ARM systems that we loan from beaker. That being said long term, the plan is to have CI on ARM starting mainly with AWS but we could also look to extend it to have a CI setup for crc which would be necessary. Let me bring this up for further discussion in our forums.

snc-library.sh
```diff
@@ -281,7 +281,7 @@ function wait_till_cluster_stable() {
     done
 
     # Wait till all the pods are either running or complete state
-    while oc get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed; do
+    while ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed; do
```
cfergeau, line 284: Feel free to split this change in a separate PR, this could be merged independently from the arm support.

tools.sh
```diff
@@ -24,6 +24,11 @@ case "${ARCH}" in
         SNC_GENERATE_MACOS_BUNDLE=1
 	SNC_GENERATE_WINDOWS_BUNDLE=1
 	;;
+    aarch64)
+        yq_ARCH="arm64"
+        SNC_GENERATE_MACOS_BUNDLE=1
+	SNC_GENERATE_WINDOWS_BUNDLE=
```
cfergeau, line 30: Is macOS bundle generation intentionally enabled because you are targetting the M1? 
Feel free to remove the mix of spaces and tabs in this area while you are changing this (should be only spaces I think)

cfergeau (commented):
If these are the only changes needed to have snc running successfully on aarch64 hardware, then I'm happy to see it's so straightforward :)
Only remaining question is how we do test this.

Fwiw, if no additional changes are needed, I'd be fine with merging this even if aarch64 is unsupported from our end.

snc-library.sh
Prashanth684: removed.

tools.sh
Prashanth684: correct. the MAC bundles are targeting M1.

snc-library.sh
Prashanth684: opened https://github.com/code-ready/snc/pull/458

praveenkumar:
@Prashanth684 Have you tested this PR, like generated bundle for arm and able to run those even on beaker infra?

Prashanth684:
> @Prashanth684 Have you tested this PR, like generated bundle for arm and able to run those even on beaker infra?

good point @praveenkumar . the single node cluster itself came up fine, but the createdisk.sh command to create the images pulls amd64 images from quay which will not work. can you point us to how to build these images so we can build them manifest listed for amd64 and arm64? these are the images in question:

```
quay.io/crcont/dnsmasq:latest
quay.io/crcont/gvisor-tap-vsock:v4
quay.io/crcont/routes-controller:latest
```

cfergeau:
The dnsmasq image is there https://github.com/code-ready/crc/tree/master/images/dnsmasq
The gvisor-tap-vsock one is https://github.com/containers/gvisor-tap-vsock/tree/main/images
The routes-controller one is https://github.com/code-ready/routes-controller/tree/master/images

Some of these images can be built on osbs, I'd have to check if it can be used to generate aarch64 images.

praveenkumar:
@Prashanth684 The images are built on osbs (Thanks to @cfergeau) and uploaded to quay so it should pull platform specific.
- https://quay.io/repository/crcont/dnsmasq?tab=tags 
- https://quay.io/repository/crcont/routes-controller?tab=tags 
- https://quay.io/repository/crcont/gvisor-tap-vsock

cfergeau:
> The dnsmasq image is there https://github.com/code-ready/crc/tree/master/images/dnsmasq
> The gvisor-tap-vsock one is https://github.com/containers/gvisor-tap-vsock/tree/main/images
> The routes-controller one is https://github.com/code-ready/routes-controller/tree/master/images
> 
> Some of these images can be built on osbs, I'd have to check if it can be used to generate aarch64 images.

I've updated these 3 images, they should be multiarch now (aarch64/ppc64le/s390x/x86_64)

Prashanth684:
many thanks again @cfergeau ! i will use these and run tests again and let you know.

Prashanth684:
ok. updated and rebased. i was able to generate a crc bundle. @praveenkumar @cfergeau any docs on how i can run crc with a custom bundle?

cfergeau:
> any docs on how i can run crc with a custom bundle?

@Prashanth684 if you build crc from git, `crc setup; crc start -b $custom_bundle` should be all you need. Couple of caveats:
- even if crc builds on aarch64, I don't know how well the preflights will work there
- at this point we support OpenShift 4.8, not sure how well OpenShift 4.9 or newer would run with crc from git

createdisk.sh
```diff
@@ -86,7 +86,7 @@ kernel_release=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'uname -r')
 kernel_cmd_line=$(${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'cat /proc/cmdline')
 
 # SCP the vmlinuz/initramfs from VM to Host in provided folder.
-${SCP} core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${BASE_OS}-${ostree_hash}/* $1
+${SCP} -r core@api.${CRC_VM_NAME}.${BASE_DOMAIN}:/boot/ostree/${BASE_OS}-${ostree_hash}/* $1
```
cfergeau, line 89: The commands from `ostree_hash=` to here could actually be wrapped in a `if [ -n "${SNC_GENERATE_MACOS_BUNDLE}" ]` as it's not going to be useful on linux/Windows.
I don't think hyperkit (which crc uses on macOS) works on M1 macs, so it might be a bit early to generate macOS  bundles by default in snc.
Could you at least expand the commit log a bit, is this causing an error if we omit the `-r`? Which additional files do we need from `/boot/ostree` which are in subdirs?

cfergeau (approved):
Looks good to me, a few questions/suggestions, but nothing blocking.

createdisk.sh
Prashanth684: so for aarch64 the kernel pulls in DTB - https://src.fedoraproject.org/rpms/kernel/blob/f34/f/kernel.spec#_1533 as part of the bootloading process which is the extra directory that requires the -r. 

Yes, so looks like hyperkit is not yet available for arm64 so I would just want to get these changes in for now so that in the future when it is available we can start on the CRC side of things.

Prashanth684:
> > any docs on how i can run crc with a custom bundle?
> 
> @Prashanth684 if you build crc from git, `crc setup; crc start -b $custom_bundle` should be all you need. Couple of caveats:
> 
> * even if crc builds on aarch64, I don't know how well the preflights will work there
> * at this point we support OpenShift 4.8, not sure how well OpenShift 4.9 or newer would run with crc from git

So i did try and the pre-flight check failed with finding the virtualization device which doesn't seem to be present on any hardware we have today. So, i think for now I would just like to get this in so we at least have support to create bundles so we can investigate more when we get an ARM system with virtualization hardware or hopefully an M1 which we can test with. But again as @cfergeau pointed out hyperkit is not supported for arm64 so we will also have to wait for that.

snc.sh
```diff
@@ -66,7 +66,7 @@ echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RE
 # Extract openshift-install binary if not present in current directory
 if test -z ${OPENSHIFT_INSTALL-}; then
     echo "Extracting OpenShift baremetal installer binary"
-    ${OC} adm release -a ${OPENSHIFT_PULL_SECRET_PATH} extract ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --command=openshift-baremetal-install --to .
+    ${OC} adm release extract -a ${OPENSHIFT_PULL_SECRET_PATH} ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --command=openshift-baremetal-install --to .
```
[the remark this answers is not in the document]
Prashanth684: yes looks like it....this used to work for me before too

createdisk.sh
Prashanth684: changed to have this block under the conditional.

cfergeau (approved):

PR #455, opened 2021-07-27
Remove replica scale from 2 to 1 for deployment 

praveenkumar:
From 4.8 we don't need to scale the deployment because it already
support single node configuration and by default deployment replicas
set to 1.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -222,8 +222,6 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo mkdir /opt/release-manife
 retry ${OC} -n openshift-cluster-version get pods -o=name
 CVO_POD_NAME=$(${OC} -n openshift-cluster-version get pods -o=name)
 retry ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo KUBECONFIG=/opt/kubeconfig oc rsync -n openshift-cluster-version ${CVO_POD_NAME}:/release-manifests/ /opt/release-manifests/"
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "s/replicas: 2/replicas: 1/" /opt/release-manifests/*deployment.yaml'
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "s/replicas: 2/replicas: 1/" /opt/release-manifests/*clusterserviceversion.yaml'
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "/memory: /d" /opt/release-manifests/*deployment.yaml'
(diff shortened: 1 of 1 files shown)


ci.sh
```diff
@@ -2,8 +2,6 @@
 
 set -exuo pipefail
 
-sudo yum install -y podman make golang rsync
-
```
cfergeau, line 6: This is tracked in https://github.com/ironcladlou/openshift4-libvirt-gcp/pull/45 which isn't merged yet, probably better to keep this commit out of this PR?

praveenkumar: @cfergeau this is not merged but the image is already updated to GCP with those changes, I don't see any strong opinion to not merge that PR.

snc.sh
```diff
@@ -222,8 +222,6 @@ ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo mkdir /opt/release-manife
 retry ${OC} -n openshift-cluster-version get pods -o=name
 CVO_POD_NAME=$(${OC} -n openshift-cluster-version get pods -o=name)
 retry ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "sudo KUBECONFIG=/opt/kubeconfig oc rsync -n openshift-cluster-version ${CVO_POD_NAME}:/release-manifests/ /opt/release-manifests/"
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "s/replicas: 2/replicas: 1/" /opt/release-manifests/*deployment.yaml'
-${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo sed -i "s/replicas: 2/replicas: 1/" /opt/release-manifests/*clusterserviceversion.yaml'
```
guillaumerose, line 226: I still see this in the 4.8.2 release image. :thinking: 

```
[root@6cbaf2e1b239 release-manifests]# grep -R "replicas: 2" .
./0000_50_olm_15-packageserver.clusterserviceversion.yaml:            replicas: 2
```

praveenkumar: @guillaumerose can you check 4.9 night side because there it is failing https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/5093/pull-ci-openshift-installer-master-e2e-crc/1419751100604485632

```
+ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc core@api.crc.testing 'sudo sed -i "s/replicas: 2/replicas: 1/" /opt/release-manifests/*clusterserviceversion.yaml'
sed: can't read /opt/release-manifests/*clusterserviceversion.yaml: No such file or directory
```

guillaumerose: OK I understand. In 4.9, they introduced a package server manager which takes care of this.

https://github.com/openshift/operator-framework-olm/blob/1dc76f08ed05a635458420ffa979aebbe59a3890/pkg/package-server-manager/config.go#L14

So looks good for 4.9 :+1:

praveenkumar:
merging only single commit.

PR #399, opened 2021-04-23
Run OpenShift conformance tests

cfergeau:
Rather than kubernetes/conformance, this attempts to run
openshift/conformance. This disables all the instrumentation, network,
node, scheduling and storage tests as I did not have time to look in
details at the several dozens of failures they had. The goal is to
eventually empty the ignore list and fix/ignore the ignored tests.

This is just a test PR to trigger the run of the conformance test in the CI.

PR #456, opened 2021-07-27
Run OpenShift reliability/minimal test suite

praveenkumar:
Rather than kubernetes/conformance, this attempts to run
openshift reliability/minimal tests. This also adds some tests to
ignore list which was failing during testing due to apiserver
unavailblity or single node constrain. Plan is to reduce this ignore
list and also try to add openshift conformance test in future.

This is not supersede #399 but a different subset of tests which is part of openshift instead plain k8s.


cfergeau:
Compared to `kubernetes/conformance`, this removes 103 tests, and adds 718, so it's a very different suite.
However, `openshift/conformance` only removes one test when compared to `experimental/reliability/minimal` (and adds 2065, but that makes sense).
Is this why you'd favour `experimental/reliability/minimal` over `kubernetes/conformance`? The `experimental` naming is making me a bit reluctant with this.

ci.sh
```diff
@@ -47,7 +47,7 @@ sudo mv out/linux-amd64/crc /usr/local/bin/
 popd
```
cfergeau, line 72: I'd expand the commit log a bit, `Giving more resources to the cluster we run the testsuite on should hopefully prevent some disruptions/timeouts/... because of low memory conditions`

snc.sh
```diff
@@ -169,6 +169,10 @@ fi
 # Wait for install to complete, this provide another 30 mins to make resources (apis) stable
 ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} wait-for install-complete ${OPENSHIFT_INSTALL_EXTRA_ARGS}
 
+# Remove the openshift-authenticator CSR to force a certificate regeneration
+# BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1978193
+retry ${OC} delete csr system:openshift:openshift-authenticator
+
```
cfergeau, line 175: we made a 4.8 release without this, I'm not sure what is the impact of this? Surely this deserves to be in a PR of its own?

cfergeau (commented):
A bunch of questions, but overall looks good!

praveenkumar:
> Compared to `kubernetes/conformance`, this removes 103 tests, and adds 718, so it's a very different suite.
> However, `openshift/conformance` only removes one test when compared to `experimental/reliability/minimal` (and adds 2065, but that makes sense).
> Is this why you'd favour `experimental/reliability/minimal` over `kubernetes/conformance`? The `experimental` naming is making me a bit reluctant with this.

@cfergeau I am considering 2 things when going with this suite.
 1. Coverage for openshift specific test cases increase from 0 to something.
 2. Duration of a successful run is around 1 hour 30 mins instead 2-3 hours in case of complete conformance test suite.

About naming I need to confirm with openshift team but as per description it says `Set of highly reliable tests.` which makes me bit more confident of using it.

About removed 103 tests from `kubernetes/conformance` is a bit concerning :(

snc.sh
praveenkumar: @cfergeau we made release of 4.8 using this https://github.com/code-ready/snc/commit/c1009676b800401cb572062715459da944ddb3d4 but for 4.9 bug is recently resolved so we don't need this change anymore.

cfergeau:
> About naming I need to confirm with openshift team but as per description it says Set of highly reliable tests. which makes me bit more confident of using it.

It's relatively recent, maybe that explains the `experimental` in the naming. A bit of digging gives a more detailed description/intent:
https://github.com/openshift/origin/pull/25946
> Adds a list of tests that pass with extremely high reliability. We will run these tests in a new e2e job. The origina list was generated by taking tests that have passed at a rate of 99% or higher over the last week, from sippy data.

Disappointing that we still need to disable ~20 tests from this list :(

> About removed 103 tests from kubernetes/conformance is a bit concerning :(

Most of these are readded when moving from `reliability/minimal` to `openshift/conformance` so this seems intentional, these are just the 'less reliable' ones.

> Duration of a successful run is around 1 hour 30 mins instead 2-3 hours in case of complete conformance test suite.

But only the complete conformance test suite gives us some "this behaves as an OpenShift cluster" label, so we'll have to take these 2-3h at some point. Running a subset of these tests is better than nothing, but is not our end goal :)

cfergeau:
`Separate retry with pipe operations` and `Add retry for pv creation commands` are already tracked in https://github.com/code-ready/snc/pull/457 , can they be split out from this PR?

praveenkumar:
> Disappointing that we still need to disable ~20 tests from this list :(

- Some of them from storage side which we are not supporting (volume snapshot stuff)
- Matrix specific test cases because we don't enable monitoring
- oc must gather one which also fails because of volume snapshots issues (This I really need to understand and create a BZ)
- Deamon set specific one which really 2 nodes to do testing (I am not sure why it doesn't skip)

> But only the complete conformance test suite gives us some "this behaves as an OpenShift cluster" label, so we'll have to take these 2-3h at some point. Running a subset of these tests is better than nothing, but is not our end goal :)

Yes this is the end goal, I am hoping we can piggy back on SNO team in near future to try to run complete suite because if they need a bump on CI time then we can directly benefited from it.

praveenkumar:
@cfergeau I run tests on this PR 3 times and no single failure https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-code-ready-snc-master-e2e-snc so I am bit confident to have it.

cfergeau (approved):
We could go with this for now.

PR #457, opened 2021-07-28
Add retry for pv creation commands

praveenkumar:
This patch will fix following error from CI
```
./openshift-clients/linux/oc create -f -
error: error when creating "STDIN": Post "https://api.crc.testing:6443/api/v1/persistentvolumes?fieldManager=kubectl-create": dial tcp 192.168.126.11:6443: connect: connection refused
```

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -170,19 +170,21 @@ function create_pvs() {
 
     for pvname in $(seq -f "pv%04g" 1 ${count}); do
         if ! ${OC} get pv "${pvname}" &> /dev/null; then
-            generate_pv "${pvdir}/${pvname}" "${pvname}" | ${OC} create -f -
+            generate_pv "${pvdir}/${pvname}" "${pvname}" > tmp_pv.yaml
+            retry ${OC} create -f tmp_pv.yaml
+	    rm -fr tmp_pv.yaml
         else
             echo "persistentvolume ${pvname} already exists"
         fi
     done
 
     # Apply registry pvc to bound with pv0001
-    ${OC} apply -f registry_pvc.yaml
+    retry ${OC} apply -f registry_pvc.yaml
 
     # Add registry storage to pvc
-    ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim": "crc-image-registry-storage"}}]' --type=json
+    retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim": "crc-image-registry-storage"}}]' --type=json
     # Remove emptyDir as storage for registry
-    ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "remove", "path": "/spec/storage/emptyDir"}]' --type=json
+    retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "remove", "path": "/spec/storage/emptyDir"}]' --type=json
 }
 
 # This follows https://blog.openshift.com/enabling-openshift-4-clusters-to-stop-and-resume-cluster-vms/
--- a/snc.sh
+++ b/snc.sh
@@ -191,7 +191,8 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
 # Add a tip in the login page
-retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
+secret_template=$(retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json)
+${JQ} -r '.data["login.html"]' <(echo "${secret_template}") | base64 -d > login.html
 ${PATCH} login.html < login.html.patch
 retry ${OC} create secret generic login-template --from-file=login.html -n openshift-config
 
--- a/ci.sh
+++ b/ci.sh
@@ -47,7 +47,7 @@ sudo mv out/linux-amd64/crc /usr/local/bin/
 popd
 
 crc setup
-crc start -p "${HOME}"/pull-secret -b crc_libvirt_*.crcbundle
+crc start --disk-size 80 -m 24000 -c 10 -p "${HOME}"/pull-secret -b crc_libvirt_*.crcbundle
 
 mkdir -p /tmp/artifacts
 export KUBECONFIG="${HOME}"/.crc/machines/crc/kubeconfig
--- a/ci.sh
+++ b/ci.sh
@@ -11,6 +11,11 @@ cat > /tmp/ignoretests.txt << EOF
 [sig-scheduling] SchedulerPreemption [Serial] validates lower priority pod preemption by critical pod [Conformance] [Suite:openshift/conformance/serial/minimal] [Suite:k8s]
 [k8s.io] [sig-node] NoExecuteTaintManager Multiple Pods [Serial] evicts pods with minTolerationSeconds [Disruptive] [Conformance] [Suite:k8s]
 [k8s.io] [sig-node] NoExecuteTaintManager Single Pod [Serial] removing taint cancels eviction [Disruptive] [Conformance] [Suite:k8s]
+[sig-api-machinery] AdmissionWebhook [Privileged:ClusterAdmin] should mutate custom resource with pruning [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
+[sig-api-machinery] AdmissionWebhook [Privileged:ClusterAdmin] should mutate pod and apply defaults after mutation [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
+[sig-api-machinery] Aggregator Should be able to support the 1.17 Sample API Server using the current Aggregator [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
+[sig-apps] Daemon set [Serial] should rollback without unnecessary restarts [Conformance] [Skipped:SingleReplicaTopology] [Suite:openshift/conformance/serial/minimal] [Suite:k8s]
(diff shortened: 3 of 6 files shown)


cfergeau:
> This patch will fix following error from CI

Is this happening with 4.8 or 4.9?

praveenkumar:
> Is this happening with 4.8 or 4.9?

@cfergeau This is from 4.9 side.

cfergeau:
Not a big fan that some temporary files need to be added, the script is nicer before these changes :(

snc-library.sh
```diff
@@ -170,19 +170,20 @@ function create_pvs() {
 
     for pvname in $(seq -f "pv%04g" 1 ${count}); do
         if ! ${OC} get pv "${pvname}" &> /dev/null; then
-            generate_pv "${pvdir}/${pvname}" "${pvname}" | ${OC} create -f -
+            generate_pv "${pvdir}/${pvname}" "${pvname}" > tmp_pv.yaml 
+            retry ${OC} create -f tmp_pv.yaml
```
cfergeau, line 174:
```
retry  ${OC} create -f - <<EOF
    $(generate_pv "${pvdir}/${pvname}" "${pvname}")
EOF
```

snc-library.sh
```diff
@@ -170,19 +170,20 @@ function create_pvs() {
...
         else
             echo "persistentvolume ${pvname} already exists"
         fi
     done
 
     # Apply registry pvc to bound with pv0001
-    ${OC} apply -f registry_pvc.yaml
+    retry ${OC} apply -f registry_pvc.yaml
```
cfergeau, line 181: Wondering if it's possible to do something like
```
OC="retry ${OC}"
```
so that all calls to `oc` go through a retry. Alternatively, this could be a `run_oc` function which calls `retry ${OC} ...`. Might be cleaner.

snc.sh
```diff
@@ -191,7 +191,8 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
 # Add a tip in the login page
-retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
+retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json > tmp_secret_template.json
+${JQ} -r '.data["login.html"]' tmp_secret_template.json | base64 -d > login.html
```
cfergeau, line 195: Wondering if
```diff
diff --git a/snc.sh b/snc.sh
index 4ffe9ad..d3dfaef 100755
--- a/snc.sh
+++ b/snc.sh
@@ -191,7 +191,7 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
 # Add a tip in the login page
-retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
+retry_xx /dev/null ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
 ${PATCH} login.html < login.html.patch
 retry ${OC} create secret generic login-template --from-file=login.html -n openshift-config
 
diff --git a/tools.sh b/tools.sh
index 288bd74..2cad5c2 100755
--- a/tools.sh
+++ b/tools.sh
@@ -88,7 +88,9 @@ if ! which ${PATCH}; then
     sudo yum -y install /usr/bin/patch
 fi
 
-function retry {
+function retry_xx {
+    local output=$1
+    shift
     local retries=10
     local count=0
     until "$@"; do
@@ -106,6 +108,10 @@ function retry {
     return 0
 }
 
+function retry {
+        retry /dev/stdout $@
+}
+
 function get_vm_prefix {
     local crc_vm_name=$1
     # This random_string is created by installer and added to each resource type,
```
could work.

Alternatively, have you tried:
```
branding_template=$(retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json)
${JQ} ...
```

Actually I'm not sure I understand why the pipe won't work when the redirect does :-/
branding_template

praveenkumar: > Actually I'm not sure I understand why the pipe won't work when the redirect does :-/
```bash
#!/bin/bash

set -exuo pipefail

export LC_ALL=C.UTF-8
export LANG=C.UTF-8

function retry {
    local retries=4
    local count=0
    until "$@"; do
        exit=$?
        wait=$((2 ** $count))
        count=$(($count + 1))
        if [ $count -lt $retries ]; then
            echo "Retry $count/$retries exited $exit, retrying in $wait seconds..."
            sleep $wait
        else
            echo "Retry $count/$retries exited $exit, no more retries left."
            return $exit
        fi
    done
    return 0
}

function s_grep {
	echo "Hello there"
	false
}

retry false > tmp_file
```

cfergeau:
Ah, so the temporary file will also contain `Retry ..., retrying in ... seconds...` together with the json data, and we are lucky that `jq` accepts to parse this? Would be better to be able to cleanly separate the logs from retry from the actual json data?

snc.sh
```diff
@@ -191,7 +191,8 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
 # Add a tip in the login page
-retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
+retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json > tmp_secret_template.json
```
cfergeau, line 194: Is this commit needed now that `retry` output goes to stderr? I think pipe only picks up stdout and not stderr?

.gitignore
```diff
@@ -9,3 +9,4 @@ yq
 openshift-clients/
 podman-remote/
 .sw[a-p]
+tmp_pv.yaml
```
cfergeau, line 12: I'd still prefer to avoid temporary files as much as possible :)

snc.sh
praveenkumar: @cfergeau yes this is still required because even the output is stderr but `retry <cmd1> | <cmd2>` still going to execute the cmd2 on first failure, what we want is cmd2 only execute if cmd1 successful.
```bash
#!/bin/bash

set -exuo pipefail

export LC_ALL=C.UTF-8
export LANG=C.UTF-8

function retry {
    local retries=4
    local count=0
    until "$@"; do
        exit=$?
        wait=$((2 ** $count))
        count=$(($count + 1))
        if [ $count -lt $retries ]; then
            echo "Retry $count/$retries exited $exit, retrying in $wait seconds..." 1>&2
            sleep $wait
        else
            echo "Retry $count/$retries exited $exit, no more retries left." 1>&2
            return $exit 
        fi
    done
    return 0
}

retry false | echo "Hello" > foo.txt
```

.gitignore
praveenkumar: @cfergeau yes but I don't see options here :(

snc-library.sh
```diff
@@ -170,19 +170,20 @@ function create_pvs() {
 
     for pvname in $(seq -f "pv%04g" 1 ${count}); do
         if ! ${OC} get pv "${pvname}" &> /dev/null; then
-            generate_pv "${pvdir}/${pvname}" "${pvname}" | ${OC} create -f -
+	    pv_template=$(generate_pv "${pvdir}/${pvname}" "${pvname}")
+            retry ${OC} create -f <(echo ${pv_template})
```
cfergeau, line 174: indentation mismatch it seems, one line with tabs, the other with spaces?

cfergeau (approved):
I'd add some explanation to the commit log for ` Add cluster-cloud-controller-manager-operator to cvo override list`
If we need to revisit that decision in 1 year, will be useful to remember why we decided it was not needed in the past.

snc.sh
```diff
@@ -191,7 +191,8 @@ retry ${OC} scale --replicas=1 ingresscontroller/default -n openshift-ingress-op
 retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch '{"spec":{"defaultRoute":true}}' --type=merge
 
 # Add a tip in the login page
-retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json | ${JQ} -r '.data["login.html"]'  | base64 -d > login.html
+secret_template=$(retry ${OC} get secrets -n openshift-authentication v4-0-config-system-ocp-branding-template -o json)
+${JQ} -r '.data["login.html"]' <(echo "${secret_template}") | base64 -d > login.html
```
cfergeau, line 195: This can be `echo ${secret_template} | ${JQ} ...`. Not familiar at all with the `<(echo ...)` syntax even if I was the one to bring it up, so don't know if this is behaving as expected.

snc-library.sh
```diff
@@ -170,19 +170,20 @@ function create_pvs() {
 
     for pvname in $(seq -f "pv%04g" 1 ${count}); do
         if ! ${OC} get pv "${pvname}" &> /dev/null; then
-            generate_pv "${pvdir}/${pvname}" "${pvname}" | ${OC} create -f -
+            pv_template=$(generate_pv "${pvdir}/${pvname}" "${pvname}")
+            retry ${OC} create -f <(echo "${pv_template}")
```
cfergeau, line 174: I think anything involving `<` `|` won't work as expected with retry, as the first try will consume the data which was produced by `echo`, and the next tries will have nothing to read from.
```
diff --git a/tools.sh b/tools.sh
index 288bd74..d22426a 100755
--- a/tools.sh
+++ b/tools.sh
@@ -91,7 +91,13 @@ fi
 function retry {
     local retries=10
     local count=0
-    until "$@"; do
+    if [ -t 0 ]; then
+        # no pipe
+        input=""
+    else
+        # pipe, save stdin
+        input="$(cat)"
+    fi
+
+    until echo "$input" | "$@"; do
         exit=$?
         wait=$((2 ** $count))
         count=$(($count + 1))
```
might help, but I expect there will be other corner cases I'm not familiar with. An intermediate function may help (one which runs `echo ${pv_template} | ${OC} create -f` so that we can then run `retry intermediate_function ${pv_template}`)

praveenkumar: Created https://github.com/code-ready/snc/issues/460

PR #461, opened 2021-08-17
Compare machine config before and after apply pull secret

praveenkumar:
In current sceario when we wait for cluster to stable after removing
the pull secret then imagepull error happen for all the marketplace
images since it require the valid pull secret.

```
+ ./openshift-clients/linux/oc get pod --no-headers --all-namespaces
+ grep -v Running
+ grep -v Completed
openshift-marketplace                        certified-operators-n4vlj                                0/1   ImagePullBackOff   0     112m
openshift-marketplace                        community-operators-h5s2d                                0/1   ImagePullBackOff   0     113m
openshift-marketplace                        redhat-marketplace-rm9lm                                 0/1   ImagePullBackOff   0     112m
openshift-marketplace                        redhat-operators-9tqpf                                   0/1   ImagePullBackOff   0     113m
```

This patch will just make sure to wait till machine config render
properly.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -294,11 +294,13 @@ function download_podman() {
       mkdir -p podman-remote/mac
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-darwin.zip -o podman-remote/mac/podman.zip
       ${UNZIP} -o -d podman-remote/mac/ podman-remote/mac/podman.zip
+      mv podman-remote/mac/podman-${version}/podman  podman-remote/mac
     fi
 
     if [ -n "${SNC_GENERATE_WINDOWS_BUNDLE}" ]; then
       mkdir -p podman-remote/windows
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-windows.zip -o podman-remote/windows/podman.zip
       ${UNZIP} -o -d podman-remote/windows/ podman-remote/windows/podman.zip
+      mv podman-remote/windows/podman-${version}/podman.exe  podman-remote/windows
     fi
 }
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -267,7 +267,7 @@ function wait_till_cluster_stable() {
     local count=0
 
     # Remove all the failed Pods
-    ${OC} delete pods --field-selector=status.phase=Failed -A || true
+    retry ${OC} delete pods --field-selector=status.phase=Failed -A
 
     local a=0
     while [ $a -lt $retryCount ]; do
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -256,7 +256,8 @@ function no_operators_degraded() {
 }
 
 function all_pods_are_running_completed() {
-    ! ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed
+    local ignoreNamespace=$1
+    ! ${OC} get pod --no-headers --all-namespaces --field-selector=metadata.namespace!="${ignoreNamespace}" | grep -v Running | grep -v Completed
 }
 
 function wait_till_cluster_stable() {
@@ -265,6 +266,7 @@ function wait_till_cluster_stable() {
     local retryCount=30
     local numConsecutive=3
     local count=0
+    local ignoreNamespace=${1:-"none"}
 
     # Remove all the failed Pods
     retry ${OC} delete pods --field-selector=status.phase=Failed -A
@@ -287,6 +289,6 @@ function wait_till_cluster_stable() {
     done
 
     # Wait till all the pods are either running or complete state
-    retry all_pods_are_running_completed
+    retry all_pods_are_running_completed "${ignoreNamespace}"
 }
 
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -255,6 +255,10 @@ function no_operators_degraded() {
     ${OC} get co -ojsonpath='{.items[*].status.conditions[?(@.type=="Degraded")].status}' | grep -v True
 }
 
+function all_pods_are_running_completed() {
+    ! ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed
+}
+
 function wait_till_cluster_stable() {
     sleep 1m
 
@@ -283,8 +287,6 @@ function wait_till_cluster_stable() {
     done
 
     # Wait till all the pods are either running or complete state
-    while ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed; do
-       sleep 2
-    done
+    retry all_pods_are_running_completed
 }
 
--- a/snc.sh
+++ b/snc.sh
@@ -239,11 +239,15 @@ retry ${OC} delete mc chronyd-mask
 # Wait for the cluster again to become stable because of all the patches/changes
 wait_till_cluster_stable
 
+mc_before_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
 # Replace pull secret with a null json string '{}'
 retry ${OC} replace -f pull-secret.yaml
+mc_after_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
 
-# Wait for the cluster again to become stable because of remove of pull secret
-wait_till_cluster_stable
+while [ "${mc_before_removing_pullsecret}" == "${mc_after_removing_pullsecret}" ]; do
+	echo "Machine config is still not rendered"
+	mc_after_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
+done
 
 # Delete the pods which are there in Complete state
 retry ${OC} delete pod --field-selector=status.phase==Succeeded --all-namespaces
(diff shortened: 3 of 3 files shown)


snc.sh
```diff
@@ -239,11 +239,15 @@ retry ${OC} delete mc chronyd-mask
...
 # Replace pull secret with a null json string '{}'
 retry ${OC} replace -f pull-secret.yaml
+mc_after_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
 
-# Wait for the cluster again to become stable because of remove of pull secret
-wait_till_cluster_stable
+while [ "${mc_before_removing_pullsecret}" == "${mc_after_removing_pullsecret}" ]; do
+	echo "Machine config is still not rendered"
+	mc_after_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
+done
```
cfergeau, line 250: When this loop ends, all operators which can stabilize are either Running or Completed? This is doing far less things than what `wait_till_cluster_stable` was doing.

snc.sh
```diff
@@ -239,11 +239,15 @@ retry ${OC} delete mc chronyd-mask
...
 wait_till_cluster_stable
 
+mc_before_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
 # Replace pull secret with a null json string '{}'
 retry ${OC} replace -f pull-secret.yaml
+mc_after_removing_pullsecret=$(retry ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname)
 
-# Wait for the cluster again to become stable because of remove of pull secret
-wait_till_cluster_stable
+while [ "${mc_before_removing_pullsecret}" == "${mc_after_removing_pullsecret}" ]; do
```
cfergeau, line 247: Should these lists be filtered on `rendered_worker` or `rendered_master` in case an unrelated machineconfig comes up or disappears?

praveenkumar: @cfergeau ideally if we are not making any change in the mco then this shouldn't happen and if that is the case then we need to file bug on mco side.

snc.sh
praveenkumar: @cfergeau yes but we we need to use the `wait_till_cluster_stable` function then need to ignore the `openshift-marketplace` pods otherwise it will always fail as I mentioned in the commit logs.

snc-library.sh
```diff
@@ -283,8 +283,15 @@ function wait_till_cluster_stable() {
     done
 
     # Wait till all the pods are either running or complete state
+    # Add 5 mins timeout
+    runtime="5 minute"
+    endtime=$(date -ud "$runtime" +%s)
```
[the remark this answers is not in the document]
praveenkumar: Sure I can just add that.

snc-library.sh
```diff
@@ -283,8 +283,15 @@ function wait_till_cluster_stable() {
...
     while ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed; do
        sleep 2
+       if [[ $(date -u +%s) -ge $endtime ]]; then
+	   echo "Not able to get all the containers in running/completed state"
+	   exit 1
+       fi
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau retry where?

snc.sh
cfergeau: I understand we cannot use `wait_till_cluster_stable` as is, I'm asking if the checks done by the new code are enough to get the cluster to the state we want?

snc.sh
cfergeau: Dunno, sounds like relying a bit too much on something we don't control :)

snc-library.sh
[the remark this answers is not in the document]
cfergeau: This `while` loop is relatively close to `retry` implementation, could `retry` be used instead of that loop?

createdisk.sh
```diff
@@ -131,7 +131,7 @@ create_tarball "$libvirtDestDir"
 # the content of $libvirtDestDir
 if [ -n "${SNC_GENERATE_MACOS_BUNDLE}" ]; then
     hyperkitDestDir="crc_hyperkit_${destDirSuffix}"
-    generate_hyperkit_bundle "$libvirtDestDir" "$hyperkitDestDir" "$1" "$kernel_release" "$kernel_cmd_line"
+    generate_hyperkit_bundle "$libvirtDestDir" "$hyperkitDestDir" "$1" "$kernel_release" "$kernel_cmd_line" "$podman_version"
```
cfergeau, line 134: Would be better to let `download_podman` handle this rather than carrying around the podman version. This is 4.9-only isn't this?

praveenkumar: @cfergeau you mean `download_podman` should download => extract => put the podman executable in top dir? 

Yes it is only 4.9 side.

cfergeau: > @cfergeau you mean download_podman should download => extract => put the podman executable in top dir?

Yup, something like this:
```
diff --git a/createdisk-library.sh b/createdisk-library.sh
index 25f4bda..a515614 100755
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -294,11 +294,13 @@ function download_podman() {
       mkdir -p podman-remote/mac
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-darwin.zip -o podman-remote/mac/podman.zip
       ${UNZIP} -o -d podman-remote/mac/ podman-remote/mac/podman.zip
+      mv podman-remote/mac/podman-${podman-version}/podman  podman-remote/mac
     fi
 
     if [ -n "${SNC_GENERATE_WINDOWS_BUNDLE}" ]; then
       mkdir -p podman-remote/windows
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-windows.zip -o podman-remote/windows/podman.zip
       ${UNZIP} -o -d podman-remote/windows/ podman-remote/windows/podman.zip
+      mv podman-remote/windows/podman-${podman-version}/podman  podman-remote/windows
     fi
 }
```

cfergeau (commented):
Still some open questions in the first and second commits, the other 2 look good.

snc.sh
praveenkumar: @cfergeau As of now I am depend on the CI for all the changes and if it is green that mean bundle which is created can work with crc and execute the required tests. If we want extra layer of surety then may be I will create another PR to wait again for all the operators except market place but this check also required to make sure MCO took action and new machine config is generated.

snc.sh
praveenkumar: @cfergeau As soon as we hit `openshift-installer` we really don't control the flow till cluster is provisioned so if there is issue it should be openshift side.

snc.sh
praveenkumar: @cfergeau Added another comment to ignore the openshift-marketplace namespace in the `wait_till_cluster_stable` function. and using along with this.

snc-library.sh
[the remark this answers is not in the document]
cfergeau: Still wondering if this could be something like:
```
function all_pods_running_or_completed {
    # not fully sure what this what needs to be done here so that retry() considers this failed/succeess
        ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed
}

    retry all_pods_running_or_completed
```

createdisk-library.sh
```diff
@@ -294,11 +294,13 @@ function download_podman() {
       mkdir -p podman-remote/mac
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-darwin.zip -o podman-remote/mac/podman.zip
       ${UNZIP} -o -d podman-remote/mac/ podman-remote/mac/podman.zip
+      mv podman-remote/mac/podman-${version}/podman  podman-remote/mac
```
cfergeau, line 297: This means `master` and the `4.8` branches are out of sync, and this  commit is not the same as the one which went into 4.8 :-/ (slightly different commit message).

praveenkumar: @cfergeau yes I can drop this and cherry pick what went in for 4.8

snc-library.sh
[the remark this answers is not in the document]
praveenkumar: @cfergeau in this scenario `${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed` should return non zero value as soon as all the containers are in running/completed state so initial phase of retry is going to successful. What we actually want `all_pods_running_or_completed` function shouldn't successful until all the pods are in running or completed state so it should be `! ${OC} get pod --no-headers --all-namespaces | grep -v Running | grep -v Completed`

createdisk-library.sh
praveenkumar: done

snc-library.sh
```diff
@@ -265,6 +266,7 @@ function wait_till_cluster_stable() {
     local retryCount=30
     local numConsecutive=3
     local count=0
+    local ignoreNamespace=${1:-"none"}
```
cfergeau, line 269: Not sure the `none` default value will do the right thing? maybe we can just ignore it?

cfergeau (approved):

snc-library.sh
praveenkumar: @cfergeau it is doing as expected until there is some default namespace named as `none` appear which is highly a case.

cfergeau: I have no idea what is expected with `none` :) a failure ?

PR #462, opened 2021-08-24
Use podman-<version> dir to copy podman executable

praveenkumar:
This should fix
```
+ unzip -o -d podman-remote/mac/ podman-remote/mac/podman.zip
Archive:  podman-remote/mac/podman.zip
   creating: podman-remote/mac/podman-3.2.3/

+ cp podman-remote/mac/podman crc_hyperkit_461/
cp: cannot stat 'podman-remote/mac/podman': No such file or directory
```


cfergeau (approved):
Short log is not really accurate, looks good otherwise.

PR #464, opened 2021-09-03
Revert "Move podman binary from podman-<version> dir to top level dir"

guillaumerose:
This reverts commit fe98dd0a449800284484e4b9cdcd2c1776978d50.

4.8.5 is still shipping podman 3.0.1, this move is not needed for the
moment.

Without this, it triggers the following error:
```
+ mv podman-remote/mac/podman-3.0.1/podman podman-remote/mac
mv: cannot stat 'podman-remote/mac/podman-3.0.1/podman': No such file or directory
```

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -294,13 +294,11 @@ function download_podman() {
       mkdir -p podman-remote/mac
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-darwin.zip -o podman-remote/mac/podman.zip
       ${UNZIP} -o -d podman-remote/mac/ podman-remote/mac/podman.zip
-      mv podman-remote/mac/podman-${version}/podman  podman-remote/mac
     fi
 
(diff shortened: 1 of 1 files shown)


createdisk-library.sh
```diff
@@ -294,13 +294,11 @@ function download_podman() {
       mkdir -p podman-remote/mac
       curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-release-darwin.zip -o podman-remote/mac/podman.zip
       ${UNZIP} -o -d podman-remote/mac/ podman-remote/mac/podman.zip
-      mv podman-remote/mac/podman-${version}/podman  podman-remote/mac
```
cfergeau, line 297: Fwiw, I'd favour something like 
```
if [ -e podman-remote/mac/podman-${version}/podman ]; then
    mv podman-remote/mac/podman-${version}/podman  podman-remote/mac
fi
```
(or checking for the directory existence), but fine with me to go with this revert now, we can polish it when we need to re-revert

cfergeau (approved):

PR #466, opened 2021-09-09
Create systemd unit for dnsmasq container

praveenkumar:
It would help to run the dnsmasq container in case there is a reboot.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -56,10 +56,13 @@ if [ -n "${SNC_GENERATE_WINDOWS_BUNDLE}" ]; then
     prepare_hyperV "$1"
 fi
 
-# Add gvisor-tap-vsock service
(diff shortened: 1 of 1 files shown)


gbraad (approved):

guillaumerose:
This container is not needed with user network mode. Perhaps we can only keep this for linux bundle?

praveenkumar:
> This container is not needed with user network mode. Perhaps we can only keep this for linux bundle?

@guillaumerose This need to be part of disk image not with the bundle and as of now we dont' have a way to separate those out different disk image for different platform. We consumed same disk image for every platform.

guillaumerose:
It doesn't hurt to have this service. It is lightweight enough. No worry 👍

praveenkumar:
@guillaumerose we are not going to enable/start this service on windows/mac it is part of the image but just inactive.

PR #475, opened 2021-10-18
Use virsh command to get internal IP instead dig

praveenkumar:
With latest changes in libvirt provider the dns is part of network
which is created by installer and the following entry is present.
```
...
$ virsh net-dumpxml crc-crzxj
  <dns enable='yes'>
    <host ip='192.168.126.10'>
      <hostname>api.crc.testing</hostname>
      <hostname>api-int.crc.testing</hostname>
    </host>
    <host ip='192.168.126.11'>
      <hostname>api.crc.testing</hostname>
      <hostname>api-int.crc.testing</hostname>
    </host>
  </dns>
...
```
So now with dig command output the both the IP when query the
`api.crc.testing`

```
$ dig +short api.crc.testing
192.168.126.11
192.168.126.10
```
This lead to corrupt the crc bundle metadata and `crc start` fails
with this.

This patch make sure it will only use the IP of master node using
virsh command.
```
$ virsh domifaddr crc-crzxj-master-0|tail -2|head -1|awk '{print $4}'| cut -d/ -f1
192.168.126.11
```

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -37,7 +37,7 @@ VM_PREFIX=$(get_vm_prefix ${CRC_VM_NAME})
 ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmi --prune'
 
 # Get the IP of the VM
-INTERNAL_IP=$(${DIG} +short api.${CRC_VM_NAME}.${BASE_DOMAIN})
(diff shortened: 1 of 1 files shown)


cfergeau:
> `$ virsh domifaddr crc-crzxj-master-0|tail -2|head -1|awk '{print $4}'| cut -d/ -f1`

Might be easier to parse as 
```
$ virsh domifaddr crc-crzxj-master-0 | grep 192.168.126 | sed 's/.*\(192.168.122.[0-9]*\).*/\1/'
```

PR #481, opened 2021-11-02
enable podman remote rootless socket

anjannath:
user services can't be directly enabled in Butane configs and has
to be manually enabled

ref: https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-user-systemd-unit-on-boot/


gbraad (approved):

gbraad:
@praveenkumar PTAL

PR #482, opened 2021-11-08
Add cockpit podman related packages

praveenkumar:
This will allow to have cockpit integration with tray.


gbraad:
Will have to test this image

createdisk-library.sh
```diff
@@ -116,7 +116,7 @@ function prepare_hyperV() {
     # Install the hyperV rpms to VM
     ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
     ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
-    ${SSH} core@${vm_ip} -- 'sudo rpm-ostree install --allow-inactive hyperv-daemons'
+    ${SSH} core@${vm_ip} -- 'sudo rpm-ostree install --allow-inactive hyperv-daemons cockpit-bridge cockpit-ws cockpit-podman'
```
cfergeau, line 119: Why is this done in prepare_hyperV? Is this specific to the Windows bundle? The commit log does not say anything about it. What is the impact on bundle size?

praveenkumar: > Why is this done in prepare_hyperV? Is this specific to the Windows bundle?

No it is not only specific to windows bundle but as of now this is the place where we enable the repo and install the packages so we might need to separate this out.

> The commit log does not say anything about it. What is the impact on bundle size?

@cfergeau  It increase from 641 => 671MB, will mention in the commit message.

praveenkumar:
/hold

Till it is tested by @gbraad

createdisk-library.sh
```diff
@@ -111,16 +111,24 @@ function copy_additional_files {
     update_json_description $srcDir $destDir $podmanVersion
 }
 
-function prepare_hyperV() {
+function install_packages() {
```
cfergeau, line 114: I'd name this `install_additional_packages`

createdisk-library.sh
```diff
@@ -111,16 +111,24 @@ function copy_additional_files {
...
     local vm_ip=$1
     # Install the hyperV rpms to VM
```
cfergeau, line 116: This comment can be removed.

createdisk-library.sh
```diff
@@ -111,16 +111,24 @@ function copy_additional_files {
...
+    fi
     ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora.repo'
     ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=1/enabled=0/ /etc/yum.repos.d/fedora-updates.repo'
     ${SSH} core@${vm_ip} -- 'sudo rpm-ostree cleanup --base'
     ${SSH} core@${vm_ip} -- 'sudo rpm-ostree cleanup --repomd'
+}
+
+function prepare_hyperV() {
+    local vm_ip=$1
+    # Install the hyperV rpms to VM
```
cfergeau, line 131: I'd remove this comment as well, `ssh $vm_ip rpm-ostree install hyperv-daemons` is self explanatory

createdisk.sh
```diff
@@ -27,6 +27,9 @@ ${SSH} core@${VM_IP} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
 
 install_packages ${VM_IP}
 
+# Enable cockpit socket
+${SSH} core@${VM_IP} -- 'sudo systemctl enable cockpit.socket'
```
cfergeau, line 31: I'd keep this in `install_packages`, `prepare_hyperV` is already doing more than just package installation. And this commit can be squashed in the previous one. The comment is also not really needed, it's easy enough to just read the shell code.

gbraad:
While Cockpit loads, the interface shows 'Not found'. Most likely parts are missing.
Will need to investigate this.

gbraad:
As per: https://cockpit-project.org/running.html#coreos
```
[root@localhost ~]# podman container runlabel INSTALL docker.io/cockpit/ws
systemctl enable cockpit.service
cockpit-ws must not be installed in the host
Error: `/proc/self/exe run --rm --privileged -v /:/host -e IMAGE=docker.io/cockpit/ws:latest docker.io/cockpit/ws:latest /container/atomic-install docker.io/cockpit/ws:latest` failed: exit status 1
The unit files have no installation config (WantedBy=, RequiredBy=, Also=,
Alias= settings in the [Install] section, and DefaultInstance= for template
units). This means they are not meant to be enabled using systemctl.

Possible reasons for having this kind of units are:
• A unit may be statically enabled by being symlinked from another unit's
  .wants/ or .requires/ directory.
• A unit's purpose may be to act as a helper for some other unit which has
  a requirement dependency on it.
• A unit may be started when needed via activation (socket, path, timer,
  D-Bus, udev, scripted systemctl call, ...).
• In case of template units, the unit is meant to be enabled with some
  instance name specified.
```

So we should not use `cockpit-ws` as installed package

createdisk-library.sh
```diff
@@ -115,6 +115,7 @@ function install_additional_packages() {
     local vm_ip=$1
     ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
     ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
+    ${SSH} core@${vm_ip} -- 'sudo rpm-ostree install --allow-inactive cockpit-bridge cockpit-ws cockpit-podman'
```
gbraad, line 118: `rpm-ostree install cockpit-system cockpit-ostree cockpit-bridge cockpit-podman`

gbraad (changes requested):
... as per instructions we can not install `-ws` directly on the host.

```
echo 'PasswordAuthentication yes' | sudo tee /etc/ssh/sshd_config.d/02-enable-passwords.conf
sudo systemctl try-restart sshd

podman container runlabel --name cockpit-ws RUN docker.io/cockpit/ws
podman container runlabel INSTALL docker.io/cockpit/ws
systemctl enable cockpit.service

gbraad:
Discussed with @martinpitt and this should work as expected. We do not need the `-system` package as this would pull in the system module that enables the terminal, etc. we do not need (want) that.

gbraad (approved):

PR #484, opened 2021-11-11
Force to navigate podman images in cockpit view

praveenkumar:
This patch add cockpit config related files which force it to put
container view during navigation.


praveenkumar:
/hold

till @gbraad test this bundle.

gbraad (approved):
LGTM, but will have to test

PR #488, opened 2021-12-02
Fix disk size to 31G from 10G

praveenkumar:
By default on the crc side we have default disk size is 31G and for
podman bundle disk resize always run which should be avoided. This
patch will increase it to default size.

If bundle size increase because of this patch then we make the change
in crc side.

- https://github.com/code-ready/crc/issues/2862


cfergeau:
Not sure we need 31GiB with podman though?

praveenkumar:
> Not sure we need 31GiB with podman though?

@cfergeau that is depend on the users around how big/how many container images they are using, 31GiB is not too much.

PR #489, opened 2021-12-08
Fix directory permission created by ignition

praveenkumar:
As per ignition documents[0] if we make a series of directory something
like `mkdir -p /a/b/c/d` then `directories` plugin only create `d` with
assigned user/group but leading directories `/a/b/c` are created as
root.

- https://coreos.github.io/ignition/configuration-v3_2/ (directories
  session)


cfergeau:
`s/session/section` in the commit log. I don't see where the documentation is saying the intermediate directories will be created as root though? And there is no `directories` directive before this change, so I'd add a bit more to the problem description. I assume one should look at the `links` section, where `/home/core/.config/systemd/user/default.target.wants/podman.socket` is created, all the subdirectiors in `/home/core/.config/systemd/user/default.target.wants/` are created as root/root?

praveenkumar:
> all the subdirectiors in /home/core/.config/systemd/user/default.target.wants/ are created as root/root?

@cfergeau yes

cfergeau:
Can you add this to the commit log? As it is, it's not very clear what to look at if one wants to reproduce the bug.

praveenkumar:
Sure I will add it in the commit logs also https://github.com/containers/podman/blob/main/pkg/machine/ignition.go#L153-L154 may be put a similar kind of comment to config also.

fcos-config.yaml
```diff
@@ -72,3 +72,22 @@ storage:
...
+      user:
+        name: core
+      group:
+        name: core
+      contents:
+        inline: |
+         [containers]
+         netns="bridge"
+         rootless_networking="cni"
+      # Set machine_enabled to true to indeicate we're in VM
```
cfergeau, line 87: 'indicate'

images/openshift-ci/google-cloud-sdk.repo
```diff
@@ -1,6 +1,6 @@
 [google-cloud-sdk]
 name=Google Cloud SDK
-baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el7-x86_64
+baseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64
 enabled=1
```
cfergeau, line 4: Should this be squashed with `CI: Update image from centos 7 to 8` ?

cfergeau (approved):

fcos-config.yaml
```diff
@@ -10,6 +10,34 @@ systemd:
    - name: podman.socket
      enabled: true
 storage:
+  # Ignition has a bug/feature? where if you make a series of dirs
+  # in one swoop, then the leading dirs are creates as root.
```
cfergeau, line 14: 'creates' -> 'created' 

'session' -> 'section' in the commit log, and 'subdirectiors' -> 'subdirectories'

cfergeau (approved):
I guess most of this needs to be backported to all other branches?

praveenkumar:
> I guess most of this needs to be backported to all other branches?

@cfergeau yes we need to backport some stuff (mostly the dockerfile related) to master also.

PR #491, opened 2021-12-11
Drop version from 'group' from cvo override

praveenkumar:
Now that the CVO looks at the group property validation and for group
it is different than apiVersion property which required the version
info. This patch drop the version info as part of group.

- https://github.com/openshift/enhancements/pull/982
- https://github.com/openshift/cluster-version-operator/pull/689


gbraad (approved):

cfergeau:
Could have been useful to explicitly mention in the commit log that without this, the various components listed in our cvo override file do not get disabled as expected.

PR #494, opened 2021-12-17
Add cockpit bearer token script and unit file

praveenkumar:
This patch should enable cockpit to use with bearer token and
make sure only user level service is running.

- https://gist.github.com/gbraad/c556d82394ca0dcc333e9609974fcaf8

It is using https://docs.fedoraproject.org/en-US/fedora-coreos/tutorial-user-systemd-unit-on-boot
doc for adding the unit file for core user.

### DIFF
--- a/fcos-config.yaml
+++ b/fcos-config.yaml
@@ -46,7 +46,34 @@ storage:
         name: core
       target: /usr/lib/systemd/user/podman.socket
       hard: false
+    - path: /home/core/.config/systemd/user/default.target.wants/cockpit-core-user.service
+      user:
+        name: core
+      group:
+        name: core
+      target: /home/core/.config/systemd/user/cockpit-core-user.service
+      hard: false
   files:
+      # Add systemd unit to access cockpit with local session using
+      # cockpit-bridge and adding bearer token to config
+    - path: /home/core/.config/systemd/user/cockpit-core-user.service
+      mode: 0600
+      user:
+        name: core
+      group:
+        name: core
+      contents:
+        inline: |
+          [Unit]
+          Description=Cockpit-ws for Core user
+
+          [Service]
+          Environment="XDG_RUNTIME_DIR=/run/user/1000"
+          ExecStart=/usr/libexec/cockpit-ws --no-tls --local-session=/usr/bin/cockpit-bridge
+          Restart=on-failure
+
+          [Install]
+          WantedBy=multi-user.target
     - path: /var/lib/systemd/linger/core
       mode: 0644
     - path: /usr/local/share/cockpit/shell/manifest.json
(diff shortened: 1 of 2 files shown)


praveenkumar:
/hold

Till bundle tested by @gbraad

fcos-config.yaml
```diff
@@ -46,7 +46,33 @@ storage:
...
+      user:
+        name: core
+      group:
+        name: core
+      target: /home/core/.config/systemd/user/cockpit-core-user.service
+      hard: false
   files:
+      # Add systemd unit for the cockpit bearer token
+    - path: /home/core/.config/systemd/user/cockpit-core-user.service
+      mode: 0755
```
cfergeau, line 59: Does this need to be executable? the default permissions are probably good enough?

fcos-config.yaml
```diff
@@ -46,7 +46,33 @@ storage:
...
       hard: false
+    - path: /home/core/.config/systemd/user/default.target.wants/cockpit-core-user.service
+      user:
+        name: core
+      group:
+        name: core
+      target: /home/core/.config/systemd/user/cockpit-core-user.service
+      hard: false
   files:
+      # Add systemd unit for the cockpit bearer token
```
cfergeau, line 57: Fwiw, it's really not obvious what in the unit file is related to a 'bearer token'. I see a cockpit service being started as the core user, this is reflected in the name of the unit file `cockpit-core-user.service`. If this is key to connecting to cockpit without auth, maybe it's worth giving more details how this works in the PR?

fcos-config.yaml
```diff
@@ -91,3 +117,48 @@ storage:
...
+        inline: |
+          #!/bin/sh
+          # This file is part of CRC
+          # Note: this is a POC and should not be used in production
+          #
+          # Copyright (C) 2021 Red Hat, Inc.
+          set -eux
+          HOST="$1"
+          echo "Bearer auth attempt from $HOST" >&2
+          # FIXME: should be random
```
cfergeau, line 132: maybe this could be fixed?

cfergeau (approved):
Some more explanations could be useful, looks good otherwise.

gbraad:
Will do this weekend or Monday. Had to focus on getting the pullsecret stored after the setup phase.

fcos-config.yaml
gbraad: We need to do that in sync with CRC, as the tray-electron would need to read this bearer and use for authentication. Let's start with this as part of the test and then implement a file or http API to retrieve the needed bearer.

fcos-config.yaml
```diff
@@ -104,3 +131,45 @@ storage:
...
+          echo "got response '$response'" >&2
+
+          # looks like '{"command":"authorize","cookie":"c1","response":"Bearer 123foo"}'
+          if [ "$(echo "$response" | jq -r .cookie)" != "$COOKIE" ]; then
+              # FIXME: send proper "problem" JSON
+              echo "bad cookie" >&2
+              exit 1
+          fi
+
+          if [ "$(echo "$response" | jq -r .response)" != "Bearer Y29yZQ==" ]; then
```
gbraad, line 161: this is the value that needs to be synced between the image and https://github.com/gbraad/podman-desktop/blob/9d13968a58eab54985fbabe23f58f0f3d9937d9a/main.js#L15

fcos-config.yaml
gbraad: https://github.com/code-ready/tray-electron/issues/24#issue-1082456798

gbraad (changes requested):
This needs the cockpit service disabled, else port 9090 is claimed.

gbraad:
```
Created symlink /var/home/core/.config/systemd/user/multi-user.target.wants/cockpit-core-user.service → /var/home/core/.config/systemd/user/cockpit-core-user.service.
Unit /var/home/core/.config/systemd/user/cockpit-core-user.service is added as a dependency to a non-existent unit multi-user.target.
```

the dependency is probably not correct

gbraad (approved):

Issue #496, opened 2022-01-05
[Podman] Remove moby packages from the image before creating bundle


praveenkumar:
As of now by default FCOS have moby/docker package and we really don't want those as  part of podman bundle so better to remove those and it will reduce the image size also.

praveenkumar:
This is now fixed and we don't have moby-engine package in our bundle.

PR #497, opened 2022-01-05
Remove moby-engine package from base layer of the image

praveenkumar:
This patch remove moby-package from the base layer of the image,
which is not required and might reduce the size of final bundle.

fixes: #496

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -25,6 +25,9 @@ VM_IP=$(arp -an | grep $(sudo virsh dumpxml ${CRC_VM_NAME} | grep '<mac' | grep 
 # Remove audit logs
 ${SSH} core@${VM_IP} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
 
+# Remove moby-engine package
+${SSH} core@${VM_IP} -- 'sudo rpm-ostree override remove moby-engine'
+
 install_additional_packages ${VM_IP}
 
 # Add gvisor-tap-vsock
@@ -39,7 +42,7 @@ EOF
 # https://bugzilla.redhat.com/show_bug.cgi?id=1956739
 ${SSH} core@${VM_IP} -- 'touch ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/authorized_keys'
 
-# Shutdown and Start the VM after installing the hyperV daemon packages.
+# Shutdown and Start the VM after modifying the set of installed packages
 # This is required to get the latest ostree layer which have those installed packages.
 shutdown_vm ${CRC_VM_NAME}
 start_vm ${CRC_VM_NAME} ${VM_IP}
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -47,6 +47,15 @@ ${SSH} core@${VM_IP} -- 'touch ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/autho
 shutdown_vm ${CRC_VM_NAME}
 start_vm ${CRC_VM_NAME} ${VM_IP}
 
+# Remove the base deployment from rpm-ostree
+${SSH} core@${VM_IP} -- 'sudo rpm-ostree cleanup -r'
+# Shutdown and Start the VM after removing base deployment tree
+# This is required because kernel commandline changed, namely
+# ostree=/ostree/boot.1/fedora-coreos/$hash/0 which switches 
+# between boot.0 and boot.1 when cleanup is run
+shutdown_vm ${CRC_VM_NAME}
+start_vm ${CRC_VM_NAME} ${VM_IP}
+
 # Only used for hyperkit bundle generation
 if [ -n "${SNC_GENERATE_MACOS_BUNDLE}" ]; then
     # Get the rhcos ostree Hash ID


cfergeau (approved):
Not sure it will actually reduce the final bundle size, as I think we'll still have a 'rollback' ostree layer with the package in the final image. 
Might be worth rephrasing this comment below:
```
# Shutdown and Start the VM after installing the hyperV daemon packages.
```
The restart of the VM is also needed to pick a layer without the moby-engine package.

praveenkumar:
> Not sure it will actually reduce the final bundle size, as I think we'll still have a 'rollback' ostree layer with the package in the final image.

@cfergeau Yes, I think we really don't need rollbacks and only need to have single commit/deployment as part of `rpm-ostree status`

```
# rpm-ostree status
State: idle
AutomaticUpdatesDriver: Zincati
  DriverState: active; periodically polling for updates (last checked Wed 2022-01-05 10:04:43 UTC)
Deployments:
● fedora:fedora/x86_64/coreos/stable
                   Version: 35.20211203.3.0 (2021-12-17T20:48:38Z)
                BaseCommit: 3f9ddd6ffe57e3bb27d1f92354d7349afd9775180b7eb75272851d42d247d3e0
              GPGSignature: Valid signature by 787EA6AE1147EEE56C40B30CDB4639719867C58F
       RemovedBasePackages: moby-engine 20.10.11-1.fc35
           LayeredPackages: cockpit-bridge cockpit-podman cockpit-ws hyperv-daemons

  fedora:fedora/x86_64/coreos/stable
                   Version: 35.20211203.3.0 (2021-12-17T20:48:38Z)
                    Commit: 3f9ddd6ffe57e3bb27d1f92354d7349afd9775180b7eb75272851d42d247d3e0
              GPGSignature: Valid signature by 787EA6AE1147EEE56C40B30CDB4639719867C58F
```

createdisk.sh
```diff
@@ -25,6 +25,11 @@ VM_IP=$(arp -an | grep $(sudo virsh dumpxml ${CRC_VM_NAME} | grep '<mac' | grep
 # Remove audit logs
 ${SSH} core@${VM_IP} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
 
+# Remove moby-engine pacakge
```
cfergeau, line 28: `pacakge` -> `package`

createdisk.sh
```diff
@@ -25,6 +25,11 @@ VM_IP=$(arp -an | grep $(sudo virsh dumpxml ${CRC_VM_NAME} | grep '<mac' | grep
...
+# Shutdown and Start the VM after installing the hyperV daemon packages to have 
+# latest rpm-ostree deployment doesn't have this package
```
cfergeau, line 30: Is this an intended change? Or an attempt to change the comment at line 42/47 which ended in the wrong place?

createdisk.sh
```diff
@@ -72,6 +72,9 @@ podman_version=$(${SSH} core@${VM_IP} -- 'rpm -q --qf %{version} podman')
 ${SSH} core@${VM_IP} -- 'sudo journalctl --rotate'
 ${SSH} core@${VM_IP} -- 'sudo journalctl --vacuum-time=1s'
 
+# Remove the base deployment from rpm-ostree
+${SSH} core@${VM_IP} -- 'sudo rpm-ostree cleanup -r'
```
cfergeau, line 76: Do you know by how much this changes (reduces hopefully :) the size of the bundle?

praveenkumar: It reduced the size around 40MB with generated bundle.

createdisk.sh
```diff
@@ -39,7 +42,7 @@ EOF
 # https://bugzilla.redhat.com/show_bug.cgi?id=1956739
 ${SSH} core@${VM_IP} -- 'touch ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/authorized_keys'
 
-# Shutdown and Start the VM after installing the hyperV daemon packages.
+# Shutdown and Start the VM after installing the hyperV daemon and removing moby packages.
```
cfergeau, line 45: `after modifying the set of installed packages` or such? This way we don't have to keep it up to date?

createdisk.sh
cfergeau: This is still misspelt.

sha256sum.txt
```diff
@@ -0,0 +1,3 @@
+38336e625c920b7a06264d5240f95844b1e65985aa06ef8886b17f402f3781d6  crc-macos-amd64.pkg
+2034e4abb1c11a0bf043af11dc260171c5289db088b352730156ccd3dca30962  crc-windows-installer.zip
+87824d2700a0125d08cfad87dce587f256cc0cefc7a9ebf1fbea863a50152dd8  crc-linux-amd64.tar.xz
```
cfergeau, line 3: probably mistakenly added?

createdisk.sh
```diff
@@ -47,6 +47,13 @@ ${SSH} core@${VM_IP} -- 'touch ~/.ssh/authorized_keys && chmod 0600 ~/.ssh/autho
 shutdown_vm ${CRC_VM_NAME}
 start_vm ${CRC_VM_NAME} ${VM_IP}
 
+# Remove the base deployment from rpm-ostree
+${SSH} core@${VM_IP} -- 'sudo rpm-ostree cleanup -r'
+# Shutdown and Start the VM after removing base deployment tree
+# This is required to get correct ostree commit in `/proc/cmdline` for macos
+shutdown_vm ${CRC_VM_NAME}
+start_vm ${CRC_VM_NAME} ${VM_IP}
```
cfergeau, line 55: This is unexpected, `rpm-ostree cleanup -r` changes the commit currently being used? Since we just stopped/started the VM, I was thinking the commit would be the right one.

praveenkumar: > This is unexpected, `rpm-ostree cleanup -r` changes the commit currently being used?

@cfergeau no it just remove the rollback commit not currently being used and then update the bootloader. Since it update the bootversion that's why we need to restart to get the latest on for mac.
```
Bootloader updated; bootconfig swap: yes; bootversion: boot.1.1, deployment count change: -1
```

> Since we just stopped/started the VM, I was thinking the commit would be the right one.
Commit is correct one.

createdisk.sh
praveenkumar: fixed

sha256sum.txt
praveenkumar: yes :(

createdisk.sh
cfergeau: The comment is misleading then, as it specifically mentions `ostree commit`: `This is required to get correct ostree commit in `/proc/cmdline` for macos`.
As you explained, it really is the kernel commandline which gets changed, namely `ostree=/ostree/boot.1/fedora-coreos/$hash/0` which switches between `boot.0` and `boot.1` when we run the cleanup.

praveenkumar: @cfergeau updated

cfergeau (approved):

PR #500, opened 2022-01-12
read the bearer token from a file instead of using a constant string

anjannath:
the file is updated with a random string on each start by crc

### DIFF
--- a/fcos-config.yaml
+++ b/fcos-config.yaml
@@ -70,8 +70,7 @@ storage:
           Description=Cockpit-ws for Core user
 
           [Service]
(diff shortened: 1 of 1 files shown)


gbraad:
This could work, as the script is invoked on each authentication call.

gbraad:
but why not read a file called something like `token` that is stored in the `core`'s user folder. This can be a simple `scp` or write in place without having to match a pattern. That somehow feels fragile as it prevents easy editing of the script file.

gbraad (approved):

PR #503, opened 2022-02-02
Misc changes to create podman bundle on different arch

praveenkumar:


tools.sh
```diff
@@ -42,6 +43,22 @@ if test -z ${YQ-}; then
     YQ=./yq
 fi
 
+# Download butane for creating ignition config
+if test -z ${BUTANE-}; then
+    echo "Downloading butane binary to create ignition config"
+    curl -L https://github.com/coreos/butane/releases/download/v0.14.0/butane-${ARCH}-unknown-linux-gnu -o butane
```
cfergeau, line 49: Did you check with the coreos people if they could build aarch64 container images instead?

praveenkumar: @cfergeau yes I did check with them and response I got `afaik coreos/ ones are auto-built directly by Quay. So the set of architectures depends on which auto-builders exist.` An on quay the auto-builders are x86_64 one.

praveenkumar:
As of now CI fails because of following.
```
$ curl -L https://github.com/coreos/ignition/releases/download/v2.13.0/ignition-validate-x86_64-linux -o ignition-validate
$ ./ignition-validate -h
./ignition-validate: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./ignition-validate)
./ignition-validate: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./ignition-validate)
$ rpm -qa | grep -i glibc
glibc-langpack-en-2.28-164.el8.x86_64
glibc-common-2.28-164.el8.x86_64
glibc-devel-2.28-164.el8.x86_64
glibc-2.28-164.el8.x86_64
glibc-headers-2.28-164.el8.x86_64
```

https://github.com/coreos/ignition/issues/1311 is recently merged so hopping if there is release soon then it should work as expected.

snc.sh
```diff
@@ -56,7 +56,7 @@ create_json_description
 sudo ${VIRT_INSTALL} --name=${CRC_VM_NAME} --vcpus=2 --ram=2048 \
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau As per man page , do we want to allow run `arm64` on `amd64` ?
```
--arch=ARCH
Request a non-native CPU architecture for the guest virtual machine. If omitted, the host CPU architecture will be used in the guest. ```

File: tools.sh (hunk 1 above)
Comment (cfergeau) [reply tools.sh:49]: We need to check the GPG signature of these binaries to have some validation regarding the origin of these binaries.

File: snc.sh (hunk 2 above)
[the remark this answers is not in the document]
Comment (cfergeau) [reply snc.sh:56]: After this PR, it's possible to download an aarch64 binary or virt image, and to try to run it with a x86_64 hypervisor. Imo, if we downloaded an aarch64 image, we should make sure virt-install knows it's using an aarch64 image (ie use the --arch aarch64 flag).

File: tools.sh (hunk 1 above)
Comment (cfergeau) [reply tools.sh:49]: Alternatively:
```
$ yum info butane
Available Packages
Name         : butane
Version      : 0.13.1
Release      : 1.fc35
Architecture : x86_64
Size         : 2.0 M
Source       : butane-0.13.1-1.fc35.src.rpm
Repository   : fedora
Summary      : Butane config transpiler
URL          : https://github.com/coreos/butane
License      : ASL 2.0
Description  : 
             : Butane translates human-readable Butane Configs into
             : machine-readable Ignition configs for provisioning operating
             : systems that use Ignition.

$ dnf info ignition-validate
Available Packages
Name         : ignition-validate
Version      : 2.13.0
Release      : 3.fc35
Architecture : x86_64
Size         : 1.4 M
Source       : ignition-2.13.0-3.fc35.src.rpm
Repository   : updates
Summary      : Validation tool for Ignition configs
URL          : https://github.com/coreos/ignition
License      : ASL 2.0
Description  : Ignition is a utility used to manipulate systems during the
             : initramfs. This includes partitioning disks, formatting
             : partitions, writing files (regular files, systemd units, etc.),
             : and configuring users. On first boot, Ignition reads its
             : configuration from a source of truth (remote URL, network
             : metadata service, hypervisor bridge, etc.) and applies the
             : configuration.
             : 
             : This package contains a tool for validating Ignition
             : configurations.
```

This is consistent with ` sudo yum -y install /usr/bin/coreos-installer`

Comment (praveenkumar) [reply tools.sh:49]: @cfergeau This is not available on the RHEL and since our CI uses RHEL system it is going to fail.

Comment (cfergeau) [reply tools.sh:49]: I can see `butane-0.14.0-1.rhaos4.10.el8` and `ignition-2.13.0-3.rhaos4.11.el8` in RHEL, maybe the CI system could be subscribed to these additional repos?

**Comment by cfergeau** [comment]
I added https://github.com/cfergeau/snc/commit/07c6a4c3a95492263aab49a4463f64b1d8a314ff to uncompress the aarch64 kernel, but I cannot easily test this now.

File: tools.sh (hunk 1 above)
Comment (cfergeau) [reply tools.sh:49]: I followed this https://docs.openshift.com/container-platform/4.10/machine_management/more-rhel-compute.html to get access to ignition-validate and butane:
```
$ sudo subscription-manager attach --pool=xxxx
$ sudo subscription-manager repos  --enable="rhocp-4.10-for-rhel-8-x86_64-rpms"
$ yum info ignition-validate butane
Not root, Subscription Management repositories not updated
Last metadata expiration check: 0:02:07 ago on Fri May 13 13:47:43 2022.
Available Packages
Name         : butane
Version      : 0.14.0
Release      : 1.rhaos4.10.el8
Architecture : x86_64
Size         : 2.0 M
Source       : butane-0.14.0-1.rhaos4.10.el8.src.rpm
Repository   : rhocp-4.10-for-rhel-8-x86_64-rpms
Summary      : Butane config transpiler
URL          : https://github.com/coreos/butane
License      : ASL 2.0
Description  : 
             : Butane translates human-readable Butane Configs into machine-readable Ignition
             : configs for provisioning operating systems that use Ignition.

Name         : ignition-validate
Version      : 2.13.0
Release      : 2.rhaos4.10.el8
Architecture : x86_64
Size         : 1.4 M
Source       : ignition-2.13.0-2.rhaos4.10.el8.src.rpm
Repository   : rhocp-4.10-for-rhel-8-x86_64-rpms
Summary      : Validation tool for Ignition configs
URL          : https://github.com/coreos/ignition
License      : ASL 2.0
Description  : Ignition is a utility used to manipulate systems during the initramfs.
             : This includes partitioning disks, formatting partitions, writing files
             : (regular files, systemd units, etc.), and configuring users. On first
             : boot, Ignition reads its configuration from a source of truth (remote
             : URL, network metadata service, hypervisor bridge, etc.) and applies
             : the configuration.
             : 
             : This package contains a tool for validating Ignition configurations.
```

**Comment by cfergeau** [comment]
With the latest changes, I've been able to generate an aarch64 bundle. I'll test it on a m1 soon.

**Comment by cfergeau** [comment]
/hold

`[Use quay.io/crcont/gvisor-tap-vsock:arm which has aarch64 support](https://github.com/code-ready/snc/pull/503/commits/efe0d9467c4627987e7b72f34a4e2e954f91811f)` is aarch64 only, and would break x86_64 bundle generation. I'll file an issue with more details about this.
The other commits in this PR are working on both platforms, so we could merge all of these while we solve the gvisor-tap-vsock image issue, this way we'd be 99% there for aarch64 support.

**Comment by cfergeau** [comment]
/unhold

After rebuilding the arm64/amd64 images outside of brew, the `latest` tag now has a working gvisor-tap-vsock version.

File: libvirt-net-default.xml [hunk 3]
```diff
@@ -0,0 +1,15 @@
...
+    </nat>
+  </forward>
+  <bridge name='virbr0' stp='on' delay='0'/>
+  <ip address='192.168.122.1' netmask='255.255.255.0'>
+    <dhcp>
+      <range start='192.168.122.2' end='192.168.122.254'/>
+    </dhcp>
+  </ip>
+</network>
+
```
praveenkumar, line 15: default network is now part of `libvirt-daemon-config-network` package so better to install that along with libvirt stuff.

createdisk.sh
```diff
@@ -28,7 +28,7 @@ install_additional_packages ${VM_IP}
 
 # Add gvisor-tap-vsock
 ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
-  podman create --name=gvisor-tap-vsock --privileged --net=host -v /etc/resolv.conf:/etc/resolv.conf -it quay.io/crcont/gvisor-tap-vsock:3231aba53905468c22e394493a0debc1a6cc6392
+  podman create --name=gvisor-tap-vsock --privileged --net=host -v /etc/resolv.conf:/etc/resolv.conf -it quay.io/crcont/gvisor-tap-vsock:latest
```
praveenkumar, line 31: 👍🏼

PR #506, opened 2022-02-07
Enforce a single search registry

praveenkumar:
This PR helps user to connect docker client from the host to
podman socket in the instance. As of now even with this PR user
have to manually perform following step to test docker client.
```
$ ssh -nNT -L $(pwd)/docker.sock:/var/run/docker.sock -i /Users/prkumar/.ssh/podman-machine-default -p 64940 root@localhost
<-- different terminal -->
$ export DOCKER_HOST=unix://$(pwd)/docker.sock
$ docker ps
CONTAINER ID   IMAGE     COMMAND   CREATED   STATUS    PORTS     NAMES
```

This is taken from https://github.com/containers/podman/blob/e06631d6c22f4d5b7a62f70ccdf623379a9d5fe7/pkg/machine/ignition.go#L325-L356

### DIFF
--- a/fcos-config.yaml
+++ b/fcos-config.yaml
@@ -182,3 +182,9 @@ storage:
       contents:
         inline: |
           podman.crc.testing
+    - path: /etc/containers/registries.conf.d/999-podman-machine.conf
+      mode: 0644
+      contents:
+        inline: |
+          unqualified-search-registries=["docker.io"]
+


praveenkumar:
/hold
Till we test bundle created from this PR.

cfergeau (commented):
I'm quite confused how this all works :(
On the VM, we create a symlink from `/run/podman/podman.sock` to `/run/docker.sock` which makes sense. Then, in the VM in `/etc/profile` we set `DOCKER HOST` to `/run/podman/podman.sock` (this is what `.Host.RemoteSocket.Path` is set to on my machine). This seems redundant with the symlink creation?
And finally, we create a ssh tunnel between `docker.sock` on the host and `/var/run/docker.sock` in the VM. Why don't we create the tunnel directly between `docker.sock` and `/run/podman/podman.sock`?

praveenkumar:
@cfergeau I think following part is not hard requirement because it only there to test docker connection in the VM after ssh mostly debugging purpose and for debugging purpose it is useful.
```
    - path: /etc/profile.d/docker-host.sh
      mode: 0644
      contents:
        inline: |
          export DOCKER_HOST="unix://$(podman info -f "{{.Host.RemoteSocket.Path}}")"
```

cfergeau:
> @cfergeau I think following part is not hard requirement because it only there to test docker connection in the VM after ssh mostly debugging purpose and for debugging purpose it is useful.
> 
> ```
>     - path: /etc/profile.d/docker-host.sh
>       mode: 0644
>       contents:
>         inline: |
>           export DOCKER_HOST="unix://$(podman info -f "{{.Host.RemoteSocket.Path}}")"
> ```

But isn't this redundant with the symlink? or docker compatibility is broken if DOCKER_HOST is unset?

fcos-config.yaml
```diff
@@ -182,3 +182,19 @@ storage:
...
+    - path: /etc/profile.d/docker-host.sh
+      mode: 0644
+      contents:
+        inline: |
+          export DOCKER_HOST="unix://$(podman info -f "{{.Host.RemoteSocket.Path}}")"
+    - path: /etc/containers/registries.conf.d/999-podman-machine.conf
+      mode: 0644
+      contents:
+        inline: |
+          unqualified-search-registries=["docker.io"]
```
cfergeau, line 199: My f35 has 
```
$ cat /etc/containers/registries.conf |grep  unqualified-search-registries
unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "docker.io", "quay.io"]
```
so I don't know why this change is needed. It does not belong in this commit.

fcos-config.yaml
```diff
@@ -182,3 +182,19 @@ storage:
       contents:
         inline: |
           podman.crc.testing
+    - path: /etc/tmpfiles.d/podman-docker.conf
+      mode: 0644
+      contents:
+        inline: |
+          L+  /run/docker.sock   -    -    -     -   /run/podman/podman.sock
```
cfergeau, line 189: What if we don't have the symlink, and we use `$ ssh -nNT -L $(pwd)/docker.sock:/run/podman/podman.sock -i /Users/prkumar/.ssh/podman-machine-default -p 64940 root@localhost`?

praveenkumar: @cfergeau No, it doesn't work.
```
<-- first terminal -->
➜  ssh -nNT -L /tmp/docker.sock:/var/run/podman.sock -p 2222 crc 
Warning: Permanently added '[127.0.0.1]:2222' (ED25519) to the list of known hosts.
channel 1: open failed: connect failed: open failed
channel 1: open failed: connect failed: open failed
channel 1: open failed: connect failed: open failed
channel 1: open failed: connect failed: open failed
channel 1: open failed: connect failed: open failed

<-- different terminal -->
➜  ~ export DOCKER_HOST=unix:///tmp/docker.sock
➜  ~ docker images                             
error during connect: Get "http://%2Ftmp%2Fdocker.sock/v1.24/images/json": EOF
```

praveenkumar: But with `/var/run/docker.sock` it works as expected.
```
<-- first terminal -->
➜ ssh -nNT -L /tmp/docker.sock:/var/run/docker.sock -p 2222 crc
Warning: Permanently added '[127.0.0.1]:2222' (ED25519) to the list of known hosts.

<-- different terminal -->
➜  ~ export DOCKER_HOST=unix:///tmp/docker.sock
➜  ~ docker images
REPOSITORY                        TAG                                        IMAGE ID       CREATED        SIZE
quay.io/crcont/gvisor-tap-vsock   3231aba53905468c22e394493a0debc1a6cc6392   f42b05e6b25e   5 months ago   6.65MB
```

cfergeau (changes requested):
The commit log makes it look as if it's a docker-specific workaround, but the same happens with direct `podman-remote` use, might be better to use it in the commit log.
And I'd link to the commit which added the workaround https://github.com/containers/podman/commit/6f36a47ac2bc5dee2785043231a6e0734c9087cf rather than some source code link.

In the end, there is nothing docker-specific in this PR ;)

praveenkumar:
> In the end, there is nothing docker-specific in this PR ;)

@cfergeau in the commit log I didn't mention it docker specific, anyway I updated the commit hash instead of code snip.

cfergeau:
> > In the end, there is nothing docker-specific in this PR ;)
> 
> @cfergeau in the commit log I didn't mention it docker specific, anyway I updated the commit hash instead of code snip.

```
$ ssh -nNT -L $(pwd)/docker.sock:/var/run/docker.sock -i /Users/prkumar/.ssh/podman-machine-default -p 64940 root@localhost
<-- different terminal -->
$ export DOCKER_HOST=unix://$(pwd)/docker.sock
$ docker ps
```

This is the only example you gave, and the commit log does not say anything about podman-remote, so I assumed this is a limitation of the docker protocol. Since docker compat is one layer higher than podman-remote, I'd use podman-remote in the example to make it obvious it has nothing to do with docker per-se.

praveenkumar:
> This is the only example you gave, and the commit log does not say anything about podman-remote, so I assumed this is a limitation of the docker protocol. Since docker compat is one layer higher than podman-remote, I'd use podman-remote in the example to make it obvious it has nothing to do with docker per-se.

@cfergeau I put both docker/podman client in the example to make it more clear.

cfergeau:
> > This is the only example you gave, and the commit log does not say anything about podman-remote, so I assumed this is a limitation of the docker protocol. Since docker compat is one layer higher than podman-remote, I'd use podman-remote in the example to make it obvious it has nothing to do with docker per-se.
> 
> @cfergeau I put both docker/podman client in the example to make it more clear.

The podman example should just be `eval $(crc podman-env); podman-remote ...`, and adding a docker example in this context is in my opinion just confusing.

PR #507, opened 2022-02-07
Add arch info to metadata and bundle name

praveenkumar:


cfergeau (changes requested):
`crc-bundle-info.json.sample` needs to be updated with this new field + doc

crc-bundle-info.json.sample
```diff
@@ -7,12 +7,15 @@
...
   # - 1.3: remove of 'clusterInfo.kubeadminPasswordFile'
-  "version": "1.3",
+  # - 1.4: addition of 'arch'
+  "version": "1.4",
   # Type of this bundle content
   # Currently the only valid type is 'snc' (which stands for 'single-node-cluster')
   "type": "snc",
   # Name of the bundle
   "name": "crc_libvirt_4.6.1",
+  # Bundle arch
```
cfergeau, line 17: What are the valid values? This follows https://gist.github.com/lizkes/975ab2d1b5f9d5fdee5d3fa665bcfde6 with amd64/arm64 being used at the moment.

crc-bundle-info.json.sample
```diff
@@ -7,12 +7,15 @@
   # - 1.1: addition of 'name'
   # - 1.2: addition of 'storage.fileList'
   # - 1.3: remove of 'clusterInfo.kubeadminPasswordFile'
-  "version": "1.3",
+  # - 1.4: addition of 'arch'
+  "version": "1.4",
```
cfergeau, line 11: 1.4 was introduced before this change for `"type": "podman"`, but `crc-bundle-info.json.sample` was not updated accordingly.

crc-bundle-info.json.sample
cfergeau: With the previous question, I meant the valid values should be documented in the sample file.

crc-bundle-info.json.sample
praveenkumar: @cfergeau `type` was already there but value was only `snc` so we just changed the `type` as `podman` for podman branch so there is no key added for info json data.

crc-bundle-info.json.sample
praveenkumar: Ack

PR #508, opened 2022-02-15
Add arch info to metadata and bundle name

praveenkumar:


anjannath (approved):
This changes the name of the bundle and will need changes on the `crc` repo to consider this new naming

cfergeau:
> This changes the name of the bundle and will need changes on the `crc` repo to consider this new naming

Since this breaks compatibility with crc, I'd rather hold this until further discussion. Might be better to keep the `crc_libvirt_4.9.10.crcbundle` naming for compatibility sake, and only append the arch on non-x86_64 platforms.

praveenkumar:
>   Since this breaks compatibility with crc, I'd rather hold this until further discussion. Might be better to keep the crc_libvirt_4.9.10.crcbundle naming for compatibility sake, and only append the arch on non-x86_64 platforms.

@cfergeau we are planning to change the master with dev branch and current master will become `old-master` branch and there I don't think it will have any compatibility issue since on current beta we already went with new bundle naming for podman side.

cfergeau:
On second thought, this should not have much of an impact in part because there were no accompanying changes to crc to validate the arch before using a bundle. The bundle validation on the crc side is mostly "is the string before the `crc_bundle` extension the one we expect"? I don't think there is code to parse that string, ... so nothing should be impacted.
However, we really really need to give more thoughts to this kind of changes, double check the impact on compatibility, and document this in the commit logs.
/unhold

PR #511, opened 2022-02-23
Fix 'arch' setting in crc-bundle-info.json

cfergeau:
${yq_ARCH} is not getting expanded as expected as it is wrapped in ''


snc-library.sh
```diff
@@ -42,7 +42,7 @@ function create_json_description {
     sncGitHash=$(git describe --abbrev=4 HEAD 2>/dev/null || git rev-parse --short=4 HEAD)
     echo {} | ${JQ} '.version = "1.4"' \
             | ${JQ} '.type = "podman"' \
-            | ${JQ} '.arch = "${yq_ARCH}"' \
+            | ${JQ} ".arch = \"${yq_ARCH}\"" \
```
praveenkumar, line 45: 👍🏼

praveenkumar (approved):

PR #512, opened 2022-02-24
Add vfkit bundle generation

cfergeau:
This PR adds support for generating vfkit bundles.
I chose to keep hyperkit bundle generation if we want to have a transition phase, code would be simpler if we directly replace the hyperkit generation.
The only difference between hyperkit and vfkit is the disk image, qcow2 in one case, raw in the other.

The same changes for the `master` branch can be found at https://github.com/cfergeau/snc/tree/vfkit (there were conflicts to resolve)

I checked the generated bundle content of vfkit/podman, but did not get a chance to try to run such a bundle.
I did not have time yet to test vfkit/openshift bundle generation.

### DIFF
--- a/crc-bundle-info.json.sample
+++ b/crc-bundle-info.json.sample
@@ -10,7 +10,8 @@
   # - 1.4: addition of 'arch'
   "version": "1.4",
   # Type of this bundle content
-  # Currently the only valid type is 'snc' (which stands for 'single-node-cluster')
+  # Currently the only valid types are 'snc' (which stands for 'single-node-cluster') and 'podman'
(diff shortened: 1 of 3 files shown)


praveenkumar:
@cfergeau From the CI following is the error 
```
 In createdisk-library.sh line 146:
    generate_macos_bundle "hyperkit" $@
                                     ^-- SC2068 (error): Double quote array expansions to avoid re-splitting elements.
In createdisk-library.sh line 160:
    generate_macos_bundle "vfkit" $@
                                  ^-- SC2068 (error): Double quote array expansions to avoid re-splitting elements.
```

praveenkumar:
/close

merged with #514

PR #519, opened 2022-03-11
Cherry pick for arch 4.9 branch

praveenkumar:


anjannath (approved):

gbraad (approved):
I do not think it is necessary, but let's prepare just in case.

PR #520, opened 2022-03-14
Append arch as part of get_dest_dir

cfergeau:
get_dest_dir can return the full destination directory suffix including
the arch, this is less error-prone this way rather than expecting every
user of destDirSuffix to know yq_ARCH must be appended as well.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -12,6 +12,7 @@ function get_dest_dir {
             DEST_DIR="$(date --iso-8601)"
         fi
     fi
+    DEST_DIR="${DEST_DIR}_${yq_ARCH}"
 }
 
 function sparsify {
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -127,7 +127,7 @@ download_podman $podman_version
 get_dest_dir
 destDirSuffix="${DEST_DIR}"
 
-libvirtDestDir="crc_libvirt_${destDirSuffix}_${yq_ARCH}"
+libvirtDestDir="crc_libvirt_${destDirSuffix}"
 mkdir "$libvirtDestDir"
 
 create_qemu_image "$libvirtDestDir"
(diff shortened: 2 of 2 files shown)


createdisk.sh
```diff
@@ -125,9 +125,9 @@ download_podman $podman_version
 
 # libvirt image generation
 get_dest_dir
-destDirSuffix="${DEST_DIR}"
+destDirSuffix="${DEST_DIR}
```
praveenkumar, line 128: You forgot the `"` here

createdisk.sh
```diff
@@ -138,15 +138,15 @@ create_tarball "$libvirtDestDir"
...
+    hyperkitDestDir="crc_hyperkit_${destDirSuffix}"
     generate_hyperkit_bundle "$libvirtDestDir" "$hyperkitDestDir" "$1" "$kernel_release" "$kernel_cmd_line"
 fi
 
 # vfkit image generation
 # This must be done after the generation of libvirt image as it reuses some of
 # the content of $libvirtDestDir
 if [ -n "${SNC_GENERATE_MACOS_BUNDLE}" ]; then
-    vfkitDestDir="crc_vfkit_${destDirSuffix}_${yq_ARCH}"
+    vfkitDestDir="crc_vfkit_${destDirSuffix}"
```
praveenkumar, line 149: Even hyperV bundle need that change L#158

createdisk-library.sh
```diff
@@ -12,6 +12,7 @@ function get_dest_dir {
             DEST_DIR="$(date --iso-8601)"
         fi
     fi
+    DEST_DIR="${DEST_DIR}_${yq_ARCH}"
```
praveenkumar, line 15: 👍🏼

praveenkumar (approved):

PR #522, opened 2022-03-24
Use latest-4.10 instead of candidate-4.10

cfergeau:
candidate-4.10 seems to be gone:

$ curl https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp-dev-preview//candidate-4.10/release.txt
<html>
<body>
<p>File not found</p>
</body>
</html>

We can use latest-4.10 instead:
https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp-dev-preview/latest-4.10/

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=2067327

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -41,7 +41,7 @@ if test -n "${OPENSHIFT_VERSION-}"; then
     OPENSHIFT_RELEASE_VERSION=${OPENSHIFT_VERSION}
     echo "Using release ${OPENSHIFT_RELEASE_VERSION} from OPENSHIFT_VERSION"
 else
-    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/candidate-4.10/release.txt | sed -n 's/^ *Version: *//p')"
+    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/latest-4.10/release.txt | sed -n 's/^ *Version: *//p')"
     if test -n "${OPENSHIFT_RELEASE_VERSION}"; then
         echo "Using release ${OPENSHIFT_RELEASE_VERSION} from the latest mirror"
     else


snc.sh
```diff
@@ -41,7 +41,7 @@ if test -n "${OPENSHIFT_VERSION-}"; then
     OPENSHIFT_RELEASE_VERSION=${OPENSHIFT_VERSION}
     echo "Using release ${OPENSHIFT_RELEASE_VERSION} from OPENSHIFT_VERSION"
 else
-    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/candidate-4.10/release.txt | sed -n 's/^ *Version: *//p')"
+    OPENSHIFT_RELEASE_VERSION="$(curl -L "${MIRROR}"/latest-4.10/release.txt | sed -n 's/^ *Version: *//p')"
```
praveenkumar, line 44: @cfergeau but we are using `MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}` not the `ocp-dev-preview` L#32, I think the BZ issue is something to do with our jobs, way we setup the mirror.

praveenkumar: Also the master should now point to `4.11` because there is no nightly on 4.10 anymore.

cfergeau: `ocp-dev-preview` comes from https://github.com/openshift/release/blob/082b1aee00ff53a18da6e3856389319830d36efd/ci-operator/step-registry/openshift/e2e/gcp/crc/test/openshift-e2e-gcp-crc-test-commands.sh#L43

I don't know why we'd be using `ocp-dev-preview` rather than `ocp`

praveenkumar: > I don't know why we'd be using ocp-dev-preview rather than ocp

@cfergeau because for e2e-crc job which runs on installer master should always use the latest nightly instead using an existing release.

cfergeau: https://github.com/code-ready/snc/commit/d5338a7e89e9e9ed0e4ae664096ff2f42cdd48f5 changed from dev-preview/latest to ocp/candidate, dunno if there was ever a dev-preview/candidate.
If we already test nightlies in e2e-crc, it may be good enough to always use `latest` in `snc` and not `candidate`

praveenkumar: @cfergeau yes it was done because as soon as candidate bits land (mostly for any major release) then it is always go in the `ocp` folder instead `ocp-dev-preview` so ideally that is the time we should branch it and make master for next upcoming release. I didn't do this time, will do that and that should solve it.

praveenkumar:
we can also close this in favor of https://github.com/openshift/release/pull/27346

PR #523, opened 2022-03-25
Switch to 4.11 nightly and changed default mirror URI

praveenkumar:
This PR changes the openshift mirror uri to `ocp-dev-preview` and
using 4.11 nightly builds.

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -29,7 +29,7 @@ BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
 HTPASSWD_FILE='users.htpasswd'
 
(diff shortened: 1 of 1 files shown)


cfergeau:
I'm still not so sure it brings us much to track 4.11 very early, as opposed to letting it settle for a bit before switching and fixing the issues we may encounter with CI

cfergeau:
> @cfergeau I think last time also we went with same discussion and to my defend it is always good to use nightly for installer e2e-crc side

Yes, and we could override this in `openshift/release/ci-operator/step-registry/openshift/e2e/gcp/crc/test/openshift-e2e-gcp-crc-test-commands.sh` to achieve this testing.

> which is the purpose of our master branch to detect issue early on instead using an released version which we already started shipping.

We don't necessarily want our `snc` master branch to be broken because it's tracking a development openshift release. We want to be able to develop against it and test our PRs against it. When we decide it's time to test 4.11 and stabilize it, we can plan time appropriately for this, do the switch and fix the issues we find.

praveenkumar:
closing this in favor of https://github.com/openshift/release/pull/27346

praveenkumar:
@cfergeau I just remember one more point which make sense to have master with upstream development branch. Assume there is some changes they do as part of CVO or some patch we need to do because a new operator is added in 4.11 then only overriding `openshift/release/ci-operator/step-registry/openshift/e2e/gcp/crc/test/openshift-e2e-gcp-crc-test-commands.sh` doesn't work and our bundle creation is always going to fail. This also happened recently with 4.10 https://github.com/code-ready/snc/commit/13524606b2318c4f59bb7600e192272f6c32cd00

cfergeau:
At this point, we are shipping 4.10 bundles with crc, so we need as much CI coverage as possible for 4.10. I'm under the impression this change would stop testing 4.10 with e2e-crc, and would switch this it to 4.11 only?
It's good to keep an eye on possible 4.11 issues, but it's secondary to testing what we are currently shipping.

praveenkumar:
> At this point, we are shipping 4.10 bundles with crc, so we need as much CI coverage as possible for 4.10. I'm under the impression this change would stop testing 4.10 with e2e-crc, and would switch this it to 4.11 only?

@cfergeau Ah I think it is miscommunication then, as of now e2e-crc job is only run for openshift/master branch not on different tag which cherry-picked for released branch :( , there are test present but those are not work because our branch schema is different then what openshift have, we might need to change it and improve our job first checkout the branch and then run test.

praveenkumar:
/close

We are creating the release-4.x branches

PR #527, opened 2022-04-21
vfkit: Add missing files

cfergeau:
For some reason, the code copying id_ecdsa_crc and kubeconfig was not
moved to generate_macos_bundle which is common to hyperkit and vfkit.
This means the vfkit bundles were missing these 2 files.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -191,9 +191,6 @@ function generate_hyperkit_bundle {
     local srcDir=$1
     local destDir=$2
 
-    mkdir "$destDir"
-    cp $srcDir/kubeconfig $destDir/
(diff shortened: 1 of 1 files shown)


createdisk-library.sh
```diff
@@ -224,6 +220,9 @@ function generate_macos_bundle {
     local kernel_release=$5
     local kernel_cmd_line=$6
 
+    mkdir "$destDir"
+    cp $srcDir/kubeconfig $destDir/
+    cp $srcDir/id_ecdsa_crc $destDir/
```
praveenkumar, line 225: 👍🏼

praveenkumar (approved):

PR #529, opened 2022-05-16
Add image format for qemu-rebase

praveenkumar:
This should fix following error for new version of qemu packages.
```
 qemu-img rebase -b fedora-coreos-qemu.x86_64.qcow2 crc_podman_libvirt_3.4.1/crc-podman.qcow2
qemu-img: Could not change the backing file to 'fedora-coreos-qemu.x86_64.qcow2': backing format must be specified
```

- https://github.com/qemu/qemu/blob/master/docs/about/removed-features.rst#qemu-img-backing-file-without-format-removed-in-61

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -68,7 +68,7 @@ function create_qemu_image {
     fi
 
     sudo chown $USER:$USER -R $destDir
-    ${QEMU_IMG} rebase -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
+    ${QEMU_IMG} rebase -f qcow2 -F qcow2 -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
     ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
 
     sparsify $destDir ${VM_PREFIX}-base ${CRC_VM_NAME}.qcow2


createdisk-library.sh
```diff
@@ -68,7 +68,7 @@ function create_qemu_image {
     fi
 
     sudo chown $USER:$USER -R $destDir
-    ${QEMU_IMG} rebase -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
+    ${QEMU_IMG} rebase -F qcow2 -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
```
cfergeau, line 71: maybe we should also add `-f qcow2`?

cfergeau (approved):

gbraad (approved):

createdisk-library.sh
gbraad: @cfergeau what's the difference?

praveenkumar: It just make sure the format is `qcow2`. I added it and it is make it clear.

cfergeau: @gbraad `-F qcow2 -b ${VM_PREFIX}-base` go together, and `-f qcow2 $destDir/${VM_PREFIX}-master-0` go together. Each option specifies the format of a different argument.

cfergeau (approved):

PR #532, opened 2022-06-10
CodeReady Containers was renamed to OpenShift Local

cfergeau:
The old name still appears in the banner displayed at the top of the
OpenShift console https://console-openshift-console.apps-crc.testing

There are still two occurrences of CodeReadyContainers in snc's code:
```
snc.sh:# Add codeReadyContainer as invoker to identify it with telemeter
snc.sh:export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
```

Since this is used for installer telemetry, not sure it's wise to change it.

### DIFF
--- a/security-notice.yaml
+++ b/security-notice.yaml
@@ -3,7 +3,7 @@ kind: ConsoleNotification
 metadata:
   name: security-notice
 spec:
-  text: CodeReady Containers OpenShift cluster is for development and testing purposes. DON'T use it for production.
+  text: OpenShift Local cluster is for development and testing purposes. DON'T use it for production.
   location: BannerTop
   color: '#fff'
   backgroundColor: darkred


security-notice.yaml
```diff
@@ -3,7 +3,7 @@ kind: ConsoleNotification
 metadata:
   name: security-notice
 spec:
-  text: CodeReady Containers OpenShift cluster is for development and testing purposes. DON'T use it for production.
+  text: OpenShift Local cluster is for development and testing purposes. DON'T use it for production.
```
praveenkumar, line 6: 👍🏼

praveenkumar (approved):

cfergeau:
> Since this is used for installer telemetry, not sure it's wise to change it.

I asked to installer people, the impact would most likely be minimal if we want to change it
Potential users are
https://github.com/openshift/cluster-version-operator/blob/0d63b9bbced8c782b109855aea4cb43a4e87c083/pkg/cvo/metrics.go#L107-L111
https://github.com/openshift/telemeter/blob/76493ca3d439dc9d1ab5476a25d36c83d1947289/jsonnet/telemeter/rules.libsonnet#L59-L73
Not sure there are more.

PR #534, opened 2022-06-13
Refactor createdisk-library.sh to make it closer to the one in the podman branch

cfergeau:
This PR reworks the createdisk-library.sh code in snc's master branch.
It should not cause any functional changes.
However, this will get the master branch code closer to the one in the podman branch.
The corresponding changes for the podman branch can be seen in https://github.com/cfergeau/snc/tree/podman , I'll file a PR for these later.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -187,19 +187,6 @@ function prepare_hyperV() {
 EOF
 }
 
-function generate_hyperkit_bundle {
-    local srcDir=$1
-    local destDir=$2
-
-    generate_macos_bundle "hyperkit" "$@"
-
-    cp $srcDir/${CRC_VM_NAME}.qcow2 $destDir/
-    # not needed, we'll reuse the data added when generating the libvirt bundle
-    #add_disk_info_to_json_description "${destDir}" "${CRC_VM_NAME}.qcow2" "qcow2"
-
-    create_tarball "$destDir"
-}
-
 function generate_vfkit_bundle {
     local srcDir=$1
     local destDir=$2
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -75,13 +75,6 @@ function create_qemu_image {
 
     sparsify ${destDir} ${base} ${overlay}
 
-    # Before using the created qcow2, check if it has lazy_refcounts set to true.
-    ${QEMU_IMG} info ${destDir}/${overlay} | grep "lazy refcounts: true" 2>&1 >/dev/null
-    if [ $? -ne 0 ]; then
-        echo "${CRC_VM_NAME}.qcow2 doesn't have lazy_refcounts enabled. This is going to cause disk image corruption when using with hyperkit"
-        exit 1;
-    fi
-
     rm -fr ${destDir}/${base}
 }
 
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -218,7 +218,7 @@ function generate_macos_bundle {
     local kernel_release=$5
     local kernel_cmd_line=$6
 
-    mkdir "$destDir"
+    mkdir -p "$destDir"
     cp $srcDir/kubeconfig $destDir/
     cp $srcDir/id_ecdsa_crc $destDir/
     cp $tmpDir/vmlinuz-${kernel_release} $destDir/
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -181,6 +181,12 @@ function install_additional_packages() {
     fi
 }
 
+function prepare_cockpit() {
+    local vm_ip=$1
+
+    install_additional_packages ${vm_ip} cockpit-bridge cockpit-ws cockpit-podman
+}
+
 function prepare_hyperV() {
     local vm_ip=$1
 
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -58,29 +58,31 @@ EOF
 
 function create_qemu_image {
     local destDir=$1
+    local base=$2
+    local overlay=$3
 
-    if [ -f /var/lib/libvirt/images/${VM_PREFIX}-master-0 ]; then
-      sudo cp /var/lib/libvirt/images/${VM_PREFIX}-master-0 $destDir
-      sudo cp /var/lib/libvirt/images/${VM_PREFIX}-base $destDir
+    if [ -f /var/lib/libvirt/images/${overlay} ]; then
+      sudo cp /var/lib/libvirt/images/${overlay} ${destDir}
+      sudo cp /var/lib/libvirt/images/${base} ${destDir}
     else
-      sudo cp /var/lib/libvirt/openshift-images/${VM_PREFIX}/${VM_PREFIX}-master-0 $destDir
-      sudo cp /var/lib/libvirt/openshift-images/${VM_PREFIX}/${VM_PREFIX}-base $destDir
+      sudo cp /var/lib/libvirt/openshift-images/${VM_PREFIX}/${overlay} ${destDir}
+      sudo cp /var/lib/libvirt/openshift-images/${VM_PREFIX}/${base} ${destDir}
     fi
 
-    sudo chown $USER:$USER -R $destDir
-    ${QEMU_IMG} rebase -f qcow2 -F qcow2 -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
-    ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
+    sudo chown $USER:$USER -R ${destDir}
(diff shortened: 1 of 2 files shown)


praveenkumar:
@cfergeau Is it ready for review or you are going to make more changes on top of this?

cfergeau:
This set of changes is ready for review. More changes will be needed, but I think these ones are good enough on their own.

createdisk-library.sh
```diff
@@ -76,36 +76,13 @@ function update_json_description {
...
-    ocSize=$(du -b $destDir/oc | awk '{print $1}')
-    ocSha256Sum=$(sha256sum $destDir/oc | awk '{print $1}')
-
-    podmanSize=$(du -b $destDir/podman-remote | awk '{print $1}')
-    podmanSha256Sum=$(sha256sum $destDir/podman-remote | awk '{print $1}')
+    crc-bundle-generator node "$destDir" --internal-ip ${INTERNAL_IP} --hostname "${VM_PREFIX}-master-0"
+    crc-bundle-generator cluster-info "$destDir"  --name "${destDir}" --kubeconfig-path "kubeconfig" --ssh-key "id_ecdsa_crc"
+    crc-bundle-generator add "$destDir" "$destDir/${CRC_VM_NAME}.qcow2" --type disk-image --format qcow2
+    crc-bundle-generator add "$destDir" "$destDir/oc" --type oc
+    crc-bundle-generator add "$destDir" "$destDir/podman-remote" --type podman
```
praveenkumar, line 83: Where is this `crc-bundle-generator` function ?

README.md
```diff
@@ -70,3 +70,16 @@ Host bootstrap
...
 Please note the SNC project is “as-is” on this Github repository. At this time, it is not an offically supported Red Hat solution.
+
+## Building podman bundles
+
+It uses Fedora CoreOS stable channel to create the bundle.
+
+### How to use?
+- Clone this repo `git clone https://github.com/code-ready/snc.git`
+- `cd <directory_to_cloned_repo>`
+- `./snc.sh`
```
praveenkumar, line 81: How does this generate the podman only bundles? I don't see any flag or environment variable to create podman/openshift bundles?

createdisk-library.sh
cfergeau: ah this part is wip, was not supposed to be pushed.

README.md
cfergeau: This is only the very beginning for this, as stated in the PR description:

>  this will get the master branch code closer to the one in the podman branch.

With these changes in the master branch+similar changes in the podman branch, `createdisk-library.sh` will be identical between both branches except for the generation of the crc-metadata-info.json file (this is what the `crc-bundle-generator` are about, but not finished yet).

createdisk-library.sh
```diff
@@ -58,29 +58,31 @@ EOF
...
 
-    sudo chown $USER:$USER -R $destDir
-    ${QEMU_IMG} rebase -f qcow2 -F qcow2 -b ${VM_PREFIX}-base $destDir/${VM_PREFIX}-master-0
-    ${QEMU_IMG} commit $destDir/${VM_PREFIX}-master-0
+    sudo chown $USER:$USER -R ${destDir}
+    ${QEMU_IMG} rebase -f qcow2 -F qcow2 -b ${base} ${destDir}/${overlay}
+    ${QEMU_IMG} commit ${destDir}/${overlay}
 
-    sparsify $destDir ${VM_PREFIX}-base ${CRC_VM_NAME}.qcow2
+    sparsify ${destDir} ${base} ${overlay}
```
praveenkumar, line 63: @cfergeau here we are sparsify `base => crc.qcow2` not `base => crc-master-0` . I can see you added `mv overlay -> vm_name.qcow2` in createdisk.sh which might be not required since we source this script to `createdisk.sh` which already have `CRC_VM_NAME` variable. Or you have something else in your mind.

createdisk-library.sh
```diff
@@ -58,29 +58,31 @@ EOF
...
+    sudo chown $USER:$USER -R ${destDir}
+    ${QEMU_IMG} rebase -f qcow2 -F qcow2 -b ${base} ${destDir}/${overlay}
+    ${QEMU_IMG} commit ${destDir}/${overlay}
 
-    sparsify $destDir ${VM_PREFIX}-base ${CRC_VM_NAME}.qcow2
+    sparsify ${destDir} ${base} ${CRC_VM_NAME}.qcow2
 
     # Before using the created qcow2, check if it has lazy_refcounts set to true.
-    ${QEMU_IMG} info ${destDir}/${CRC_VM_NAME}.qcow2 | grep "lazy refcounts: true" 2>&1 >/dev/null
+    ${QEMU_IMG} info ${destDir}/${overlay} | grep "lazy refcounts: true" 2>&1 >/dev/null
```
praveenkumar: @cfergeau This shouldn't be change. we want to check the `lazy refcounts` for final disk image which is `${CRC_VM_NAME}.qcow2` as L#76 after sparsify.

createdisk-library.sh
```diff
@@ -58,29 +58,31 @@ EOF
...
     if [ $? -ne 0 ]; then
         echo "${CRC_VM_NAME}.qcow2 doesn't have lazy_refcounts enabled. This is going to cause disk image corruption when using with hyperkit"
         exit 1;
     fi
 
-    rm -fr $destDir/${VM_PREFIX}-master-0 $destDir/${VM_PREFIX}-base
+    rm -fr ${destDir}/${base}
```
praveenkumar, line 75: shouldn't be `rm -fr ${destDir}/${base} ${destDir}/${overlay}` ?

createdisk-library.sh
cfergeau: Fixed locally.

createdisk-library.sh
cfergeau: I went back to the initial commit (don't use `CRC_VM_NAME` in create_qemu_image and use a `mv` after calling it)

createdisk-library.sh
```diff
@@ -75,6 +75,8 @@ function create_qemu_image {
 
     sparsify ${destDir} ${base} ${overlay}
 
+    chmod 0644 ${destDir}/${CRC_VM_NAME}.qcow2
```
praveenkumar, line 73: @cfergeau then it should be `chmod 0644 ${destDir}/${overlay}.qcow2` because till now there is no `${CRC_VM_NAME}.qcow2`

createdisk-library.sh
cfergeau: In case someone reads this in the future :)
In the openshift branch, we indeed have 3 distinct filenames, ${base} ${overlay} and ${CRC_VM_NAME}.qcow2, and create_qemu_image deletes ${base} and $overlay} once it's done creating ${CRC_VM_NAME}.qcow2
However in the podman branch,  ${overlay} is the same as ${CRC_VM_NAME}.qcow2, so we only have 2 distinct files. At the end of create_qemu_image, we only delete ${base} in the podman branch.
So it's better to not use ${CRC_VM_NAME} here, and to keep the mv, otherwise it's harder to hav esimilar code between both branches.

createdisk-library.sh
```diff
@@ -2,17 +2,12 @@
...
-    if [ "${OPENSHIFT_VERSION-}" != "" ]; then
-        DEST_DIR=$OPENSHIFT_VERSION
-    else
-        DEST_DIR=${PULL_NUMBER}
-        if [ -z ${DEST_DIR} ]; then
-            DEST_DIR="$(date --iso-8601)"
-        fi
+function get_dest_dir_suffix {
+    local version=$1
+    DEST_DIR_SUFFIX="$version_${yq_ARCH}"
```
praveenkumar, line 7: @cfergeau this should be `DEST_DIR_SUFFIX="${version}_${yq_ARCH}"` otherwise CI fails with unbound variable `$version_`

praveenkumar (changes requested):

createdisk-library.sh
```diff
@@ -2,17 +2,12 @@
...
-    else
-        DEST_DIR=${PULL_NUMBER}
-        if [ -z ${DEST_DIR} ]; then
-            DEST_DIR="$(date --iso-8601)"
-        fi
+function get_dest_dir_suffix {
+    local version=$1
+    DEST_DIR_SUFFIX="${version}_${yq_ARCH}"
+    if [ -n "${PULL_NUMBER-}" ]; then
+         DEST_DIR_SUFFIX="$DEST_DIR_SUFFIX.pr${PULL_NUMBER}"
```
praveenkumar, line 9: 👍🏼

praveenkumar (approved):

PR #537, opened 2022-06-14
increased timeout  for first csr request to 500

adrianriobo:
Some SNC executions fails due to timeout while waiting for the first CSR request, this PR increase the timeout for waiting for the first CSR to 500

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -202,7 +202,7 @@ function renew_certificates() {
     start_vm ${vm_prefix}
 
     # After cluster starts kube-apiserver-client-kubelet signer need to be approved
-    timeout 300 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
+    timeout 500s bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
     ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
 
     # Retry 10 times to make sure kubelet certs are rotated correctly.


cfergeau:
The podman commits break 4.10 bundle generation which is what `master` is currently based on.

adrianriobo:
yeah sorry undoing... was playing to have this all together for testing

cfergeau:
> yeah sorry undoing... was playing to have this all together for testing

Ah ok, no worries, play as much as you want as long as this does not get mistakenly pushed to master :)

snc-library.sh
```diff
@@ -202,7 +202,7 @@ function renew_certificates() {
     start_vm ${vm_prefix}
 
     # After cluster starts kube-apiserver-client-kubelet signer need to be approved
-    timeout 300 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
+    timeout 500 bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
```
cfergeau, line 205: Can you change this to `500s` to avoid the need to check the manpage? Otherwise, nothing against it if this changed is required for you for a successful run. 5 minutes timeout or 8min20s timeout are not going to make a significant difference

adrianriobo:
nah that is it, had it for a long time but did not open the PR ... so now forgot the PR was pointing to my play branch...now PR keep as it is...and have a new play branch :) .... Super thx for the reminder tough!

snc-library.sh
adrianriobo: sure, will add time units to make it clear :+1:

adrianriobo: done

cfergeau (approved):

PR #541, opened 2022-07-07
Add gen-bundle-image script to create bundle image

praveenkumar:


tools.sh
```diff
@@ -5,6 +5,7 @@ JQ=${JQ:-jq}
 QEMU_IMG=${QEMU_IMG:-qemu-img}
 VIRT_FILESYSTEMS=${VIRT_FILESYSTEMS:-virt-filesystems}
 GUESTFISH=${GUESTFISH:-guestfish}
```
cfergeau, line 7: commit log says "virsh is part of different package", I'm not sure what this implies, different package from what?

README.md
```diff
@@ -48,6 +48,14 @@ export BUNDLED_PULL_SECRET_PATH="/tmp/pull_secret.json"
 ./createdisk.sh crc-tmp-install-data
 ```
 
+## Creating container image for bundles
+
+```
+./gen-bundle-image.sh <version> <openshift/okd/podman>
+```
+
+Note: gpg key is needed to create detach signature of the bundle file which added to image.
```
cfergeau, line 57: `Note: a GPG key is needed to sign the bundles before they are wrapped in a container image` ?

gen-bundle-image.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+set -exuo pipefail
+
+function set_bundle {
```
cfergeau, line 5: `set_bundle_variables`?

README.md
```diff
@@ -48,6 +48,14 @@ export BUNDLED_PULL_SECRET_PATH="/tmp/pull_secret.json"
 ./createdisk.sh crc-tmp-install-data
 ```
 
+## Creating container image for bundles
+
+```
+./gen-bundle-image.sh <version> <openshift/okd/podman>
```
cfergeau, line 54: Maybe add a bit more context?

> After running `snc.sh` / `createdisk.sh`, the generated bundles can be uploaded to a container registry using this command:

gen-bundle-image.sh
```diff
@@ -0,0 +1,80 @@
...
+FROM scratch
+COPY crc_podman_vfkit_${version}_arm64.crcbundle crc_podman_vfkit_${version}_arm64.crcbundle.sig /
+EOF
+   fi
+
+   cat <<EOF | podman build --os darwin --arch amd64 --tag ${preset}-bundle:darwin-amd64 -f - .
+FROM scratch
+COPY ${vfkit_bundle} ${vfkit_bundle}.sig /
+EOF
+
```
cfergeau, line 34: Not really enthusiastic about copying and pasting this N times. Looks like this could be parametrized relatively easily?
`generate_image "darwin" "amd64" "podman"`? `set_bundle_variables` could probably be removed.

gen-bundle-image.sh
```diff
@@ -0,0 +1,80 @@
...
+   if [[ ${preset} = "podman" ]]; then
+      podman manifest add ${preset}-bundle:${version} containers-storage:localhost/${preset}-bundle:darwin-arm64
+   fi
+   podman manifest add ${preset}-bundle:${version} containers-storage:localhost/${preset}-bundle:darwin-amd64
+   podman manifest add ${preset}-bundle:${version} containers-storage:localhost/${preset}-bundle:windows-amd64
+   podman manifest add ${preset}-bundle:${version} containers-storage:localhost/${preset}-bundle:linux-amd64
+   podman manifest inspect ${preset}-bundle:${version}
+}
+
+function generate_sign {
```
cfergeau, line 60: `sign_bundle_files`?

cfergeau (commented):
Minor comments, looks good overall.

gen-bundle-image.sh
praveenkumar: @cfergeau you mean to call generate_image for each platform and arch? Also it means we have to create bundle names as per arch and platform so it will be around same so better to have current implementation. `set_bundle_variables` also used in the `sign_bundle_files`. As of now I would differ it.

tools.sh
cfergeau: I still don't get the commit log. `libvirt-client` has been there since RHEL6, so "is now part" is kind of odd, this is not a recent change. As far as I can tell snc does not install any libvirt related packages, so isn't this change just making sure `virsh` is present before we try to use it?

gen-bundle-image.sh
cfergeau: Yes, we'd be creating the bundle names 'manually', just like we already this function already does for the aarch64 bundle: `crc_podman_vfkit_${version}_arm64.crcbundle crc_podman_vfkit_${version}_arm64.crcbundle.sig`. 
If you prefer to keep the copy & pasted Containerfile and `set_bundle_variables`, then `crc_podman_vfkit_${version}_arm64.crcbundle` needs to be moved there for consistency.

praveenkumar: Added `vfkit_bundle_arm64` for `crc_podman_vfkit_${version}_arm64.crcbundle` for consistency.

gen-bundle-image.sh
```diff
@@ -0,0 +1,81 @@
...
+   if [[ ${preset} = "podman" ]]; then
+       vfkit_bundle_arm64=crc_podman_vfkit_${version}_arm64.crcbundle
+       vfkit_bundle=crc_podman_vfkit_${version}_amd64.crcbundle
+       libvirt_bundle=crc_podman_libvirt_${version}_amd64.crcbundle
+       hyperv_bundle=crc_podman_hyperv_${version}_amd64.crcbundle
+   fi
+}
+
+function generate_image {
+   local version=$1
```
cfergeau, line 21: version is unused.

gen-bundle-image.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
```
cfergeau, line 2: This file does not seem to have executable permissions

gen-bundle-image.sh
praveenkumar: @cfergeau fixed

tools.sh
cfergeau: Also "and needs to be installed separately". But I still don't' fully get why the commit log is written tihs way :(

cfergeau (approved):
2 minor things to fix, but ACK otherwise.

PR #544, opened 2022-07-14
Remove failed pods once cluster is stable

praveenkumar:
During some of the CI run, we found out that pods which are in
Failed state doesn't removed automatic and cause the failure of snc
script. This PR removes the failed pods once cluster is stable.


anjannath (approved):

snc-library.sh
```diff
@@ -283,6 +283,7 @@ function wait_till_cluster_stable() {
        fi
        if [ $count -eq $numConsecutive ]; then
            echo "Cluster has stabilized"
+           retry ${OC} delete pods --field-selector=status.phase=Failed -A
```
cfergeau, line 286: Not 100% clear from the commit log what happens exactly. It's `retry all_pods_are_running_completed "${ignoreNamespace}"` which would fail eventually? Maybe this function should just ignore failed pods?

praveenkumar: > It's `retry all_pods_are_running_completed "${ignoreNamespace}"` which would fail eventually? 

No, this is just ignore all the pods from that namespace irrespective of the state of pod (passed/failed/Error)

> Maybe this function should just ignore failed pods?

No, We can't just ignore any failed pods (like ignoring api-server or any core-component pods result to unstable cluster) This is just removing the failed pods which are stuck in failed state (mostly of those are some component complete pods or garbage collector one).

PR #545, opened 2022-07-14
Add env variable for gpg secret key passphrase

praveenkumar:
It will allow user to provide the secret key path as env variable
instead of putting it in the same directory with `gpg_key_pass` name.

### DIFF
--- a/gen-bundle-image.sh
+++ b/gen-bundle-image.sh
@@ -2,6 +2,8 @@
 
 set -exuo pipefail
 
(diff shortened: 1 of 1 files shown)


praveenkumar:
> The `${GPG_SECRET_KEY_PASSPHRASE}` is not correctly indicating this is the path to a file

Changed it to `${GPG_SECRET_KEY_PASSPHRASE_PATH}`

gbraad (approved):

PR #547, opened 2022-07-18
Change network type to OVNKubernetes from OpenshiftSDN

praveenkumar:
OpenshiftSDN is deprecated and replace with OVNKubernetes and this
PR makes that change.
- https://docs.openshift.com/container-platform/4.10/installing/installing_sno/install-sno-preparing-to-install-sno.html

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -23,7 +23,7 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OpenShiftSDN
+  networkType: OVNKubernetes
   serviceNetwork:
   - 10.217.4.0/23
 platform:


cfergeau:
I'd prefer to stay with OpenShiftSDN for the 4.10 bundles unless there are known crc bugs which would be solved by changing it. We are already quite late in the 4.10 series (4.10.20), and this could be an impactful change. Definitely something we want in 4.11 though!

praveenkumar:
> I'd prefer to stay with OpenShiftSDN for the 4.10 bundles unless there are known crc bugs which would be solved by changing it. 

Even in 4.10 it is deprecated so even till now we are using OpenShiftSDN, we should be trying out OVNkubernetes

cfergeau:
> > I'd prefer to stay with OpenShiftSDN for the 4.10 bundles unless there are known crc bugs which would be solved by changing it.
> 
> Even in 4.10 it is deprecated so even till now we are using OpenShiftSDN, we should be trying out OVNkubernetes

Deprecated, but not gone/removed/out of support/... So it's good we are planning to move to a non-deprecated networking infra in 4.11. We can discuss making the change in 4.10, but in my opinion:
- this is a big enough change to be discussed at the team level, not just in a PR
- we need to make sure QE has enough bandwidth to carefully test this

cfergeau:
Vaguely related enhancement request https://github.com/openshift/enhancements/pull/1193/files

praveenkumar:
> * this is a big enough change to be discussed at the team level, not just in a PR
> 
> * we need to make sure QE has enough bandwidth to carefully test this

These both are discussed on Openshift level I don't know why our QE need to involve in this. Our goal is
- Does bundle creation happen with OVNKubenetes mode?
- Resource consumption shouldn't higher the OpenshiftSDN
- Our e2e should run on created bundle

If all those checks are green then we should include it.

gbraad:
I think thia should be discussed with/informed to the team. This creates awareness if problems do crop up. Changes, even tested, can cause side effects. So, no. I do think QE should have bandwidth. I do think the team should be informed.

cfergeau:
Actually, we are likely to only be releasing a few more 4.10 bundles before starting to make 4.11 releases, so I'd like to understand what makes it so urgent to change networking backend now in 4.10? What is broken and requires us to switch to OpenshiftSDN in 4.10?
I'm assuming OVNKubernetes has been marked as deprecated since the initial 4.10 release, but we just missed this?

praveenkumar:
> I'm assuming OVNKubernetes has been marked as deprecated since the initial 4.10 release, but we just missed this?

`OVNKkubernetes` => `OpenshiftSDN`. And yes it marked deprecated at the initial of 4.10 and we just missed it :(

anjannath:
Also yesterday there's an email with subject: `OVNKubernetes to be the default CNI in 4.12` in aos-devel that might be have more details, it mentions this CNI will become default only for 4.12+ ?

install-config.yaml
```diff
@@ -23,7 +23,7 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OpenShiftSDN
+  networkType: OVNKubernetes
```
cfergeau, line 26: Maybe we should not specify it, and just use what OpenShift wants to use as a default?

praveenkumar: @cfergeau we have to check if we remove this field does it auto added or result to an error.

praveenkumar:
Closing this since we decided to not to go with it in 4.10 side but have it for 4.11 branch.

install-config.yaml
cfergeau: See https://github.com/openshift/installer/blob/release-4.10/docs/user/customization.md :
```
networkType (optional string): The type of network to install. 
```

PR #550, opened 2022-07-21
Podman binfmt support

### LINKED ISSUE #549: Add binfmt support for use on Aarch64/M1
Include `qemu-user-static-x86_64` and `qemu-user-binfmt` so containers can be used from other platforms

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1088

gbraad:
Fixes #549


gbraad:
Using the wrong master

PR #551, opened 2022-07-21
Add binftmt support

### LINKED ISSUE #549: Add binfmt support for use on Aarch64/M1

gbraad:
Fixes #549

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -24,6 +24,11 @@ ${SSH} core@${VM_IP} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
 # Remove moby-engine package
 ${SSH} core@${VM_IP} -- 'sudo rpm-ostree override remove moby-engine'
 
+# Add binfmt for using other platform
+${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+  rpm-ostree install qemu-user-binfmt qemu-user-static-aarch64 qemu-user-static-x86_64
+EOF
+
 prepare_cockpit ${VM_IP}
 prepare_hyperV ${VM_IP}
 


createdisk.sh
```diff
@@ -24,6 +24,11 @@ ${SSH} core@${VM_IP} -- 'sudo find /var/log/ -iname "*.log" -exec rm -f {} \;'
 # Remove moby-engine package
 ${SSH} core@${VM_IP} -- 'sudo rpm-ostree override remove moby-engine'
 
+# Add binfmt for using other platform
+${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+  rpm-ostree install qemu-user-binfmt qemu-user-static-aarch64 qemu-user-static-x86_64
+EOF
```
cfergeau, line 30: Could be just `sudo rpm-ostree install qemu-user-binfmt qemu-user-static-aarch64 qemu-user-static-x86_64` as the line above instead adding `bash -x -s`.
Do we want to make this conditional on `ARCH` to only install the correct package? Dunno how often one will want to run aarch64 images on x86_64.

cfergeau:
Some details in the commit log why we want to do this, and a link to the issue would be useful

createdisk.sh
gbraad: It would be ideal to only include x86_64 on an Aarch64 image, and vice versa. Though the size does not increase that much

praveenkumar:
Since https://github.com/coreos/fedora-coreos-config/pull/1813 already now part of FCOS have it so we might not need this PR. This adds `qemu-user-static-x86` for aarch64 so I am putting this PR on hold until we test latest bundle.

/hold

praveenkumar:
Created 4.1.1 bundle and tested which does work as expected for M1.
```
$ podman run --arch=amd64 --rm registry.fedoraproject.org/fedora:36 arch
Trying to pull registry.fedoraproject.org/fedora:36...
[...]
x86_64
```

So I am closing this PR.

PR #552, opened 2022-07-21
install-config: Don't specify networkType

cfergeau:
We want to use the value OpenShift is using by default, and the field is
optional:
https://github.com/openshift/installer/blob/release-4.10/docs/user/customization.md
```
networkType (optional string): The type of network to install. The default is OpenShiftSDN.
```
This commit removes the setting of networkType in install-config.yaml,
this way we'll always be using the value OpenShift wants to use by
default.

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -23,7 +23,6 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OVNKubernetes
   serviceNetwork:
   - 10.217.4.0/23
 platform:


cfergeau:
/retest
(it was cert rotation failing, dunno if that's something 4.11 related, or just CI slowness)

praveenkumar:
So in Openshift installer side SingleNode means bootstrap in place [https://github.com/openshift/installer/commit/3e58797b340c1ea9890fb2e5b530ed87bba3af4a#diff-ab4fb360ac4a8344c506c58e[…]1216c118b11cc079d375R181-R186](https://github.com/openshift/installer/commit/3e58797b340c1ea9890fb2e5b530ed87bba3af4a#diff-ab4fb360ac4a8344c506c58e1b799894584bcd0daf8f1216c118b11cc079d375R181-R186) and that's the reason even with 4.11 for our snc script openshift SDN is taken as default even for us it is single node :(

cfergeau:
We can drop this PR then, and if we remember, we can reopen for 4.12. Or we can take it, and keep using OpenShift SDN for a while longer. No strong preference, even if I'd just close this for now.

praveenkumar:
> We can drop this PR then, and if we remember, we can reopen for 4.12. Or we can take it, and keep using OpenShift SDN for a while longer. No strong preference, even if I'd just close this for now.

@cfergeau yes, I am closing this right now and then we again revisit it from 4.12 once it is default.

Issue #558, opened 2022-07-28
Add preset info to bundle metadata


praveenkumar:
Currently ocp/okd doesn't have any differentiator on bundle metadata level. It is bit hard to detect the bundle mismatch on crc side. We should add the `preset` to bundle metadata which would be set as openshift/okd as per bundle type.

cfergeau:
This can be done with the existing `type` field, which crc already maps to a preset: https://github.com/code-ready/crc/blob/b91afcd751135be1d3955f1c714d70b99bd4db93/pkg/crc/machine/bundle/metadata.go#L176-L187

praveenkumar:
@cfergeau existing `type` filed is same for okd/ocp bundle. so I think better to have this `preset` filed?

cfergeau:
Depends if there's a strong technical reason for using `snc` for OKD? Or is this just something which was missed when OKD support was added? It seems simpler to just have one field for the bundle type, instead of having to look at 2 fields.

praveenkumar:
@cfergeau when we added `snc` it was for `single node cluster` and it true for OKD/OCP side. Initial plan was if we ever have multi node cluster then we should use this `type` field to distinguish. Do you want to have `snc` and `snc-okd` kind of differentiation?

cfergeau:
When podman was added, we used this field to make the distinction between podman and openshift. I would use it also for okd as it seems simpler/easier to understand/...

praveenkumar:
@cfergeau My question is should we go with `snc/snc-okd` or `openshift/okd` for it. I am tend to go with `snc-okd` so that older bundle of ocp side still works. I am open for suggestion.

cfergeau:
Yes, we would keep `snc` for openshift bundles, I don't think it's worth the complication to rename it. I would use just `okd`, but `snc-okd` is fine too.

praveenkumar:
I will just go for `okd` because I don't think now we are going to go with multinode cluster (looking at the resource consumption)

cfergeau:
even if we do, we can use `openshift-mnc` and `okd-mnc`

praveenkumar:
Updated PR with `okd` as type.

praveenkumar:
This is now part  of the bundle.

PR #559, opened 2022-07-28
Add preset info to bundle metadata

praveenkumar:


cfergeau:
Discussion regarding whether to use `preset` or `type` is ongoing in https://github.com/code-ready/snc/issues/558
/hold

createdisk.sh
```diff
@@ -123,7 +129,7 @@ download_podman $podman_version
 get_dest_dir_suffix "${OPENSHIFT_VERSION}"
 destDirSuffix="${DEST_DIR_SUFFIX}"
 
-libvirtDestDir="crc_libvirt_${destDirSuffix}"
+libvirtDestDir="${LIBVIRT_BUNDLE_PREFIX}_${destDirSuffix}"
 mkdir "$libvirtDestDir"
```
cfergeau, line 133: Could also be 
```
bundleType=""
    bundleType="_okd"
fi

libvirtDestDir="crc${bundleType}_libvirt_${destDirSuffix}"
```

This only needs one variable instead of 3. Or could be `destDirPrefix="crc_okd"` if that is more readable/understandable

praveenkumar:
/unhold

We agreed to use `type` as part of discussion on #558

anjannath (approved):

PR #560, opened 2022-07-29
Revert "Change network type to OVNKubernetes from OpenshiftSDN"

praveenkumar:
This reverts commit e4a2d40e6ded085d1d59d4357af62c21e3cce2ed.

During our testing of the bundle with OVN looks like the resource
usage is high for ovn pods than sdn. So during cluster reconcile
it break when using default mem/cpu for crc.

```
Allocated resources:
  cpu                1990m (52%)   0 (0%)
  memory             8480Mi (99%)  0 (0%)
  ephemeral-storage  0 (0%)        0 (0%)
  hugepages-1Gi      0 (0%)        0 (0%)
  hugepages-2Mi      0 (0%)        0 (0%)
```

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -23,7 +23,7 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OVNKubernetes
+  networkType: OpenShiftSDN
   serviceNetwork:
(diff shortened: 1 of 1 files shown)


adrianriobo:
The issues are solved with the revert on `4.11.0-rc.6`

PR #562, opened 2022-08-01
Arm64 bundle changes

praveenkumar:


anjannath (approved):

cfergeau:
For the record, the code for [Add function to swap kernel bits in case of aarch64](https://github.com/code-ready/snc/pull/562/commits/cabb10551bb56619be7d58354988b95f9c3fa4aa) was inspired by https://github.com/javipolo/openshift-assisted-installer-tests/blob/main/test-ocp-rhel9/kernel-replace.sh (this is where the download URL and the use of `SYSTEMD_OFFLINE` are coming from)

PR #565, opened 2022-08-10
Add cifs-utils package to hyperV bundle

anjannath:
this package provides utilities for managing mounts
of CIFS network file systems

its needed to make use of windows default  CIFS/SMB
feature for network file sharing between hosts

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -158,7 +158,7 @@ function prepare_cockpit() {
 function prepare_hyperV() {
     local vm_ip=$1
 
-    install_additional_packages ${vm_ip} hyperv-daemons
+    install_additional_packages ${vm_ip} hyperv-daemons cifs-utils
 
     # Adding Hyper-V vsock support
     ${SSH} core@${vm_ip} 'sudo bash -x -s' <<EOF


createdisk-library.sh
```diff
@@ -158,7 +158,7 @@ function prepare_cockpit() {
 function prepare_hyperV() {
     local vm_ip=$1
 
-    install_additional_packages ${vm_ip} hyperv-daemons
+    install_additional_packages ${vm_ip} hyperv-daemons cifs-utils
```
praveenkumar, line 161: 👍🏼

praveenkumar (approved):

anjannath:
I've downloaded the bundle from CI and can use it for testing, we should wait to merge this until we decide if we really want to use CIFS/SMB, see https://github.com/code-ready/crc/issues/3230#issuecomment-1211686987

gbraad:
I am not convinced CIFS is currently the priority.
I would say sshfs is more generic.

praveenkumar:
/hold

Till we finalize we should go with cifs for windows file share work.

anjannath:
I think we can go ahead with using CIFS/SMB for file sharing on windows, opened https://github.com/code-ready/crc/pull/3308

> I am not convinced CIFS is currently the priority. I would say sshfs is more generic.

the reason to choose this over `sshfs` was mainly because of performance, while doing a very basic copy test using the SMB file sharing, it had good performance as mentioned in https://github.com/code-ready/crc/issues/3230#issuecomment-1211686987 also WSL and docker uses CIFS for their file sharing

although `sshfs` is generic and will also work on macOS and linux, we currently have working file sharing with `virtiofs` on these platforms which is more performant.

gbraad:
> WSL 

WSL2 does not use CIFS from host to VM (for this they use 9p). They however use CIFS to mount VM to host using `\\wsl\`.

gbraad (approved):

createdisk-library.sh
gbraad: I am wondering if this needs to be generic for 'all' images, or just on Windows.
CIFS shares are not only usable on Windows.

For the current PR ofcourse it is, but this does not have to be the case.

WDYT?

anjannath: don't have any preference in this case, on one hand since there is a more performant `virtiofs` option available users won't need CIFS, but if they can't use virtiofs for some reason and want to use CIFS we are preventing them, but to support CIFS for all the platforms we'll need more work on `crc`

i think when we do the necessary work on `crc` to make `CIFS` available to all platforms we can include this package for all then?

cfergeau: > i think when we do the necessary work on crc to make CIFS available to all platforms we can include this package for all then?

makes sense!

PR #569, opened 2022-08-19
gen-bundle-image: Add bundle as label to have bundle file as value

praveenkumar:
After building the image using this lable we can identify which
bundle ship with container image without extracting it.

```
$ skopeo inspect --authfile /home/prkumar/pull-secret --no-tags docker://quay.io/praveenkumar/podman-bundle:4.1.1
{
    "Name": "quay.io/praveenkumar/podman-bundle",
    "Digest": "sha256:0e9ba567738f1639fcaf7c368532dea59933e2bfd5c8db4f2d9540c371c1f9b2",
    "RepoTags": [],
    "Created": "2022-08-18T11:50:10.628039178Z",
    "DockerVersion": "",
    "Labels": {
        "bundle": "crc_podman_libvirt_4.1.1_amd64.crcbundle",
        "io.buildah.version": "1.26.2"
    },
    "Architecture": "amd64",
    "Os": "linux",
    "Layers": [
        "sha256:554b666c655e35321b370defa4448ac66507655a5eeb7b321d6b2faefd890fb8"
    ],
    "Env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ]
}
```

### DIFF
--- a/gen-bundle-image.sh
+++ b/gen-bundle-image.sh
@@ -23,23 +23,23 @@ function generate_image {
    local preset=$1
 
    if [[ ${preset} = "podman" ]]; then
-       cat <<EOF | podman build --os darwin --arch arm64 --tag podman-bundle:darwin-arm64 -f - .
+       cat <<EOF | podman build --label bundle=${vfkit_bundle_arm64} --os darwin --arch arm64 --tag podman-bundle:darwin-arm64 -f - .
 FROM scratch
 COPY ${vfkit_bundle_arm64} ${vfkit_bundle_arm64}.sig /
 EOF
    fi
 
-   cat <<EOF | podman build --os darwin --arch amd64 --tag ${preset}-bundle:darwin-amd64 -f - .
+   cat <<EOF | podman build --label bundle=${vfkit_bundle} --os darwin --arch amd64 --tag ${preset}-bundle:darwin-amd64 -f - .
 FROM scratch
 COPY ${vfkit_bundle} ${vfkit_bundle}.sig /
 EOF
 
-   cat <<EOF | podman build --os windows --arch amd64 --tag ${preset}-bundle:windows-amd64 -f - .
+   cat <<EOF | podman build --label bundle=${hyperv_bundle} --os windows --arch amd64 --tag ${preset}-bundle:windows-amd64 -f - .
 FROM scratch
 COPY ${hyperv_bundle} ${hyperv_bundle}.sig /
 EOF
 
-   cat <<EOF | podman build --os linux --arch amd64 --tag ${preset}-bundle:linux-amd64 -f - .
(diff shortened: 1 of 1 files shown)


anjannath:
This'll provide a convenient way to get the full name of the included bundle, where do we plan to use it? 

also it seems we can already know what bundle the image will have based on the name:tag `podman-bundle:4.1.1` and the OS architecture?

praveenkumar:
> This'll provide a convenient way to get the full name of the included bundle, where do we plan to use it?

https://github.com/code-ready/crc/issues/3277 

> 
> also it seems we can already know what bundle the image will have based on the name:tag `podman-bundle:4.1.1` and the OS architecture?

Plan is to use the crc release for tag as part of bundle pull instead bundle version.

anjannath:
> Plan is to use the crc release for tag as part of bundle pull instead bundle version.

I am not against having this label, just wondering if the crc tag/version should be used as the tag for the image, since goal is also to decouple the bundle/image from the crc version?

praveenkumar:
> > Plan is to use the crc release for tag as part of bundle pull instead bundle version.
> 
> I am not against having this label, just wondering if the crc tag/version should be used as the tag for the image, since goal is also to decouple the bundle/image from the crc version?

No, plan is decouple crc binary with bundle version not the other way.

cfergeau:
> Plan is to use the crc release for tag as part of bundle pull instead bundle version.

Can you share the details/discussions/conclusions around this plan? I'm not sure we ever reached a final decision on this, but this is still used as a justification for some incremental patches helping to reach this goal.

praveenkumar:
> > Plan is to use the crc release for tag as part of bundle pull instead bundle version.
> 
> Can you share the details/discussions/conclusions around this plan? I'm not sure we ever reached a final decision on this, but this is still used as a justification for some incremental patches helping to reach this goal.

@cfergeau https://github.com/code-ready/crc/issues/3206#issuecomment-1154839207 (after this comments). Yes we didn't reach to final decision but it is incremental patch to reach the goal of independently use different bundle with a specific crc release.

cfergeau:
> but it is incremental patch to reach the goal of independently use different bundle with a specific crc release.

It is an incremental patch towards what is described in https://github.com/code-ready/crc/issues/3206#issuecomment-1154839207 , for which there is no final decision. Shouldn't we first have a clear idea where we are going before starting pushing patches? This would avoid some back and forth, potential reverts, ...

praveenkumar:
> > but it is incremental patch to reach the goal of independently use different bundle with a specific crc release.
> 
> It is an incremental patch towards what is described in [code-ready/crc#3206 (comment)](https://github.com/code-ready/crc/issues/3206#issuecomment-1154839207) , for which there is no final decision. Shouldn't we first have a clear idea where we are going before starting pushing patches? This would avoid some back and forth, potential reverts, ...

@cfergeau Sure, let's discuss it tomorrow team sync and plan it out. I will put hold on this PR till then.

/hold

PR #570, opened 2022-08-23
Add logic to create custom release after patching KAO/KCMO

praveenkumar:
Currently after bootstrap cert rotation, cert validity become 30 days
which is affecting crc users (who start cluster after 30 day). In this
PR we are adding patch to kube-apiserver-operator (KAO) and
kube-controller-manager-operator (KCMO) increase this limit to 1 year.

We are also creating custom release and use that to create cluster
initially.

How to test
-----------
- Update the pull secret file which should have auth for `quay.io/crcont` and `registry.ci.openshift.org` 
To get the credentials for `registry.ci.openshift.org` check https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-log-in-to-pull-images-that-require-authentication and the token would be valid for a month https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#i-have-logged-in-to-the-registry-in-the-past 
```
$ SNC_USE_PATCHED_KAO_KCMO_IMAGE=enabled snc.sh 
```

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -26,6 +26,21 @@ function download_oc() {
     fi
 }
 
+function create_new_release_with_patched_images() {
+    upstream_registry=quay.io/crcont
+    openshift_version=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} -ojsonpath='{.config.config.Labels.io\.openshift\.release}')
+    # As of now `oc adm release new` not able to parse images which have multiple arch manifest file so we first need to get the digest of the image for specific
+    # arch and then use that digest with `oc adm release new`.
+    kao_image_digest=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-apiserver-operator:${openshift_version} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+    kcmo_image_digest=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator:${openshift_version} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+
+    ${OC} adm release new -a ${OPENSHIFT_PULL_SECRET_PATH} --from-release=${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} \
+	    cluster-kube-apiserver-operator=${upstream_registry}/openshift-crc-cluster-kube-apiserver-operator@${kao_image_digest} \
+	    cluster-kube-controller-manager-operator=${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator@${kcmo_image_digest} \
+	    --to-image=quay.io/crcont/ocp-release:${openshift_version}
+    # Replace the release image override with crcont image
+    OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=quay.io/crcont/ocp-release:${openshift_version}
+}
 
 function run_preflight_checks() {
         if [ -z "${OPENSHIFT_PULL_SECRET_PATH-}" ]; then
--- a/snc.sh
+++ b/snc.sh
@@ -33,6 +33,7 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
+USE_PATCHED_RELEASE_IMAGE=${SNC_USE_PATCHED_RELEASE_IMAGE:-disabled}
 HTPASSWD_FILE='users.htpasswd'
 
 run_preflight_checks
@@ -72,6 +73,12 @@ if test -z ${OPENSHIFT_INSTALL-}; then
     OPENSHIFT_INSTALL=./openshift-baremetal-install
 fi
 
+if [[ ${USE_PATCHED_RELEASE_IMAGE} == "enabled" ]]
+then
+   echo "Creating a new image with patched KAO/KCMO images"
+   create_new_release_with_patched_images
+   echo "OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE set to ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE}"
+fi
 
 # Allow to disable debug by setting SNC_OPENSHIFT_INSTALL_NO_DEBUG in the environment
 if test -z "${SNC_OPENSHIFT_INSTALL_NO_DEBUG-}"; then
@@ -128,7 +135,7 @@ replace_pull_secret ${INSTALL_DIR}/install-config.yaml
 ${YQ} eval ".sshKey = \"$(cat id_ecdsa_crc.pub)\"" --inplace ${INSTALL_DIR}/install-config.yaml
 
 # Create the manifests using the INSTALL_DIR
-${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests
+OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests
 
 # Add CVO overrides before first start of the cluster. Objects declared in this file won't be created.
 ${YQ} eval-all --inplace 'select(fileIndex == 0) * select(filename == "cvo-overrides.yaml")' ${INSTALL_DIR}/manifests/cvo-overrides.yaml cvo-overrides.yaml
--- a/cluster-kube-apiserver-operator.patch
+++ b/cluster-kube-apiserver-operator.patch
@@ -1,0 +1,183 @@
+diff --git a/pkg/operator/certrotationcontroller/certrotationcontroller.go b/pkg/operator/certrotationcontroller/certrotationcontroller.go
+index 1bf5d3224..a28ce71ed 100644
+--- a/pkg/operator/certrotationcontroller/certrotationcontroller.go
++++ b/pkg/operator/certrotationcontroller/certrotationcontroller.go
+@@ -129,8 +129,8 @@ func newCertRotationController(
+ 		certrotation.RotatedSigningCASecret{
+ 			Namespace:              operatorclient.OperatorNamespace,
+ 			Name:                   "aggregator-client-signer",
+-			Validity:               30 * rotationDay,
+-			Refresh:                15 * rotationDay,
++			Validity:               365 * rotationDay,
++			Refresh:                180 * rotationDay,
+ 			RefreshOnlyWhenExpired: refreshOnlyWhenExpired,
+ 			Informer:               kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets(),
+ 			Lister:                 kubeInformersForNamespaces.InformersFor(operatorclient.OperatorNamespace).Core().V1().Secrets().Lister(),
+@@ -148,8 +148,8 @@ func newCertRotationController(
+ 		certrotation.RotatedSelfSignedCertKeySecret{
+ 			Namespace:              operatorclient.TargetNamespace,
+ 			Name:                   "aggregator-client",
+-			Validity:               30 * rotationDay,
+-			Refresh:                15 * rotationDay,
++			Validity:               365 * rotationDay,
++			Refresh:                180 * rotationDay,
+ 			RefreshOnlyWhenExpired: refreshOnlyWhenExpired,
+ 			CertCreator: &certrotation.ClientRotation{
+ 				UserInfo: &user.DefaultInfo{Name: "system:openshift-aggregator"},
+@@ -188,8 +188,8 @@ func newCertRotationController(
+ 		certrotation.RotatedSelfSignedCertKeySecret{
+ 			Namespace:              operatorclient.TargetNamespace,
+ 			Name:                   "kubelet-client",
+-			Validity:               30 * rotationDay,
+-			Refresh:                15 * rotationDay,
++			Validity:               365 * rotationDay,
++			Refresh:                180 * rotationDay,
+ 			RefreshOnlyWhenExpired: refreshOnlyWhenExpired,
+ 			CertCreator: &certrotation.ClientRotation{
+ 				UserInfo: &user.DefaultInfo{Name: "system:kube-apiserver", Groups: []string{"kube-master"}},
+@@ -228,8 +228,8 @@ func newCertRotationController(
+ 		certrotation.RotatedSelfSignedCertKeySecret{
+ 			Namespace:              operatorclient.TargetNamespace,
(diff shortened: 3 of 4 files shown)


praveenkumar:
As of now we are depending on the external registry which are behind auth so pull secrets need to be updated and on the internal CI we need to change it to have those auth in place. I still need to figure out how to build those component for OKD so the bundle generated for OKD also valid for an year. 

Another way to create a local registry first and let installer to use it instead pushing those patched image to quay but that again need more code (running the local registry and adding imageregistry resource to pull it from that local registry) but then `oc adm release info` might fails which always fetch details from release artifacts.

cfergeau:
Regarding building these components, since these are core OpenShift components, it would be good to be as close as possible to OpenShift when building them, ie build their images on RH infrastructure in the same way as OpenShift Release Team builds them.

praveenkumar:
> Regarding building these components, since these are core OpenShift components, it would be good to be as close as possible to OpenShift when building them, ie build their images on RH infrastructure in the same way as OpenShift Release Team builds them.

We are using https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Dockerfile.rhel7 (for KAO) and it use the images which is created on RH infra or you want to upload it to same infra because that need some conversation between team how to get push access to those registry.

snc.sh
```diff
@@ -72,6 +73,12 @@ if test -z ${OPENSHIFT_INSTALL-}; then
     OPENSHIFT_INSTALL=./openshift-baremetal-install
 fi
 
+if [[ ${CLUSTER_PATCH} == "enabled" ]]
```
gbraad, line 76: Please name more intuitively

snc.sh
```diff
@@ -72,6 +73,12 @@ if test -z ${OPENSHIFT_INSTALL-}; then
...
+then
+   echo "Patching cluster KAO and KCMO"
+   patch_operators_and_create_new_release
```
gbraad, line 79: please describe more clearly 'which operators' are patched here.

cfergeau:
> > Regarding building these components, since these are core OpenShift components, it would be good to be as close as possible to OpenShift when building them, ie build their images on RH infrastructure in the same way as OpenShift Release Team builds them.
> 
> We are using https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/Dockerfile.rhel7 (for KAO) and it use the images which is created on RH infra or you want to upload it to same infra because that need some conversation between team how to get push access to those registry.

I don't think we'll easily get access to the registries OpenShift uses, but we should try to build this on RH approved build infra and be as close as possible to how the released OpenShift components are built

snc-library.sh
```diff
@@ -25,6 +25,37 @@ function download_oc() {
     fi
 }
 
+function patch_operators_and_create_new_release() {
```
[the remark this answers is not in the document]
praveenkumar: @gbraad I did thought but the commits and image build works better in single function otherwise we need to pass argument on this generic function.

snc.sh
praveenkumar: There is `echo "Patching cluster KAO and KCMO"` or you mean not using short form for those?

cfergeau:
> @cfergeau there are 2 parts, 1. building image for KAO and KCMO 2. creating a new release using those image. Which part you want to be in RH approved infra? first, second or both?

I mostly had 1. in mind.

snc-library.sh
```diff
@@ -25,6 +25,37 @@ function download_oc() {
...
+    kao_commit_id=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} --commits ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} -o go-template='{{range .references.spec.tags}}{{if eq .name "cluster-kube-apiserver-operator"}}{{index .annotations "io.openshift.build.commit.id"}}{{end}}{{end}}')
+    git clone ${kao_source_location}
+    pushd $(basename ${kao_source_location})
+    git checkout ${kao_commit_id}
+    git apply ../cluster-kube-apiserver-operator.patch
+    podman build --authfile ${OPENSHIFT_PULL_SECRET_PATH} -t quay.io/crcont/cluster-kube-apiserver-operator:${kao_commit_id} -f Dockerfile.rhel7 .
+    podman push --authfile ${OPENSHIFT_PULL_SECRET_PATH} quay.io/crcont/cluster-kube-apiserver-operator:${kao_commit_id}
+    popd
+
+    # Clone, patch and push image for cluster-kube-controller-manager-operator
```
gbraad, line 40: lot of repetition to do the same/similar thing. this can be abstracted

snc.sh
```diff
@@ -33,6 +33,7 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
+CLUSTER_PATCH=${SNC_CLUSTER_PATCH:-disable}
```
cfergeau, line 36: disabled by default seems error prone? There's also a `disable` / `enabled` discrepancy.

snc.sh
cfergeau: `USE_PATCHED_RELEASE_IMAGE` is more descriptive, but I'm not sure how to mention cert validity in there :)

snc.sh
```diff
@@ -128,7 +135,7 @@ replace_pull_secret ${INSTALL_DIR}/install-config.yaml
 ${YQ} eval ".sshKey = \"$(cat id_ecdsa_crc.pub)\"" --inplace ${INSTALL_DIR}/install-config.yaml
 
 # Create the manifests using the INSTALL_DIR
-${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests
+OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create manifests
```
cfergeau, line 138: Is this doing the right thing when `CLUSTER_PATCH` isn't set? May be better to export `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` after we set it?

praveenkumar: yes it is still doing the correct thing even cluster is not set, to get the manifest for override image.

snc-library.sh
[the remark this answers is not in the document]
cfergeau: Looks like there could at least be a `get_image_digest` helper? `get_image_digest cluster-kube-apiserver-operator` ?

snc-library.sh
```diff
@@ -26,6 +26,27 @@ function download_oc() {
...
+
+    kcmo_image=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=cluster-kube-controller-manager-operator)
+    kcmo_vcs_ref=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${kcmo_image} -ojson | jq -r '.config.config.Labels."vcs-ref"')
+    kcmo_image_diget=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator:${kcmo_vcs_ref} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+
+    release_hash=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} --commits ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} -ojsonpath='{.digest}' |  cut -d":" -f2)
+    ${OC} adm release new -a ${OPENSHIFT_PULL_SECRET_PATH} --from-release=${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} \
+	    cluster-kube-apiserver-operator=${upstream_registry}/openshift-crc-cluster-kube-apiserver-operator@${kao_image_diget} \
+	    cluster-kube-controller-manager-operator=${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator@${kcmo_image_diget} \
+	    --to-image=quay.io/crcont/ocp-release:${release_hash}
```
cfergeau, line 46: Since we have to generate the kao/kcmo images beforehand, I wonder if we should do this work in advance as well instead of redoing it for each `snc.sh` run?

snc.sh
```diff
@@ -33,6 +33,7 @@ SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
+USE_PATCHED_KAO_KCMO_IMAGE=${SNC_USE_PATCHED_KAO_KCMO_IMAGE:-disabled}
```
cfergeau, line 36: Or just `USE_PATCHED_RELEASE_IMAGE`? with a comment explaining that this release image uses patched KAO/KCMO with 1 year certificates?

snc.sh
cfergeau: Now it's only about creating a new release without doing the patching.

snc-library.sh
praveenkumar: This should happen for arch specific so better to do it here (snc.sh) then beforehand.

cfergeau: it's not possible to generate an aarch64 release image on an amd64 machine?

praveenkumar: `oc adm release new` doesn't have any arch or `filter-by-os` options as of now it happen in 2 steps, first get the digest and then we run the `oc adm release new` command.

praveenkumar: Do you mean that using internal script we should do something like following?
- Get the release artifact of arm also https://mirror.openshift.com/pub/openshift-v4/arm64/clients/ocp/latest-4.11/release.txt using the `arm64` location
- Get the KAO/KCMO image digest on arm64 using filter-by-os
- Use that arm64 location and digest to create new release of arm64 using amd64 machine and upload to quay?

praveenkumar: >There is a discrepancy between the container images which are rebuilt once per openshift release, and the release images which are rebuilt once per snc.sh run.

@cfergeau which kind of discrepancy can happen in case release image created during each snc.sh run, I hope even we run it multiple time for a specific version of Openshift it should ended up same artifacts (if not then there is some logical wrong in our approach). I do understand your point of having it part of internal script and create the release artifacts as soon as we have patched images for KAO/KCMO but that shouldn't because of any discrepancy but convenient to use with snc script and avoid pushing same content again and again (even it push just say already existing same layer).

cfergeau: Fwiw, we can probably live with that for now, but just pointing out a potential simplification for this code.

snc-library.sh
```diff
@@ -26,6 +26,22 @@ function download_oc() {
...
+    # arch and then use that digest with `oc adm release new`.
+    kao_image_diget=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-apiserver-operator:${openshift_version} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+    kcmo_image_diget=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator:${openshift_version} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+
+    release_hash=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} --commits ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} -ojsonpath='{.digest}' |  cut -d":" -f2)
+    ${OC} adm release new -a ${OPENSHIFT_PULL_SECRET_PATH} --from-release=${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} \
+	    cluster-kube-apiserver-operator=${upstream_registry}/openshift-crc-cluster-kube-apiserver-operator@${kao_image_diget} \
+	    cluster-kube-controller-manager-operator=${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator@${kcmo_image_diget} \
+	    --to-image=quay.io/crcont/ocp-release:${release_hash}
```
cfergeau, line 41: Wondering what makes the most sense for the image tag in `--to-image=quay.io/crcont/ocp-release:${release_hash}`? Might be better to use `${openshift_version}` to have the same tagging as the patched images?

snc-library.sh
```diff
@@ -26,6 +26,22 @@ function download_oc() {
...
 
+function create_new_release_with_patched_images() {
+    upstream_registry=quay.io/crcont
+    openshift_version=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} -ojsonpath='{.config.config.Labels.io\.openshift\.release}')
+    # As of now `oc adm release new` not able to parse images which have multiple arch manifest file so we first need to get the digest of the image for specific
+    # arch and then use that digest with `oc adm release new`.
+    kao_image_diget=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-apiserver-operator:${openshift_version} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+    kcmo_image_diget=$(${OC} image info -a ${OPENSHIFT_PULL_SECRET_PATH} ${upstream_registry}/openshift-crc-cluster-kube-controller-manager-operator:${openshift_version} --filter-by-os=linux/${yq_ARCH} -ojson | jq -r .digest)
+
+    release_hash=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} --commits ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} -ojsonpath='{.digest}' |  cut -d":" -f2)
```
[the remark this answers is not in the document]
praveenkumar: yes it is digest

snc.sh
```diff
@@ -161,7 +168,7 @@ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRI
 # mask the chronyd service on the bootstrap node
 cat <<< $(${JQ} '.systemd.units += [{"mask": true, "name": "chronyd.service"}]' ${INSTALL_DIR}/bootstrap.ign) > ${INSTALL_DIR}/bootstrap.ign
 
-${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create cluster ${OPENSHIFT_INSTALL_EXTRA_ARGS} || ${OC} adm must-gather --dest-dir ${INSTALL_DIR}
+OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create cluster ${OPENSHIFT_INSTALL_EXTRA_ARGS} || ${OC} adm must-gather --dest-dir ${INSTALL_DIR}
```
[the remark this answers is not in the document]
praveenkumar: Ah yes, I have to take another look into this one.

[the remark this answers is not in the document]
praveenkumar: @cfergeau I took another look and we need to have it part of `manifests` and `ignitions` generation.

snc-library.sh
praveenkumar: @cfergeau updated.

cfergeau (approved):

PR #575, opened 2022-08-29
Gen-bundle-image: Add okd bundle support

praveenkumar:


gen-bundle-image.sh
```diff
@@ -11,6 +11,12 @@ function set_bundle_variables {
    libvirt_bundle=crc_libvirt_${version}_amd64.crcbundle
    hyperv_bundle=crc_hyperv_${version}_amd64.crcbundle
 
+   if [[ ${preset} = "okd" ]]; then
+       vfkit_bundle=crc_okd_vfkit_${version}_amd64.crcbundle
+       libvirt_bundle=crc_okd_libvirt_${version}_amd64.crcbundle
+       hyperv_bundle=crc_okd_hyperv_${version}_amd64.crcbundle
+   fi
+
```
cfergeau, line 19: Alternatively, could become something like:
```
if [[ ${preset} = "podman" ]]; then
    vfkit_bundle_arm64=crc_podman_vfkit_${version}_arm64.crcbundle
fi
if [[ ${preset} = "openshift" ]]; then
    preset=""
else
    preset="_${preset}"
fi

vfkit_bundle=crc${preset}_vfkit_${version}_amd64.crcbundle
libvirt_bundle=crc${preset}_libvirt_${version}_amd64.crcbundle
hyperv_bundle=crc${preset}_hyperv_${version}_amd64.crcbundle
```

cfergeau (approved):

gen-bundle-image.sh
praveenkumar: @cfergeau I will use this suggestion for another refactor where now we also release OCP for M1 so include vfkit arm64 bundle for ocp also.

gen-bundle-image.sh
```diff
@@ -8,16 +8,16 @@ function set_bundle_variables {
...
    if [[ ${preset} = "openshift" ]]; then
        preset=""
    else
        preset="_${preset}"
    fi
 
+   if [ ${preset} = "_podman" -o ${preset} = "" ]; then
+       vfkit_bundle_arm64=crc${preset}_vfkit_${version}_arm64.crcbundle
+   fi
+
```
cfergeau, line 20: If you keep this in the same place as the `podman` check was, this is more readable, no test on an empty preset, no need to prepend `_`

gen-bundle-image.sh
```diff
@@ -26,7 +26,7 @@ function set_bundle_variables {
 function generate_image {
    local preset=$1
 
-   if [[ ${preset} = "podman" ]]; then
+   if [ ${preset} = "podman"  -o  ${preset} = "openshift" ]; then
        cat <<EOF | podman build --os darwin --arch arm64 --tag podman-bundle:darwin-arm64 -f - .
```
cfergeau, line 30: `podman-bundle` does not look right.

ci.sh
```diff
@@ -51,7 +51,7 @@ make cross
 sudo mv out/linux-amd64/crc /usr/local/bin/
 popd
 
-crc setup
+crc setup -b crc_libvirt_*.crcbundle
 crc start --disk-size 80 -m 24000 -c 10 -p "${HOME}"/pull-secret -b crc_libvirt_*.crcbundle
```
cfergeau, line 55: maybe `crc config set bundle ...` rather than duplicating this in both `setup` and `start`?

gen-bundle-image.sh
```diff
@@ -14,6 +14,7 @@ function set_bundle_variables {
 
    if [[ ${preset} = "openshift" ]]; then
        preset=""
+       vfkit_bundle_arm64=crc_vfkit_${version}_arm64.crcbundle
```
cfergeau, line 17: I'd add this above to the `if [[ ${preset} = "podman" ]]; then` to keep related things together. Or group all these tests in a `case` statement?

ci.sh
```diff
@@ -51,8 +51,9 @@ make cross
 sudo mv out/linux-amd64/crc /usr/local/bin/
 popd
 
+crc config set bundle crc_libvirt_*.crcbundle
```
cfergeau, line 54:
```
This patch allow setup and start command to use the same bundle which created
using the PR and don't use what part of crc Makefile.
```
-> `This patch alloows the setup and start commands to usue the same bundle which was created using the PR instead of using the one hardcoded in crc Makefile`

cfergeau (approved):

cfergeau (approved):

PR #582, opened 2022-09-12
Revert "Workaround corrupted podman-remote-linux binaries"

praveenkumar:
latest release of podman 4.1.1 and 4.2.0 have correct podman-remote binary so reverting this workaround.

This reverts commit 198c4b5e5df7bb93fbe6da3902a23d1106e82670.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -276,7 +276,7 @@ function download_podman() {
     local arch=$2
 
     mkdir -p podman-remote/linux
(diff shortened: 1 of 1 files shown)


cfergeau (approved):
What matters is what podman version is shipped with latest fcos 36, since this is the version we are going to download. I assume it uses 4.1.1 or newer, but it is not crystal clear from the commit log.

praveenkumar:
> What matters is what podman version is shipped with latest fcos 36, since this is the version we are going to download. I assume it uses 4.1.1 or newer, but it is not crystal clear from the commit log.

Yes it is now 4.2.0

PR #588, opened 2022-10-06
Update kao-kcmo patch script for 4.12 nightly

praveenkumar:
As of now 4.12 is produced as engineering candidate (ec) bits and it is available ocp-dev-preview location. We made change to this patch script to adopt this change and also just create scratch builds instead tag one so that QE can able to hook it with CI.

### DIFF
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -4,14 +4,14 @@
 # 1 year certificates and then push them to quay.io/crcont. The provided pull secret should allow
 # push access to `quay.io/crcont` before providing to this script.
 #    - Since this script uses rhpkg and kinit commands, it is only tested on linux.
-#    - As of now this script works with 4.11 releases because only the `rhaos-4.11-rhel-8` branch
+#    - As of now this script works with 4.12 nightly because only the `rhaos-4.12-rhel-8` branch
 #      has been created in dist-git and tested.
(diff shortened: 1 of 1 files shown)


build-patched-kao-kcmo-images.sh
```diff
@@ -72,14 +72,15 @@ function patch_and_push_image() {
     rhpkg clone containers/crc-${image_name}
     pushd crc-${image_name}
     git remote add upstream git://pkgs.devel.redhat.com/containers/ose-${image_name}
-    # Just fetch the upstream/rhaos-4.11-rhel-8 instead of all the branches and tags from upstream
-    git fetch upstream rhaos-4.11-rhel-8 --no-tags
-    git checkout --track origin/rhaos-4.11-rhel-8
+    # Just fetch the upstream/rhaos-4.12-rhel-8 instead of all the branches and tags from upstream
+    git fetch upstream rhaos-4.12-rhel-8 --no-tags
+    git checkout --track origin/private-rhaos-4.12-rhel-8
```
cfergeau, line 77: Would be good to get this from `$version`, or to have a constant whose value is `4.12`, we would not need to change many places for each new openshift version.

cfergeau (approved):
Not sure how this will work out if we need to run this very often

build-patched-kao-kcmo-images.sh
praveenkumar: Yes, constant would be much better instead extract it from `$version` .

PR #591, opened 2022-10-19
Setup kubevirt-hostpath-provisioner

### LINKED ISSUE #444: Use hostpath provisioner to create on demand PV instead manually creating
Openshift support hostpath-provisioner https://docs.openshift.com/container-platform/4.7/virt/virtual_machines/virtual_disks/virt-configuring-local-storage-for-vms.html to use with storage class to create on demand pv/pvc which we should consume instead of creating those manually. It also means we are going to have one extra pod which is running for the provisioner.

anjannath:
this sets up dynamic provisioner for pvs backed by
hosts's filesystem

the CSI driver is part of the kubevirt project and
the manifests used are available in their repo at:
https://github.com/kubevirt/hostpath-provisioner/tree/main/deploy/csi

this also modifies the pvc manifest for the internal registry
to set the StorageClass that uses the CSI based provisioner

when creating new pvc one should use the storageClassName
"crc-csi-hostpath-provisioner"

Fixes #444

### DIFF
--- a/registry_pvc.yaml
+++ b/registry_pvc.yaml
@@ -8,5 +8,5 @@ spec:
     - ReadWriteMany
   resources:
     requests:
-      storage: 100Gi
+      storage: 20Gi
   storageClassName: crc-csi-hostpath-provisioner
--- a/registry_pvc.yaml
+++ b/registry_pvc.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: crc-image-registry-storage	
+  name: crc-image-registry-storage
   namespace: openshift-image-registry
 spec:
   accessModes:
@@ -9,6 +9,4 @@ spec:
   resources:
     requests:
       storage: 100Gi
-  selector:
-    matchLabels:
-      volume: "pv0001"
+  storageClassName: crc-csi-hostpath-provisioner
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -141,61 +141,24 @@ function create_json_description {
             | ${JQ} ".clusterInfo.appsDomain = \"apps-${CRC_VM_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
 }
 
-function generate_pv() {
-  local pvdir="${1}"
-  local name="${2}"
-cat <<EOF
-apiVersion: v1
-kind: PersistentVolume
-metadata:
-  name: ${name}
-  labels:
-    volume: ${name}
-spec:
-  capacity:
-    storage: 100Gi
-  accessModes:
-    - ReadWriteOnce
-    - ReadWriteMany
-    - ReadOnlyMany
-  hostPath:
-    path: ${pvdir}
-  persistentVolumeReclaimPolicy: Recycle
-EOF
-}
-
-function setup_pv_dirs() {
-    local dir="${1}"
-    local count="${2}"
-
-    ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
-    for pvsubdir in \$(seq -f "pv%04g" 1 ${count}); do
-        mkdir -p "${dir}/\${pvsubdir}"
-    done
-    if ! chcon -R -t svirt_sandbox_file_t "${dir}" &> /dev/null; then
-        echo "Failed to set SELinux context on ${dir}"
-    fi
-    chmod -R 770 ${dir}
-EOF
-}
 
 function create_pvs() {
-    local pvdir="${1}"
-    local count="${2}"
+    # Create hostpath-provisioner namespace
+    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/namespace.yaml
+    # Add external health-monitor, provisioner RBACs
+    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/external-provisioner-rbac.yaml -n hostpath-provisioner
+    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/external-health-monitor-rbac.yaml -n hostpath-provisioner
+    # Create CSIDriver/kubevirt.io.hostpath-provisioner resource
+    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/csi-driver-hostpath-provisioner.yaml -n hostpath-provisioner
+    # Apply SCC allowin hostpath-provisioner containers to run as root and access host network
+    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/kubevirt-hostpath-security-constraints-csi.yaml
 
-    setup_pv_dirs "${pvdir}" "${count}"
+    # Deploy csi driver components
+    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/csi-kubevirt-hostpath-provisioner.yaml -n hostpath-provisioner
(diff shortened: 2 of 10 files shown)


kubevirt-hostpath-provisioner/kubevirt-hostpath-provisioner.yaml
```diff
@@ -0,0 +1,97 @@
...
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: kubevirt-hostpath-provisioner
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubevirt.io/hostpath-provisioner
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
```
praveenkumar, line 14: This is old way as per https://github.com/kubevirt/hostpath-provisioner#deployment doc looks like it should be https://github.com/kubevirt/hostpath-provisioner-operator/blob/main/deploy/storageclass-wffc-csi.yaml like this one

snc-library.sh
```diff
@@ -141,62 +141,32 @@ function create_json_description {
...
     local dir="${1}"
-    local count="${2}"
 
     ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} 'sudo bash -x -s' <<EOF
-    for pvsubdir in \$(seq -f "pv%04g" 1 ${count}); do
-        mkdir -p "${dir}/\${pvsubdir}"
-    done
-    if ! chcon -R -t svirt_sandbox_file_t "${dir}" &> /dev/null; then
+    mkdir -p "${dir}"
+    if ! chcon -R -t container_file_t "${dir}" &> /dev/null; then
```
praveenkumar, line 149: https://github.com/kubevirt/hostpath-provisioner#deployment-in-openshift it says we don't need to relabel the directory and driver suppose to take care of it, do test it and if that is the case then we just need to create the directory not to change the label.

registry_pvc.yaml
```diff
@@ -1,14 +1,14 @@
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: crc-image-registry-storage	
+  name: crc-image-registry-storage
   namespace: openshift-image-registry
+  annotations:
+    volume.kubernetes.io/selected-node:
```
praveenkumar, line 7: This is again with legacy provision spec, have we tried the latest one?

anjannath:
about resource consumption in the cluster, with just the plain _dynamic provisioner_ the cluster doesn't seem to consume more resource, while deploying the operator it takes around ~750mb more memory
```
# CRC with the hostpath-provisioner (deployed as a provisioner)
Capacity:
  cpu:                5
  ephemeral-storage:  31970284Ki
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             13999196Ki
  pods:               250

Allocatable:
  cpu:                4800m
  ephemeral-storage:  29463813686
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             13538396Ki
  pods:               250

Allocated resources:
  cpu                2250m (46%)   0 (0%)
  memory             7392Mi (55%)  0 (0%)
```
----
```
# CRC with the hostpath-provisioner-operator (provisioner is managed by the operator)
Capacity:
  cpu:                5
  ephemeral-storage:  31970284Ki
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             13999204Ki
  pods:               250

Allocatable:
  cpu:                4800m
  ephemeral-storage:  29463813686
  hugepages-1Gi:      0
  hugepages-2Mi:      0
  memory:             13538404Ki
  pods:               250

Allocated resources:
  cpu                2300m (47%)   0 (0%)
  memory             8142Mi (61%)  0 (0%)

**Comment by anjannath** [comment]
after much hit and trial, CSIDriver deployment is working now, finally :slightly_smiling_face: the provided manifests don't work ootb

**Comment by praveenkumar** [comment]
Somehow the CI failed because it detect that `hostpath-provisioner` namespace already present. I think you should add the namespace resource to one of the yaml or create a separate one and use `oc apply` to create it so that if you run same apply multiple time it doesn't change the resource and also not exist with non-zero.

```
$ cat host.yaml 
apiVersion: v1
kind: Namespace
metadata:
  name: hostpath-provisioner

$ oc apply -f host.yaml 
namespace/hostpath-provisioner unchanged
```

**Comment by anjannath** [comment]
downstream snc job passed with `4.11.10` not sure what is the reason for consistent failure with the openshift CI job :/

**Comment by anjannath** [comment]
the image registry pvc requests volume with `capacity: 100Gi` the CSI driver needs actual disk space, for this and since the VM created is only 30GB it fails to create the pv

as this was working previously when manually creating the pv with the right size, we might have to do that for the registry operator and the csi driver will be used for user workloads

```
Events:
  Type     Reason                Age                From                                                                                            Message
  ----     ------                ----               ----                                                                                            -------
  Normal   WaitForPodScheduled   81s (x2 over 83s)  persistentvolume-controller                                                                     waiting for pod image-registry-84489d56b9-5wzhh to be scheduled
  Normal   Provisioning          15s (x7 over 78s)  kubevirt.io.hostpath-provisioner_csi-hostpathplugin-2lknn_f802a19c-3b61-4624-ae77-348077c21b0b  External provisioner is provisioning volume for claim "openshift-image-registry/crc-image-registry-storage"
  Warning  ProvisioningFailed    15s (x7 over 78s)  kubevirt.io.hostpath-provisioner_csi-hostpathplugin-2lknn_f802a19c-3b61-4624-ae77-348077c21b0b  failed to provision volume with StorageClass "crc-csi-hostpath-provisioner": created volume capacity 32212254720 less than requested capacity 107374182400
  Normal   ExternalProvisioning  6s (x7 over 78s)   persistentvolume-controller                                                                     waiting for a volume to be created, either by external provisioner "kubevirt.io.hostpath-provisioner" or manually created by system administrator
```

File: registry_pvc.yaml [hunk 4]
```diff
@@ -8,5 +8,5 @@ spec:
     - ReadWriteMany
   resources:
     requests:
-      storage: 100Gi
+      storage: 20Gi
```
praveenkumar, line 11: @anjannath After that if a user again request for 20G storage does it able to provision it or is it going to fail because 20GB already provided and only 10G remaining?

anjannath: have to check this again, i think anything with less then 30G should work

anjannath: verified it can provision new ones even though the available free space doesn't match up, looks like all the pvs the csi driver creates has capacity 30G which _i guess_ it takes from the size of the volume

```
[anjan@anath-test linux]$ ./oc get pvc
NAME            STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS                   AGE
task-pv-claim   Bound    pvc-ed92ac60-9ae9-445e-b5d4-1a15c22cc74d   30Gi       RWO            crc-csi-hostpath-provisioner   55s
[anjan@anath-test linux]$ ./oc get all -n testingns
NAME              READY   STATUS    RESTARTS   AGE
pod/task-pv-pod   1/1     Running   0          97s
[anjan@anath-test linux]$ ./oc get pvc -n openshift-image-registry
NAME                         STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS                   AGE
crc-image-registry-storage   Bound    pvc-517e0782-d71a-4bfe-8cfe-aba1941a6c19   30Gi       RWX            crc-csi-hostpath-provisioner   67m
```

Issue #16, opened 2019-05-14
Add some default pv as part of snc script


praveenkumar:
We are not creating any PV which make it difficult for some of the tools like `odo` to work correctly, we need to fix it so it should work as expected.

The way `oc cluster up` used to do it https://github.com/openshift/origin/blob/release-3.11/pkg/oc/clusterup/coreinstall/components/persistent-volumes/setup_persistent_volumes.go#L39-L91

praveenkumar:
So using below scripts I am able to create 50 pv and also tested odo which works.

```console
$ cat create_files.sh 
#/bin/bash

set -xe
function setup_pv_dir() {
  local dir="${1}"
  if [[ ! -d "${dir}" ]]; then
    sudo mkdir -p "${dir}"
  fi
  if ! sudo chcon -t svirt_sandbox_file_t "${dir}" &> /dev/null; then
    echo "Not setting SELinux content for ${dir}"
  fi
  sudo chmod 770 "${dir}"
}

basedir=/var/data
setup_pv_dir "${basedir}"

for i in $(seq 1 50); do
  setup_pv_dir "${basedir}/pv${i}"
done

$ ssh core@192.168.130.11 'bash -s' < create_files.sh

$ cat create_files.sh 
#/bin/bash

set -xe
function setup_pv_dir() {
  local dir="${1}"
  if [[ ! -d "${dir}" ]]; then
    sudo mkdir -p "${dir}"
  fi
  if ! sudo chcon -t svirt_sandbox_file_t "${dir}" &> /dev/null; then
    echo "Not setting SELinux content for ${dir}"
  fi
  sudo chmod 770 "${dir}"
}

basedir=/var/data
setup_pv_dir "${basedir}"

for i in $(seq 1 50); do
  setup_pv_dir "${basedir}/pv${i}"
done
[prkumar@prkumar-test ~]$ cat create_pv.sh 
#/bin/bash -xe
function generate_pv() {
  local basedir="${1}"
  local name="${2}"
cat <<EOF
apiVersion: v1
kind: PersistentVolume
metadata:
  name: ${name}
  labels:
    volume: ${name}
spec:
  capacity:
    storage: 100Gi
  accessModes:
    - ReadWriteOnce
    - ReadWriteMany
    - ReadOnlyMany
  hostPath:
    path: ${basedir}/${name}
  persistentVolumeReclaimPolicy: Recycle
EOF
}

function create_pv() {
  local basedir="${1}"
  local name="${2}"
[...]

PR #592, opened 2022-10-19
Initial bits to create Microshift bundle

praveenkumar:
As of now we are using internal rhel-9 repo because those are not available by any subscription. We still not able to run `createdisk.sh` as it is ( it will need more changes.)


microshift.sh
```diff
@@ -0,0 +1,90 @@
...
+rm -fr microshift
+# Clone microshift repo
+git clone https://github.com/openshift/microshift.git
+pushd microshift
+ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
+sed -i '/# customizations/,$d' scripts/image-builder/config/blueprint_v0.0.1.toml
+sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
+sed -i 's/epel-release-latest-8.noarch/epel-release-latest-9.noarch/g' scripts/image-builder/configure.sh
+sed -i 's/rhocp-4.11-for-rhel-8-\${BUILD_ARCH}-rpms/latest-RHAOS/g' scripts/image-builder/build.sh
+sed -i 's/fast-datapath-for-rhel-8-\${BUILD_ARCH}-rpms/latest-FDP/g' scripts/image-builder/build.sh
```
cfergeau, line 66: fwiw, it's a bit confusing that you say rhel9 is mandatory, but microshift from github expects rhel8
Would be useful to document which repo/packages are missing from public repositories so that we can reevaluate at a later time.

praveenkumar: @cfergeau We want to start creating bundle from RHEL-9 so that we don't have to make any custom changes for M1, as of now `rhocp-4.11-for-rhel-9-$(uname -i)-rpms` and `fast-datapath-for-rhel-9-$(uname -i)-rpms` repo not available even after subscription and we change this with internal repo where dependent packages are available and those internal repo part of this commit. Do you want this detail as part of commit message or as comment?

cfergeau: The fact that RHEL9 is a requirement because of M1 support would be useful to have in the commit log or as a comment.

> so that we don't have to make any custom changes for M1

seeing all these `sed`s lines, it seems the tradeoff for this is that we are going to be using custom changes for *all* platforms. In a way it's good because there's wider testing, but if we have rhel9-specific issues, they will impact all our platforms instead of just M1. Will we get help from upstream microshift for rhel9-related problems?

praveenkumar: As soon as 4.12 RHEL-9 dev preview available we will get required help for rhel-9 related problems. `sed` command around repo is require for RHEL-9 but changing the username to `core` is required for even rhel-8.

microshift.sh
```diff
@@ -0,0 +1,90 @@
...
+# Download the oc binary for specific OS environment
+download_oc
+OC=./openshift-clients/linux/oc
+
+# Generate a new ssh keypair for this cluster
+# Create a 521bit ECDSA Key
+rm id_ecdsa_crc* || true
+ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
+
+# Configure host (RHEL-9) with required internal repo
```
gbraad, line 41: > internal 

ouch. nothing public?

praveenkumar: @gbraad not at this moment.

snc-library.sh
```diff
@@ -53,12 +54,17 @@ function run_preflight_checks() {
...
 
         # check if libvirtd is listening on a TCP socket
-        if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
-                preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
-        fi
+	if [ ${preset} == "snc" ]; then
+            if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
+                    preflight_failure  "libvirtd is not listening for plain-text TCP connections, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
+            fi
+	fi
```
cfergeau, line 67: maybe better as `run_generic_preflight_checks()` and `run_snc_preflight_checks()`? Just suggestions, not sure which is better. Need to start moving snc to some higher level language, go or python ;)

microshift.sh
cfergeau: I'd just stick with rhel8 for now to be closer to what microshift people test daily (my understanding is that they focus on rhel8?)

microshift.sh
```diff
@@ -0,0 +1,90 @@
...
+pushd microshift
+ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
+sed -i '/# customizations/,$d' scripts/image-builder/config/blueprint_v0.0.1.toml
+sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
+sed -i 's/epel-release-latest-8.noarch/epel-release-latest-9.noarch/g' scripts/image-builder/configure.sh
+sed -i 's/rhocp-4.11-for-rhel-8-\${BUILD_ARCH}-rpms/latest-RHAOS/g' scripts/image-builder/build.sh
+sed -i 's/fast-datapath-for-rhel-8-\${BUILD_ARCH}-rpms/latest-FDP/g' scripts/image-builder/build.sh
+scripts/image-builder/configure.sh
+scripts/image-builder/cleanup.sh -full
+make rpm
```
cfergeau, line 69: Will microshift provide at some point pre-built rpm that we can reuse? Or will we have to build them ourselves?

microshift.sh
```diff
@@ -0,0 +1,90 @@
...
+packaging/rpm/make-microshift-images-rpm.sh rpm ${OPENSHIFT_PULL_SECRET_PATH} ${ARCH}:${yq_ARCH} rhel-9-${ARCH}
+CONTAINERS_RPM=$(find $(pwd -P)/_output/paack-result -name microshift-containers-\*.$(uname -i).rpm)
+scripts/image-builder/build.sh -pull_secret_file ${OPENSHIFT_PULL_SECRET_PATH} -custom_rpms $CONTAINERS_RPM -authorized_keys_file $(realpath ../id_ecdsa_crc.pub)
```
cfergeau, line 72: Do you know if this is going to be improved upstream? Feels weird/error-prone to run a command to generate microshift rpms, and then have to manually fetch the list of built rpms and provide it to another script from microshift.

microshift.sh
praveenkumar: @cfergeau as of now with 4.12 branch is locked they are going to start with RHEL-9 it is just matter of time when those internal repo become public. Also with rhel8 I need to again put a patch to upgrade the kernel to rhel-9 to make it work with M1.

microshift.sh
praveenkumar: @cfergeau Once they start shipping mostly around 4.12 timeframe, meanwhile we have to build it ourself from master.

microshift.sh
praveenkumar: @cfergeau I don't know as of now this this is the way to add container images and relative dependencies. It would be better if after GA/TP they just provide a base ISO file which have all the packages and configurations.

microshift.sh
cfergeau: Upgrading to the rhel-9 kernel for M1 has been working well enough so far, are there microshift-specific issues which prevent us from doing the same for these bundles?
I'd much prefer to first let microshift do the full switch to rhel9. Once they have switched, we also start using rhel9 if this did not happen automatically.

praveenkumar: I didn't tried to have rhel-9 kernel upgrade on RHEL-8 and with 4.12 there are already early dev-preview (internal) with RHEL-9 so openshift bits are already tested with it and works as expected. That's the reason I directly went with RHEL-9 instead RHEL-8.

cfergeau: I'd expect Microshift people to know best what works well for them, I don't understand why we need to second guess them and decide on our own that rhel9 is a better pick for testing.

microshift.sh
```diff
@@ -0,0 +1,101 @@
...
+create_json_description ${BUNDLE_TYPE}
+
+# Start the VM with generated ISO
+sudo virt-install \
+    --name ${CRC_VM_NAME} \
+    --vcpus 2 \
+    --memory 2048 \
+    --arch=${ARCH} \
+    --disk path=/var/lib/libvirt/images/${CRC_VM_NAME}.qcow2,size=31 \
+    --network network=default,model=virtio \
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau As per man page of `virt-install` doesn't have default option for `model` 

```
If  --network  is  omitted a single NIC will be created in the guest. If there is a
       bridge device in the host with a physical interface attached, that will be used for
       connectivity.  Failing  that, the virtual network called default will be used. This
       option can be specified multiple times to setup more than one NIC.

model.type or model
              Network device model as seen by the guest. Value can be any nic  model  sup‐
              ported by the hypervisor, e.g.: 'e1000', 'rtl8139', 'virtio', ...
```

microshift.sh
```diff
@@ -0,0 +1,101 @@
...
+    --os-type generic \
+    --nographics \
+    --cdrom /var/lib/libvirt/images/microshift-installer-*.iso \
+    --events on_reboot=restart \
```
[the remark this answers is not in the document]
praveenkumar: on the man page I didn't find default is mention as `restart` for this.

createdisk_microshift.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# Get the IP of the VM
+VM_IP=$(sudo virsh domifaddr ${CRC_VM_NAME} | tail -2 | head -1 | awk '{print $4}' | cut -d/ -f1)
+INTERNAL_IP="127.0.0.1"
+
+# Enable the podman.socket service for API V2
+${SSH} core@${VM_IP} -- sudo systemctl enable podman.socket
+
+# Set the hostname to api.crc.testing
+${SSH} core@${VM_IP} -- sudo hostnamectl set-hostname api.crc.testing
```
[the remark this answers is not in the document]
praveenkumar: yes, we can try to set hostname as part of kickstart file.

microshift.sh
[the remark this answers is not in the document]
cfergeau: The man page does to specify what the default value will be, but something has to pick a default for the networking. I might have checked that on a rhel box with this cmdline, this uses virtio

createdisk_microshift.sh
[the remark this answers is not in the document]
praveenkumar: @cfergeau updated.

microshift.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# Start the VM with generated ISO
+sudo virt-install \
+    --name ${CRC_VM_NAME} \
+    --vcpus 2 \
+    --memory 2048 \
+    --arch=${ARCH} \
+    --disk path=/var/lib/libvirt/images/${CRC_VM_NAME}.qcow2,size=31 \
+    --network network=default,model=virtio \
+    --os-type generic \
```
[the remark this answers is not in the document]
praveenkumar: looks like `--os-type generic` is already outdated and default it takes what is the host type so removed it.

createdisk_microshift.sh
```diff
@@ -0,0 +1,141 @@
...
+get_dest_dir_suffix "${OPENSHIFT_VERSION}"
+destDirSuffix="${DEST_DIR_SUFFIX}"
+
+libvirtDestDir="${destDirPrefix}_libvirt_${destDirSuffix}"
+mkdir "$libvirtDestDir"
+
+sudo cp /var/lib/libvirt/images/${CRC_VM_NAME}.qcow2 ${libvirtDestDir}
+sudo chown $USER:$USER -R ${libvirtDestDir}
+export LIBGUESTFS_BACKEND=direct
+virt-sparsify --in-place ${libvirtDestDir}/${CRC_VM_NAME}.qcow2
```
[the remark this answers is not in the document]
praveenkumar: created a helper function for same.

createdisk_microshift.sh
```diff
@@ -0,0 +1,141 @@
...
+#
+# This must be done after the generation of libvirt image as it reuses some of
+# the content of $libvirtDestDir
+if [ -n "${SNC_GENERATE_WINDOWS_BUNDLE}" ]; then
+    hypervDestDir="${destDirPrefix}_hyperv_${destDirSuffix}"
+    generate_hyperv_bundle "$libvirtDestDir" "$hypervDestDir"
+fi
+
+# Cleanup up vmlinux/initramfs files
+rm -fr "$INSTALL_DIR/vmlinuz*" "$INSTALL_DIR/initramfs*"
```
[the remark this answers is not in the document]
praveenkumar: Now createdisk.sh script can create bundle for microshift also.

microshift.sh
```diff
@@ -0,0 +1,112 @@
...
+[latest-RHAOS]
+name = latest-RHAOS-puddle
+baseurl = http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/plashets/4.12-el9/building/\$basearch/os/
+enabled = 1
+gpgcheck = 0
+EOF
+
+rm -fr microshift
+# Clone microshift repo
+git clone -b release-4.12 https://github.com/openshift/microshift.git
```
cfergeau, line 64: With microshift being officially released, I guess this can now be updated to follow https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.12/html/installing/microshift-embed-in-rpm-ostree or https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.12/html/installing/microshift-install-rpm more closely?

praveenkumar: @cfergeau updated as per installation doc, now we are consuming released rpm instead building one. we still need to do cloning of microshift repo to consume the image-builder scripts and kickstart files.

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+
+if grep -q -i "release 8" /etc/redhat-release
+then
+  echo "Check if system is registered"
+  # Check the subscription status and register if necessary
+  if ! sudo subscription-manager status >& /dev/null ; then
+     sudo subscription-manager register
+  fi
+else
+  echo "This script only works for RHEL-8"
```
cfergeau, line 26: Isn't this missing an `exit`?
I'd "unroll" the test, `if !running_on_rhel8; then echo "only works on RHEL-8"; exit; fi`
and then `if ! subscription-manager status; then echo "machine must be registered"; exit; fi"`

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    rm -fr $pkgDir
+}
+
+enable_repo
+configure_host
+microshift_pkg_dir=$(download_microshift_rpm)
+create_iso ${microshift_pkg_dir}
+sudo cp microshift/_output/image-builder/microshift-installer-*.iso /var/lib/libvirt/images/microshift-installer.iso
+
+OPENSHIFT_RELEASE_VERSION=$(sudo dnf list | awk "/^microshift\./ {print \$2;}")
```
cfergeau, line 95: Since the rpms are already downloaded, you can directly use `rpm -qp --qf='%{VERSION}' microshift...rpm`

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+create_json_description ${BUNDLE_TYPE}
+
+# For microshift we create a empty kubeconfig file
```
cfergeau, line 98: `an empty`

podman_changes.ks
```diff
@@ -0,0 +1,26 @@
+# This file is used as part of microshift VM creation
+# to have podman specific configuration in place which
+# we do using ignition for podman bundle
```
cfergeau, line 3: I'd keep podman specific tweaks in a separate commit

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    --network network=default,model=virtio \
+    --nographics \
+    --cdrom /var/lib/libvirt/images/microshift-installer.iso \
+    --events on_reboot=restart \
+    --autoconsole none \
+    --wait
+
+# Sleep for 3 mins because after starting the vm, kickstart file takes time to
+# install from iso
+sleep 180
```
cfergeau, line 119: Can't this be replaced with `virt-install ... --wait 3`?

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    local pkgDir=$(mktemp -p /tmp -d tmp-rpmXXX)
+    sudo dnf download --downloadonly --downloaddir ${pkgDir} microshift microshift-networking microshift-selinux >/dev/null 2>&1
+    sudo chmod 0777 ${pkgDir}
+    echo ${pkgDir}
+}
+
+function create_iso {
+    local pkgDir=$1
+    rm -fr microshift
+    # Clone microshift repo
```
cfergeau, line 73: This comment is not needed,  `git clone` is self explanatory

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    git clone -b release-4.12 https://github.com/openshift/microshift.git
+    cp podman_changes.ks microshift/
+    pushd microshift
+    sudo chmod 0755 $HOME
```
cfergeau, line 77: This is done in `configure_host`

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
```
cfergeau, line 78: This is also already done elsewhere in this script

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    sed -i '/# customizations/,$d' scripts/image-builder/config/blueprint_v0.0.1.toml
+    sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
+    sed -i '/--bootproto=dhcp/a\network  --hostname=api.crc.testing' scripts/image-builder/config/kickstart.ks.template
+    sed -i '$e cat podman_changes.ks' scripts/image-builder/config/kickstart.ks.template
+    scripts/image-builder/cleanup.sh -full
+    scripts/image-builder/build.sh -microshift_rpms ${pkgDir} -pull_secret_file ${OPENSHIFT_PULL_SECRET_PATH} -authorized_keys_file $(realpath ../id_ecdsa_crc.pub)
+    popd
+    rm -fr $pkgDir
```
cfergeau, line 86: Since $pkgDir was passed as an argument, I would not delete inside this function. If you want this function to own $pkgDir, then move `microshift_pkg_dir=$(download_microshift_rpm)` to `create_iso`

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+}
+
+enable_repo
```
cfergeau, line 89: `enable_microshift_repos` maybe? At least change `repo` to `repos`

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+      cockpit-composer cockpit-machines bash-completion podman genisoimage \
+      createrepo yum-utils selinux-policy-devel jq wget lorax
+    sudo systemctl enable osbuild-composer.socket --now
+    sudo systemctl enable cockpit.socket --now
+    sudo firewall-cmd --add-service=cockpit --permanent
+    sudo chmod 0755 $HOME
+}
+
+function enable_repo {
+    # Add respective repo to fetch the microshift and dependency packages
```
cfergeau, line 57: I don't think this comment is needed.

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+
+# Generate a new ssh keypair for this cluster
+# Create a 521bit ECDSA Key
+rm id_ecdsa_crc* || true
+ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
+
+function configure_host {
+    sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
+      cockpit-composer cockpit-machines bash-completion podman genisoimage \
+      createrepo yum-utils selinux-policy-devel jq wget lorax
```
cfergeau, line 49: Can this list be limited to what the script will use directly ? I'm guessing most of these packages will be downloaded as dependencies of other packages (`bash-completion` for example). If this is c&p'ed from some doc, please add the doc URL in some comment.

microshift.sh
```diff
@@ -0,0 +1,120 @@
...
+    sudo systemctl enable osbuild-composer.socket --now
+    sudo systemctl enable cockpit.socket --now
+    sudo firewall-cmd --add-service=cockpit --permanent
+    sudo chmod 0755 $HOME
```
cfergeau, line 53: This is quite a lot of changes to do on the host generating the bundle :( Some explanations about what needs cockpit, what needs osbuild-composer, why $HOME needs to be 0755, ... would be welcome.

microshift.sh
praveenkumar: it is installed, don't know the download location for image-builder scripts also `dnf` step was mentioned in the document before ( it is removed due to removal of offine container image section)

microshift.sh
praveenkumar: This is taken from the configure host script which is part of microshift repo. I can reference it as comment. https://github.com/openshift/microshift/blob/main/scripts/image-builder/configure.sh , this script also contain the mock package which we dont' need since we are going to use built RPMs.

microshift.sh
praveenkumar: will add comment it is mostly image builder requirements.

microshift.sh
praveenkumar: Done.

microshift.sh
praveenkumar: Done

microshift.sh
praveenkumar: Done use `-wait 5` to be safer side.

microshift.sh
cfergeau: yes, but the only reason we are adding these repos is to be able to install microshift, so `microshift repos` is not that bad of a description.

microshift.sh
cfergeau: `download_microshift_rpm` does `sudo dnf download --downloadonly --downloaddir ${pkgDir} microshift microshift-release-info microshift-networking microshift-selinux >/dev/null 2>&1`, `rpm -qp` will tell rpm to look at a locally downloaded `.rpm` file, so it should be possible to use `rpm` directly rather than slowly going through `dnf`. And we'd be sure the version matches what we are going to install.

snc-library.sh
```diff
@@ -46,6 +46,7 @@ function download_oc() {
 }
 
 function run_preflight_checks() {
+	local preset=${1:-snc}
```
cfergeau, line 49: Might be better to name this "bundle_type" as this is the terminology used in snc?

snc-library.sh
```diff
@@ -56,12 +57,17 @@ function run_preflight_checks() {
...
         echo "Checking libvirt and DNS configuration"
 
-        LIBVIRT_URI=qemu+tcp://localhost/system
+	LIBVIRT_URI=qemu:///system
+	if [ ${preset} == "snc" ]; then
+           LIBVIRT_URI=qemu+tcp://localhost/system
+	fi
 
         # check if libvirtd is listening on a TCP socket
-        if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
```
cfergeau, line 62: Imo it does not hurt to keep this even for microshift to check that we can access `qemu:///system`.

snc-library.sh
```diff
@@ -56,12 +57,17 @@ function run_preflight_checks() {
 
         echo "Checking libvirt and DNS configuration"
 
-        LIBVIRT_URI=qemu+tcp://localhost/system
+	LIBVIRT_URI=qemu:///system
+	if [ ${preset} == "snc" ]; then
```
cfergeau, line 61: We have `BUNDLE_TYPE` which can be `snc`, `okd`, `microshift`, but here we have `preset` which defaults to `snc`, so `preset == snc` covers both `snc` and `okd` `BUNDLE_TYPE`. This is *very* confusing, can this be done slightly differently?

snc-library.sh
```diff
@@ -136,7 +136,11 @@ function replace_pull_secret() {
 }
 
 function create_json_description {
-    openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
+    local preset=${1:-snc}
+    openshiftInstallerVersion=""
+    if [ ${preset} == "snc" ]; then
+        openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
+    fi
```
cfergeau, line 143: Does this mean `crc-bundle-info.json` will contain:
```
"buildInfo": {
    "openshiftInstallerVersion": ""
}
``` ?
Would be better if we could not add it at all :)

File: snc-library.sh [hunk 30]
```diff
@@ -136,7 +136,11 @@ function replace_pull_secret() {
 }
 
 function create_json_description {
-    openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
+    local preset=${1:-snc}
```
cfergeau, line 139: Same comment as in the previous commit regarding `preset` VS `BUNDLE_TYPE`.

createdisk.sh
```diff
@@ -74,7 +74,7 @@ EOF
 cat crio-wipe.service | ${SSH} core@${VM_IP} "sudo tee -a /etc/systemd/system/crio-wipe.service"
 
 # Preload routes controller
-${SSH} core@${VM_IP} -- "sudo crictl pull quay.io/crcont/routes-controller:${image_tag}"
+${SSH} core@${VM_IP} -- "sudo podman pull quay.io/crcont/routes-controller:${image_tag}"
```
cfergeau, line 96: Fwiw, the `right now` part of the commit log is a bit worrying, sounds as if it's going to change any time in the future ^^

microshift.sh
```diff
@@ -0,0 +1,128 @@
...
+    exit 1
+fi
+
+if ! grep -q -i "release 8" /etc/redhat-release
+then
+  echo "This script only works for RHEL-8"
+  exit 1
+fi
+
+echo "Check if system is registered"
```
cfergeau, line 30: Aren't these prefligth checks as well? (subscription and rhel8 system)

microshift.sh
cfergeau: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/composing_a_customized_rhel_system_image/installing-composer_composing-a-customized-rhel-system-image which you mention in the comment only has `yum install osbuild-composer composer-cli cockpit-composer bash-completion` which is a lot more reasonable imo.
The microshift documentation then has `sudo yum install -y yum-utils`
The script uses `git` directly, `git-core` is probably good enough though.

microshift.sh
```diff
@@ -0,0 +1,128 @@
...
+# This requirement is taken from https://github.com/openshift/microshift/blob/main/scripts/image-builder/configure.sh
+# Also https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/composing_a_customized_rhel_system_image/installing-composer_composing-a-customized-rhel-system-image
+# list out the dependencies and usecase.
+function configure_host {
+    sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
+      cockpit-composer cockpit-machines bash-completion podman genisoimage \
+      createrepo yum-utils selinux-policy-devel jq wget lorax
+    sudo systemctl enable osbuild-composer.socket --now
+    sudo systemctl enable cockpit.socket --now
+    sudo firewall-cmd --add-service=cockpit --permanent
```
cfergeau, line 63: Maybe don't make this `--permanent`?

microshift.sh
cfergeau: I don't see this chmod 0755 $HOME in the various microshift documentation, did anything bad happen without it?

microshift.sh
```diff
@@ -0,0 +1,128 @@
...
+       --enable rhocp-4.12-for-rhel-8-$(uname -i)-rpms \
+       --enable fast-datapath-for-rhel-8-$(uname -i)-rpms
+}
+
+function download_microshift_rpm {
+    local pkgDir=$1
+    pushd ${pkgDir}
+    brew download-build --quiet --arch noarch --arch ${ARCH} ${MICROSHIFT_NVR}
+    popd
+    sudo chmod 0777 ${pkgDir}
```
cfergeau, line 78: `sudo` and `0777`? this is a temp dir the script just created, really unclear why this is needed.

microshift.sh
```diff
@@ -0,0 +1,128 @@
...
+}
+
+enable_repos
+configure_host
+microshift_pkg_dir=$(mktemp -p /tmp -d tmp-rpmXXX)
+download_microshift_rpm ${microshift_pkg_dir}
+create_iso ${microshift_pkg_dir}
+OPENSHIFT_RELEASE_VERSION=$(rpm -qp  --qf '%{VERSION}' ${microshift_pkg_dir}/microshift-4.*.rpm)
+rm -fr ${microshift_pkg_dir}
+sudo cp microshift/_output/image-builder/microshift-installer-*.iso /var/lib/libvirt/images/microshift-installer.iso
```
cfergeau, line 109: `cp -Z` might be more correct.

microshift.sh
```diff
@@ -0,0 +1,128 @@
...
+
+source tools.sh
+source snc-library.sh
+
+BUNDLE_TYPE="microshift"
+INSTALL_DIR=crc-tmp-install-data
+CRC_VM_NAME=${CRC_VM_NAME:-crc}
+BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+OPENSHIFT_RELEASE_VERSION=4.12.0
```
cfergeau, line 16: Is 4.12.0 still correct? This probably needs to be built from `MICROSHIFT_NVR`, or `MICROSHIFT_NVR` should be `${OPENSHIFT_RELEASE_VERSION}-${MICROSHIFT_RELEASE_SUFFIX}`?

cfergeau (commented):
Haven't finished reviewing yet, but the PR needs rebase and I'm worried I will lose my work so far if the rebase happens before I post the review.

microshift.sh
cfergeau: It's not part of the doc ( https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.12/html/installing/microshift-embed-in-rpm-ostree or such), if it's required it should be in the doc, not in an internal script.

microshift.sh
cfergeau: Well, these could still be a `microshift_preflight` function instead of cluttering the main code path.

microshift.sh
praveenkumar: @cfergeau I know, what I am saying is we don't need `OPENSHIFT_RELEASE_VERSION` at line #16 and later in the script we are getting it from NVR which is good enough for this script.

snc-library.sh
praveenkumar: I already added this logic.

microshift.sh
```diff
@@ -0,0 +1,134 @@
...
+microshift_pkg_dir=$(mktemp -p /tmp -d tmp-rpmXXX)
+# This directory contains the microshift rpm  passed to osbuilder, worker for osbuilder
+# running as non-priviledged user and this tmp directory have 0700 permission. To allow
+# worker to read/execute this file we need to change the permission to 0755
+chmod 0755 ${microshift_pkg_dir}
+download_microshift_rpm ${microshift_pkg_dir}
+create_iso ${microshift_pkg_dir}
+OPENSHIFT_RELEASE_VERSION=$(rpm -qp  --qf '%{VERSION}' ${microshift_pkg_dir}/microshift-4.*.rpm)
+rm -fr ${microshift_pkg_dir}
+sudo cp -Z microshift/_output/image-builder/microshift-installer-*.iso /var/lib/libvirt/images/microshift-installer.iso
```
cfergeau, line 111: If you move this right after `create_iso`, things are grouped in more natural order, we create the iso, copy it, we get the microshift version and use it to download `oc`.

snc-library.sh
```diff
@@ -46,7 +46,7 @@ function download_oc() {
 }
 
 function run_preflight_checks() {
-	local preset=${1:-snc}
+	local bundle_type=$1
```
cfergeau, line 49: This can be squashed in the previous commit.

microshift.sh
cfergeau: Still needs to be cleaned up imo, `sudo dnf install osbuild-composer composer-cli cockpit-composer bash-completion yum-utils lorax git-core  jq` and may `podman`. Some of the packages you listed seem redundant.

microshift.sh
```diff
@@ -0,0 +1,134 @@
...
+
+# This requirement is taken from https://github.com/openshift/microshift/blob/main/scripts/image-builder/configure.sh
+# Also https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/composing_a_customized_rhel_system_image/installing-composer_composing-a-customized-rhel-system-image
+# list out the dependencies and usecase.
+function configure_host {
+    sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
+      cockpit-composer cockpit-machines bash-completion podman genisoimage \
+      createrepo yum-utils selinux-policy-devel jq wget lorax
+    sudo systemctl enable osbuild-composer.socket --now
+    sudo systemctl enable cockpit.socket --now
```
cfergeau, line 58: Now I'm wondering if this should just be `systemctl start ...` rather than permanently enabling.

microshift.sh
```diff
@@ -0,0 +1,134 @@
+#!/bin/bash
+
+set -exuo pipefail
```
cfergeau, line 3: The commit log for ` Add microshift.sh script to create VM with microshift packages ` could add a high level description of how this all works for future reference.

createdisk-library.sh
```diff
@@ -95,7 +96,7 @@ function update_json_description {
         | ${JQ} '.clusterInfo.kubeConfig = "kubeconfig"' \
         | ${JQ} '.nodes[0].kind[0] = "master"' \
         | ${JQ} '.nodes[0].kind[1] = "worker"' \
-        | ${JQ} ".nodes[0].hostname = \"${VM_PREFIX}-master-0\"" \
+        | ${JQ} ".nodes[0].hostname = \"${vm_name}\"" \
```
cfergeau, line 99: I'd use `${VM_NAME}` directly rather than passing this as an argument.

createdisk.sh
```diff
@@ -15,36 +15,57 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
...
 
+if [ ${BUNDLE_TYPE} == "microshift" ]; then
+    # Pull openshift release images
+    ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+     jq --raw-output '.images | to_entries | map(.value) | join("\n")' /usr/share/microshift/release/release-$(uname -i).json | xargs -n1 podman pull --authfile /etc/crio/openshift-pull-secret
+EOF
+    # Extend the lvm to use all free space
+    ${SSH} core@${VM_IP} -- sudo lvextend -r -l +100%FREE /dev/mapper/rhel-root
+    # Disable firewalld
+    ${SSH} core@${VM_IP} -- sudo systemctl disable firewalld
```
cfergeau, line 66: Why? Since you added a comment before the command, it's a good opportunity to record why we don't want firewalld.

createdisk.sh
```diff
@@ -15,36 +15,57 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
...
 
 # Enable the podman.socket service for API V2
 ${SSH} core@${VM_IP} -- sudo systemctl enable podman.socket
 
+if [ ${BUNDLE_TYPE} == "microshift" ]; then
+    # Pull openshift release images
+    ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+     jq --raw-output '.images | to_entries | map(.value) | join("\n")' /usr/share/microshift/release/release-$(uname -i).json | xargs -n1 podman pull --authfile /etc/crio/openshift-pull-secret
+EOF
+    # Extend the lvm to use all free space
```
cfergeau, line 63: Why?

createdisk.sh
```diff
@@ -15,36 +15,57 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
...
+    
+    # Stop the kubelet service so it will not reprovision the pods
+    ${SSH} core@${VM_IP} -- sudo systemctl stop kubelet
+fi
 
 # Enable the podman.socket service for API V2
 ${SSH} core@${VM_IP} -- sudo systemctl enable podman.socket
 
+if [ ${BUNDLE_TYPE} == "microshift" ]; then
+    # Pull openshift release images
```
cfergeau, line 59: Why aren't these images pulled already?

createdisk-library.sh
```diff
@@ -383,3 +383,12 @@ function download_podman() {
       mv podman-remote/windows/podman-${version}/usr/bin/podman.exe  podman-remote/windows
     fi
 }
+
+function create_sparsify_qemu_image {
```
cfergeau, line 387: Do you mean`create_sparse_qemu_image`?

createdisk.sh
```diff
@@ -136,8 +159,12 @@ libvirtDestDir="${destDirPrefix}_libvirt_${destDirSuffix}"
 rm -fr ${libvirtDestDir} ${libvirtDestDir}.crcbundle
 mkdir "$libvirtDestDir"
 
-create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
-mv "${libvirtDestDir}/${VM_NAME}" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
+if [ $BUNDLE_TYPE != "microshift" ]; then
+    create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
+    mv "${libvirtDestDir}/${VM_NAME}" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
+else
+    create_sparsify_qemu_image "${libvirtDestDir}"
```
cfergeau, line 166: `create_qemu_image` already creates a sparse qemu image, so it's quite confusing what's being done here.

microshift.sh
```diff
@@ -0,0 +1,134 @@
...
+# Download the oc binary for specific OS environment
+OC=./openshift-clients/linux/oc
+download_oc
+
+create_json_description ${BUNDLE_TYPE}
+
+# For microshift we create an empty kubeconfig file
+# to have it as part of bundle
+mkdir -p ${INSTALL_DIR}/auth
+touch ${INSTALL_DIR}/auth/kubeconfig
```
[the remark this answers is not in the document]
praveenkumar: This kubeconfig file is generated when microshift service is started, as part of snc we dont' start the microshift service because then we have to clean the state so that's why here we use empty kubeconfig file.

microshift.sh
```diff
@@ -0,0 +1,134 @@
...
+    local pkgDir=$1
+    pushd ${pkgDir}
+    brew download-build --quiet --arch noarch --arch ${ARCH} ${MICROSHIFT_NVR}
+    popd
+}
+
+function create_iso {
+    local pkgDir=$1
+    rm -fr microshift
+    git clone -b release-4.12 https://github.com/openshift/microshift.git
```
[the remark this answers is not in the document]
praveenkumar: may be a follow up issue, not blocker and we are using the `release-4.12` branch to make sure we are not using something latest.

microshift.sh
```diff
@@ -0,0 +1,134 @@
...
+    sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
+    sed -i '/--bootproto=dhcp/a\network  --hostname=api.crc.testing' scripts/image-builder/config/kickstart.ks.template
+    scripts/image-builder/cleanup.sh -full
+    scripts/image-builder/build.sh -microshift_rpms ${pkgDir} -pull_secret_file ${OPENSHIFT_PULL_SECRET_PATH} -authorized_keys_file $(realpath ../id_ecdsa_crc.pub)
+    popd
+}
+
+enable_repos
+configure_host
+microshift_pkg_dir=$(mktemp -p /tmp -d tmp-rpmXXX)
```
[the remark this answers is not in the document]
praveenkumar: No since we are not using `build.sh` so `_output/rpmbuild` directory is not created and this is just downloading the rpm file doesn't matter the directory since we remove it at the end.

createdisk-library.sh
praveenkumar: Again not blocker, something can be take care as part of refactor.

createdisk.sh
praveenkumar: We are not running the microshift service so the images are not pulling automatic. We don't want to run the microshift service as part of snc since it takes only 1 mins once the images are present and can be done on crc side and each time user have fresh cluster instead something already provisioned and can make changes.

createdisk.sh
praveenkumar: the one we use have different use since there we commit a backup image to master and then sparsify it but here we don't do any of that and just need to sparsify it.

praveenkumar:
@cfergeau All the suggestions in latest review either can be done as follow up or if you already tested some of those suggestion then I don't mind to have a commit on top of it. But turn around time for each changes is way too much, I do need to create the bundle and make sure it works ( which takes ~2h) and looking at the priority I think we can have some of those in later sprint. Everything need not to be correct in first go. With current state I generated the bundle which I am testing (linux it works).

cfergeau:
> @cfergeau All the suggestions in latest review either can be done as follow up or if you already tested some of those suggestion then I don't mind to have a commit on top of it. But turn around time for each changes is way too much, I do need to create the bundle and make sure it works ( which takes ~2h) and looking at the priority I think we can have some of those in later sprint. Everything need not to be correct in first go. With current state I generated the bundle which I am testing (linux it works).

The lack of explanations for many important details is also hindering the PR review, turn around would be a lot faster if there was less magic in it. Since this will hinder future maintainance, I don't think "this is good enough to generate a bundle" is well good enough.
/hold  from me.

cfergeau (changes requested):
Comments for the non-obvious steps need to be improved.

createdisk-library.sh
```diff
@@ -383,3 +383,18 @@ function download_podman() {
...
     fi
 }
+
+# As of now sparsify helper is very specific to OCP/OKD kind of bundle where we get the
+# partition for the root label and then mount it with guestfish to cleanup /boot. With
+# microshift vm we are using lvm and guestfish error out during mount
+# mount /dev/sda3 /
+# libguestfs: error: mount: mount exited with status 32: mount: /sysroot: unknown filesystem type 'LVM2_member'
+# There might be other way for guestfish to mount lvm but as of now using a seperate helper is easy.
+function create_sparse_qemu_image {
```
cfergeau, line 393: Maybe name this `sparsify_lvm` if it's the main reason for it. I'd move it to `sparsify` though, and use one codepath or the other depending on `BUNDLE_TYPE`. This way all the code is in a single place.

createdisk.sh
```diff
@@ -15,36 +15,64 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
...
+    # don't run microshift service which fetch these image but instead service is run
+    # as part of crc so user have a fresh cluster instead something already provisioned
+    # but images we cache it as part of bundle.
+    ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+     jq --raw-output '.images | to_entries | map(.value) | join("\n")' /usr/share/microshift/release/release-$(uname -i).json | xargs -n1 podman pull --authfile /etc/crio/openshift-pull-secret
+EOF
+    # Extend the lvm to use all free space
+    # As part of kickstart file it only uses 8GB and we create our VM using 30GB so
+    # free space need to be extended. Instead of patching kickstart file as part of
+    ${SSH} core@${VM_IP} -- sudo lvextend -r -l +100%FREE /dev/mapper/rhel-root
```
cfergeau, line 69: You could use the `--lvm_sysroot_size` argument to `build.sh` instead of extending the volume later.
https://github.com/openshift/microshift/blob/d85cb5043b82d8f5a865a6afc1b0cf082e73e160/scripts/image-builder/build.sh#L50-L52

However, it seems we need to leave some free space at the end of our image for data stored by the CSI driver?
https://github.com/openshift/microshift/blob/d85cb5043b82d8f5a865a6afc1b0cf082e73e160/scripts/image-builder/config/kickstart.ks.template#L10-L20

createdisk.sh
cfergeau: Can you move this to the commit adding the podman changes then?

createdisk.sh
praveenkumar: > However, it seems we need to leave some free space at the end of our image for data stored by the CSI driver?

This is something I am testing with CSI driver if we need to keep some space always available for dynamic PV and how much.

createdisk-library.sh
cfergeau: This makes the changes simpler, if you agree with the change, not much point in adding complexity now in order to remove it later on.

createdisk.sh
praveenkumar: @cfergeau looks like to create the `pvc` storage driver get the requested resource from free space so now I am using 15G for `/sysroot` ( `-lvm_sysroot_size 15320` ) and remaining it is used for pvs by storage driver (CSI).

createdisk.sh
cfergeau: I think you updated the comment to explain why it's not podman specific?

createdisk.sh
```diff
@@ -15,36 +15,61 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
...
+    # don't run microshift service which fetch these image but instead service is run
+    # as part of crc so user have a fresh cluster instead something already provisioned
+    # but images we cache it as part of bundle.
+    ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+     jq --raw-output '.images | to_entries | map(.value) | join("\n")' /usr/share/microshift/release/release-$(uname -i).json | xargs -n1 podman pull --authfile /etc/crio/openshift-pull-secret
+EOF
+    # Disable firewalld otherwise generated bundle have it running and each podman container
+    # which try to expose a port need to added to firewalld rule manually
+    # also in case of microshift the ports like 2222, 443, 80 ..etc need to be manually added
+    # and OCP/OKD/podman bundles have it disabled by default.
```
cfergeau, line 69: Isn't it odd that microshift's firewalld configuration would block ports 443/80 by default? How am I supposed to reach applications running in a 'real' microshift installation?

createdisk-library.sh
```diff
@@ -383,3 +383,18 @@ function download_podman() {
...
     fi
 }
+
+# As of now sparsify helper is very specific to OCP/OKD kind of bundle where we get the
+# partition for the root label and then mount it with guestfish to cleanup /boot. With
+# microshift vm we are using lvm and guestfish error out during mount
+# mount /dev/sda3 /
+# libguestfs: error: mount: mount exited with status 32: mount: /sysroot: unknown filesystem type 'LVM2_member'
+# There might be other way for guestfish to mount lvm but as of now using a seperate helper is easy.
+function sparsify_lvm() {
```
cfergeau, line 393: Can't find a comment I thought I had added, feel free to ignore this comment if you have already answered it elsweher :) imo it would make sense to add a `bundle_type` parameter to `sparsify`, and have a `lvm` and a non `lvm` codepath.

createdisk-library.sh
```diff
@@ -383,3 +383,18 @@ function download_podman() {
       mv podman-remote/windows/podman-${version}/usr/bin/podman.exe  podman-remote/windows
     fi
 }
+
+# As of now sparsify helper is very specific to OCP/OKD kind of bundle where we get the
+# partition for the root label and then mount it with guestfish to cleanup /boot. With
+# microshift vm we are using lvm and guestfish error out during mount
+# mount /dev/sda3 /
+# libguestfs: error: mount: mount exited with status 32: mount: /sysroot: unknown filesystem type 'LVM2_member'
+# There might be other way for guestfish to mount lvm but as of now using a seperate helper is easy.
```
cfergeau, line 392: I'd aim at the opposite, make use of virt-sparsify in all codepaths. Initially we moved away from it because of issues with encrypted partitions iirc, but I think these are gone. However, now `/` is immutable in the VM image, which is why we use /boot instead. and I'm not sure there's a way around this in virt-sparsify.

cfergeau (commented):
A few interrogations pending, but should be almost good to go!

microshift.sh
```diff
@@ -0,0 +1,141 @@
...
+version = "*"
+EOF
+    sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
+    sed -i '/--bootproto=dhcp/a\network  --hostname=api.crc.testing' scripts/image-builder/config/kickstart.ks.template
+    scripts/image-builder/cleanup.sh -full
+    # The home dir and files must have read permissions to group
+    # and others because osbuilder is running from another non-priviledged user account
+    # and allow it to read the files on current user home (like reading yum repo which is created as part of build script), it is required.
+    # https://github.com/openshift/microshift/blob/main/scripts/image-builder/configure.sh#L29-L32
+    chmod 0755 $HOME
```
cfergeau, line 99: fwiw, an alternative for these chmods is something like:
```
+    setfacl -m u:_osbuild-composer:x ${HOME}
+    setfacl -m u:_osbuild-composer:x ${HOME}/snc
+    setfacl -m u:_osbuild-composer:x $(pwd)
+    mkdir _output
+    setfacl --default -m u:_osbuild-composer:rx $(pwd)/_output
+    setfacl --recursive -m u:_osbuild-composer:rx ${microshift_pkg_dir}
```

createdisk.sh
praveenkumar: https://github.com/openshift/microshift/blob/release-4.12/docs/howto_firewall.md

microshift.sh
gbraad: Improvements; let's create a new PR for this

createdisk-library.sh
gbraad: This is an open question, and therefore needs more time.

> I'd aim at the opposite, make use of virt-sparsify in all codepaths.

Sounds like improvements that can be a new PR; please file an issue and see if this an improvement.

createdisk.sh
gbraad: > MicroShift does not require a firewall for normal operation. 
> The following ports are optional and they should be considered for MicroShift if a firewall is enabled.

80 | TCP | HTTP port used to serve applications through the OpenShift router
-- | -- | --
443 | TCP | HTTPS port used to serve applications through the OpenShift router

microshift.sh
```diff
@@ -0,0 +1,137 @@
...
+
+    scripts/image-builder/build.sh -microshift_rpms ${pkgDir} -pull_secret_file ${OPENSHIFT_PULL_SECRET_PATH} -lvm_sysroot_size 15360 -authorized_keys_file $(realpath ../id_ecdsa_crc.pub)
```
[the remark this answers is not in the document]
gbraad: Not sure what you mean with this comment.

microshift.sh
cfergeau: I don't even think it's a lot better than the `chmod` to be honest, just sharing alternatives here.

microshift.sh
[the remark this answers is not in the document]
gbraad: Resolving as not a blocker.

microshift.sh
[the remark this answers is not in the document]
gbraad: `make rpm` not used.

createdisk-library.sh
cfergeau: Yes, this is an open question and won't be fixed now. However it should  not have taken multiple iterations on this PR to get a clear explanation why code was duplicated.

microshift.sh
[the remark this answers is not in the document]
cfergeau: `-lvm_sysroot_size 15360` here, `--disk path=...,size=31` in the `virt-install` commandline, these 2 values are related, we'll need to introduce constants for these at some point, or even better, to change `15360` to `0.5 * 31 * 1024`. (emphasis on "at some point")

createdisk.sh
cfergeau: This is also documented in https://access.redhat.com/documentation/en-us/red_hat_build_of_microshift/4.12/html/installing/microshift-embed-in-rpm-ostree#accessing-microshift-cluster-remotely-as-admin_microshift-embed-in-rpm-ostree , quite unexpected that a microshift cluster is unable to communicate externally by default ;)

microshift.sh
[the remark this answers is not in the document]
cfergeau: This does not have much to do with `make rpm`, `mktemp` becomes `mkdir -p _output/rpmbuild`, and then we can skip the `-microshift_rpms ${pkgDir}` argument to `build.sh`, we don't need the `chmod` nor to delete it afterwards.

createdisk.sh
cfergeau: Fwiw, I'd put these links as comment/justification and remove the manually written one.

cfergeau (approved):
Thanks a lot for working on all these comments during the review process !
Looks good to me :)
The few remaining things which came up during the process can be polished later on.

PR #593, opened 2022-11-02
Remove in place release image modification

praveenkumar:
All those steps not works any more because CVO reconcile it so we need to use an alternative approach like https://issues.redhat.com/browse/OTA-799 one.


cfergeau:
> All those steps not works any more because CVO reconcile it so we need to use an alternative approach like https://issues.redhat.com/browse/OTA-799 one.

Not sure the issue is public, a quick summary of the approach could be useful to have here.

praveenkumar:
> > All those steps not works any more because CVO reconcile it so we need to use an alternative approach like https://issues.redhat.com/browse/OTA-799 one.
> 
> Not sure the issue is public, a quick summary of the approach could be useful to have here.

@cfergeau we will create another PR which have the steps which is part of this jira issue.

PR #595, opened 2022-11-04
podman: Install qemu-guest-agent in the bundle

cfergeau:
This will be needed by the timesync code (set the VM time automatically when the machine gets out of sleep).

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -165,6 +165,25 @@ function prepare_hyperV() {
             echo 'CONST{virt}=="microsoft", RUN{builtin}+="kmod load hv_sock"' > /etc/udev/rules.d/90-crc-vsock.rules
 EOF
 }
+function prepare_qemu_guest_agent() {
+    local vm_ip=$1
+
+    install_additional_packages ${vm_ip} qemu-guest-agent
+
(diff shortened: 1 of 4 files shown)


createdisk-library.sh
```diff
@@ -165,6 +165,25 @@ function prepare_hyperV() {
...
+function prepare_qemu_guest_agent() {
+    local vm_ip=$1
+
+    install_additional_packages ${vm_ip} qemu-guest-agent
+
+    # f36 default selinux policy blocks usage of qemu-guest-agent over vsock
+    # checkpolicy
+    /usr/bin/checkmodule -M -m -o qemuga-vsock.mod qemuga-vsock.te
+    # policycoreutils
+    /usr/bin/semodule_package -o qemuga-vsock.pp -m qemuga-vsock.mod
```
praveenkumar, line 177: @cfergeau these commands need to run on the host can't we run it on the VM because our CI system is based on the RHEL8 host so I am wondering if those steps going to work as expected.

cfergeau: The needed packages are not available in the VM, this is why I run them on the host. We may need to install additional packages (the comments are the package names), but hopefully this will work fine on rhel8 when they are installed.

gbraad (approved):

PR #611, opened 2022-11-23
Wait for kubelet-serving signer request also

praveenkumar:
`kubelet-serving` csr is created in few sec after `kube-apiserver-client-kubelet` is approved. Sometime `kubelet-serving` approved automatic but sometime it remain pending and our bundle is created. `kubelet-serving` csr request is responsible for kubelet-server cert and if that remain pending then `kubelet-server` cert is not valid for longer period.

This PR adds another check just to make sure it is approved as soon as created and also add check for cert vailidy for kubelet-server-current certs.

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -189,7 +189,11 @@ function renew_certificates() {
     start_vm ${vm_prefix}
 
     # After cluster starts kube-apiserver-client-kubelet signer need to be approved
-    timeout 500s bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
+    timeout 500s bash -c -- "until ${OC} get csr | grep kube-apiserver-client-kubelet | grep Pending; do echo 'Waiting for first kube-apiserver-client-kubelet CSR request.'; sleep 2; done"
+    ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
+
+    # After kube-apiserver-client-kubelet signer, kubelet-serving signer need to be approved
+    timeout 500s bash -c -- "until ${OC} get csr | grep kubelet-serving | grep Pending; do echo 'Waiting for first kubelet-serving CSR request.'; sleep 2; done"
     ${OC} get csr -ojsonpath='{.items[*].metadata.name}' | xargs ${OC} adm certificate approve
 
     # Retry 10 times to make sure kubelet certs are rotated correctly.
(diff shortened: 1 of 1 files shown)


snc-library.sh
```diff
@@ -189,7 +189,11 @@ function renew_certificates() {
     start_vm ${vm_prefix}
 
     # After cluster starts kube-apiserver-client-kubelet signer need to be approved
-    timeout 500s bash -c -- "until ${OC} get csr | grep Pending; do echo 'Waiting for first CSR request.'; sleep 2; done"
```
anjannath, line 192: Maybe keep this line around for catching ths CSRs other then `kube-apiserver-client-kubelet` and `kubelet-serving`? now we are filtering on these two names only and there is also the `kube-apiserver-client` CSR, not sure if there can be more!

praveenkumar: @anjannath as of now I think we are only care about `kube-apiserver-client-kubelet` and `kubelet-serving` because other csr automatic approved by machine approver. Actually we only care about `kube-apiserver-client-kubelet` but as of now this is just make sure if not approved then we do it.

anjannath: Okay, thanks for explaining!

praveenkumar:
> Any explanations for this "sometimes" behaviour?

@cfergeau it is something related to certificate approver pod but unfortunately I don't have any logs, also it is not 100% reproducible, in my experience out of 10 retry it happen 1-2 times. If it happen again I will try to capture the pod logs.

snc-library.sh
```diff
@@ -206,7 +210,12 @@ function renew_certificates() {
     done
 
     if ! ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- sudo openssl x509 -checkend 2160000 -noout -in /var/lib/kubelet/pki/kubelet-client-current.pem; then
-        echo "Certs are not yet rotated to have 30 days validity"
+        echo "kublet client certs are not yet rotated to have 30 days validity"
```
cfergeau, line 213: `kubelet`, not `kublet`, and same for the similar message below.

cfergeau:
> > Any explanations for this "sometimes" behaviour?
> 
> @cfergeau it is something related to certificate approver pod but unfortunately I don't have any logs, also it is not 100% reproducible, in my experience out of 10 retry it happen 1-2 times. If it happen again I will try to capture the pod logs.

This sounds like a workaround for an unidentified bug lurking somewhere :-/

praveenkumar:
> > > Any explanations for this "sometimes" behaviour?
> > 
> > 
> > @cfergeau it is something related to certificate approver pod but unfortunately I don't have any logs, also it is not 100% reproducible, in my experience out of 10 retry it happen 1-2 times. If it happen again I will try to capture the pod logs.
> 
> This sounds like a workaround for an unidentified bug lurking somewhere :-/

Agree, somewhere in openshift side, right now it is causing some blocking issue so just careful on our side.

cfergeau:
>  right now it is causing some blocking issue so just careful on our side.

Yes, of course we need a workaround of some sort, but until my messages *nothing* in the PR was explicitly saying that this most likely workarounds some bug some place else.

cfergeau (approved):
Hope the root cause for this issue will be found and fixed.

PR #614, opened 2022-11-23
Add capabilities for 4.12

praveenkumar:


cfergeau:
What are the differences before/after this PR? do we have exactly the same operators enabled/disabled? or are there more/less operators disabled? Would this change how we enable the monitoring stack for example?

cfergeau:
> Yes as of now we have exactly same operators are disabled as we have now.

This would be useful to have in the commit log to know there are no expected behaviour changes from this PR.

By the way, the end result is a mix of capabilities + cvo override, what do we gain from this?

praveenkumar:
> By the way, the end result is a mix of capabilities + cvo override, what do we gain from this? It seems more complex than before?

@cfergeau Right now it is mix with capabilities + cvo but moving forward it will be only capability if you check after this PR I think we only have override for monitoring and some MCO side other that that everything handled by capability and may be soon monitoring also become part of capability.

cfergeau:
> > By the way, the end result is a mix of capabilities + cvo override, what do we gain from this? It seems more complex than before?
> 
> @cfergeau Right now it is mix with capabilities + cvo but moving forward it will be only capability if you check after this PR I think we only have override for monitoring and some MCO side other that that everything handled by capability and may be soon monitoring also become part of capability.

Are there bugs/enhancements/... tracking the use of capabilities for monitoring/... ?

praveenkumar:
> > > By the way, the end result is a mix of capabilities + cvo override, what do we gain from this? It seems more complex than before?
> > 
> > 
> > @cfergeau Right now it is mix with capabilities + cvo but moving forward it will be only capability if you check after this PR I think we only have override for monitoring and some MCO side other that that everything handled by capability and may be soon monitoring also become part of capability.
> 
> Are there bugs/enhancements/... tracking the use of capabilities for monitoring/... ?

https://issues.redhat.com/browse/MON-2854

cfergeau (approved):

PR #615, opened 2022-11-23
Remove render machine config which have pull secret info

praveenkumar:
Right now we are removing the machineconfig which is generated after pull secret changes happen and machine config operator fails to reconcile in case of openshift-4.12 because as soon as we delete the pull secret it is not able to pull image to inspect the content which is added

- https://github.com/openshift/machine-config-operator/commit/1671a453f1a95ff9e3720719c4151f289e2fdd0c

```
I1123 10:31:08.119905    4150 run.go:19] Running: podman pull -q --authfile /var/lib/kubelet/config.json quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ffa3568233298408421ff7da60e5c594fb63b2551c6ab53843eb51c8cf6838ba
Error: initializing source docker://quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ffa3568233298408421ff7da60e5c594fb63b2551c6ab53843eb51c8cf6838ba: reading manifest sha256:ffa3568233298408421ff7da60e5c594fb63b2551c6ab53843eb51c8cf6838ba in quay.io/openshift-release-dev/ocp-v4.0-art-dev: unauthorized: access to the requested resource is not authorized
```

This is still a workaround because ideally mco should able to reconcile in case there is no pull secret on the cluster (OKD usecase).

In this PR we are not removing one of the older machineconfig which MCO wants to reconcile from and not able to because of podman pull issue so when crc provision the cluster it also add the correct pull secret and then MCO reconcile happen.

Upstream Jira issue: https://issues.redhat.com/browse/OCPBUGS-4049

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -254,6 +254,6 @@ wait_till_cluster_stable openshift-marketplace
 retry ${OC} delete pod --field-selector=status.phase==Succeeded --all-namespaces
 
 # Delete outdated rendered master/worker machineconfigs and just keep the latest one
-${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master | head -n -1 | xargs -t ${OC} delete
+${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-master | head -n -2 | xargs -t ${OC} delete
 ${OC} get mc --sort-by=.metadata.creationTimestamp --no-headers -oname | grep rendered-worker | head -n -1 | xargs -t ${OC} delete
 


cfergeau:
>  as soon as we delete the pull secret it is not able to pull image to inspect the content which is added

I don't really understand the end of this sentence :-/ we delete the pull secret, but the MCO still needs it for some background work it's doing? 

> In this PR we are not removing one of the older machineconfig which MCO wants to reconcile from and not able to because of podman pull issue so when crc provision the cluster it also add the correct pull secret and then MCO reconcile happen.

And I did not understand the description of what we are doing either :(

praveenkumar:
> > as soon as we delete the pull secret it is not able to pull image to inspect the content which is added
> 
> I don't really understand the end of this sentence :-/ we delete the pull secret, but the MCO still needs it for some background work it's doing?

yes MCO still using it to pull the image for inspect the os layering part.

> 
> > In this PR we are not removing one of the older machineconfig which MCO wants to reconcile from and not able to because of podman pull issue so when crc provision the cluster it also add the correct pull secret and then MCO reconcile happen.
> 
> And I did not understand the description of what we are doing either :(

Till now we are removing all the  rendered machine config for master except the latest one. In this PR we are remove all but except latest 2. 

Why? - Because of upstream bug https://issues.redhat.com/browse/OCPBUGS-4049 

How this workaround work:-

After the `snc.sh` script run the state of machine config operator will be same as what is listed in the bug and we create bundle out of it. ( This is not blocking snc.sh script to run successfully)
```
$ oc get co machine-config
NAME             VERSION       AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
machine-config   4.12.0-rc.1   True        False         True       93m     Failed to resync 4.12.0-rc.1 because: error during syncRequiredMachineConfigPools: [timed out waiting for the condition, error pool master is not ready, retrying. Status: (pool degraded: true total: 1, ready 0, updated: 0, unavailable: 0)]
```

Once we run this bundle MCO will find the user pull secret and start reconciling which makes the machine-config to available again. 

If we delete the second last MC (like we are doing till now) then machin-config operator fails with below error and not able to reconcile because it point to an rendered machine config which already deleted.
```
$ oc logs machine-config-daemon-6pczk -n openshift-machine-config-operator
[...]
E1123 08:41:54.131381    3674 writer.go:200] Marking Degraded due to: prepping update: machineconfig.machineconfiguration.openshift.io "rendered-master-1217872a72f41af7514c83cc6defb701" not found
```

cfergeau:
> yes MCO still using it to pull the image for inspect the os layering part.

Imo it's a bit too many details, what matters is that the MCO still tries to use the pull secret after we delete it 

> Till now we are removing all the rendered machine config for master except the latest one. In this PR we are remove all but except latest 2.

Ok, written this way I understand better :) 

> If we delete the second last MC (like we are doing till now) then machine-config operator fails with below error and not able to reconcile because it point to an rendered machine config which already deleted.

Do you know why the MCO isn't using the last `rendered-master` in the list, but the one before the last? And even if we keep 2 rendered-master, the pull secret is still removed? (I think so according to the logs you posted).

praveenkumar:
> Do you know why the MCO isn't using the last `rendered-master` in the list, but the one before the last? 

@cfergeau This is the bug, second last have details about removing the pull secret and it doesn't apply it correctly because of inspect failure. 

> And even if we keep 2 rendered-master, the pull secret is still removed? (I think so according to the logs you posted).

Yes, pull secret is not part of second-last rendered MC.

cfergeau (commented):
> Till now we are removing all the rendered machine config for master except the latest one.
> In this PR we are remove all but except latest because of upstream bug

Shouldn't this "In this PR we keep the latest one and the one before the last because of ..." ?

> This is still a workaround because ideally mco should able to reconcile in case there is no pull secret on the cluster (OKD usecase).
>
> How this workaround work:
>
> After the snc.sh script executed, the state of machine config operator will be same
as what is listed in the bug and we create bundle out of it. (This is not blocking snc.sh script to run successfully)

I find all of this confusing :( Maybe this if this is accurate:

> With this PR, after the snc.sh script runs, the state of machine config operator will be degraded (see example below),  but this is not blocking snc.sh script to run successfully.
> Ideally mco should able to reconcile in case there is no pull secret on the cluster (OKD usecase).

PR #618, opened 2022-12-05
[Master] Add `SYSTEM_RESERVED_ES` value to node sizing env

praveenkumar:
Looks like in case of 4.12 setting `SYSTEM_RESERVED_ES` is mandatory, otherwise kubelet start with `--system-reserved=cpu=200m,memory=350Mi,ephemeral-storage=` and failed with error.

```
command failed" err="failed to run Kubelet: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP][-+]?[0-9])$'"
```

This will fix https://issues.redhat.com/browse/OCPBUGS-3758 from crc side.

MCO change: https://github.com/openshift/machine-config-operator/commit/28f062ba1815c622ed17718baf3194ecd75d4194


cfergeau:
> [openshift/machine-config-operator#3381](https://github.com/openshift/machine-config-operator/pull/3381) this change is now part of 4.11 also.

Have you experienced the failure no 4.11? Or only noticed that this was cherry-picked? Can you add this comment to the commit log?
Apart from this
/lgtm

praveenkumar:
> Have you experienced the failure no 4.11? Or only noticed that this was cherry-picked?

@cfergeau  Yes, I have experienced a failure for 4.11 and then cherry picked.

cfergeau:
I wonder if we should look into using https://github.com/openshift/machine-config-operator/blob/cbd7d9514d6a03c84192058edae7a6ff4f9d6941/pkg/controller/kubelet-config/helpers.go#L435-L480
and https://github.com/openshift/machine-config-operator/blob/cbd7d9514d6a03c84192058edae7a6ff4f9d6941/vendor/k8s.io/kubelet/config/v1beta1/types.go#L596-L602

The end result seems similar to what we want, but there is a fallback when SYSTEM_RESERVED_ES is not specified. Probably more future-proof than the approach we are taking right now.

PR #620, opened 2022-12-05
Pick M1 kernel from rhel9.1

cfergeau:
With RHEL 9.1 officially released, we can get the kernel from there, it
should be newer than the ones in RHEL 9.0

### DIFF
--- a/repos/rhel9-internal.repo
+++ b/repos/rhel9-internal.repo
@@ -1,5 +1,5 @@
 [rhel-9.0-baseos]
 name=rhel9 baseos
-baseurl=https://rhsm-pulp.corp.redhat.com/content/eus/rhel9/9.0/$basearch/baseos/os/
+baseurl=https://rhsm-pulp.corp.redhat.com/content/eus/rhel9/9.1/$basearch/baseos/os/
 enabled=1
 gpgcheck=0


repos/rhel9-internal.repo
```diff
@@ -1,5 +1,5 @@
 [rhel-9.0-baseos]
 name=rhel9 baseos
-baseurl=https://rhsm-pulp.corp.redhat.com/content/eus/rhel9/9.0/$basearch/baseos/os/
+baseurl=https://rhsm-pulp.corp.redhat.com/content/eus/rhel9/9.1/$basearch/baseos/os/
```
praveenkumar, line 3: @cfergeau Right now https://rhsm-pulp.corp.redhat.com/content/eus/rhel9/ doesn't have 9.1 directory, it is going to break the link.

praveenkumar (approved):

praveenkumar (changes requested):
It is breaking change

repos/rhel9-internal.repo
cfergeau: Ah thanks! I was relying on our CI to tell us if the link exists :)

praveenkumar: our CI doesn't run arm bits so this might not even run :(

cfergeau:
OpenShift sticks to even RHEL versions which are EUS, we can follow the same releases for this m1-specific workaround. Closing this issue.

PR #623, opened 2022-12-08
build qemu-ga selinux policy package inside the guest

anjannath:
the version of selinux policy kit where the package is built should be the same in the target machine where the package is installed in our case the host where we are building is rhel9 and the targets os is rhel8, where it fails to install with the following error:

```
libsemanage.semanage_pipe_data: Child process /usr/libexec/selinux/hll/pp failed with code: 255. (No such file or directory).
qemuga-vsock: libsepol.policydb_read: policydb module version 20 does not match my version range 4-19
qemuga-vsock: libsepol.sepol_module_package_read: invalid module in module package (at section 0)
qemuga-vsock: Failed to read policy package
libsemanage.semanage_direct_commit: Failed to compile hll files into cil files.
 (No such file or directory).
semodule:  Failed!
```

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -210,13 +210,11 @@ function prepare_qemu_guest_agent() {
     local vm_ip=$1
 
     # f36 default selinux policy blocks usage of qemu-guest-agent over vsock
-    # checkpolicy
-    /usr/bin/checkmodule -M -m -o qemuga-vsock.mod qemuga-vsock.te
-    # policycoreutils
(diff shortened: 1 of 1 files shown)


anjannath:
some terminologies might be off in the commit log! is the selinux version or the selinux policykit version that should be similar?

cfergeau (commented):
The error message uses `policydb module version 20`, I don't think it has anything to do with policykit, just selinux. You can use `The version of pre-built selinux modules must be the same as or older than the one the target machine expects. In our case ...`

PR #625, opened 2022-12-12
Install checkpolicy package to fcos to execute checkmodule command

praveenkumar:
After b13f44cfa09ca65c2d68665dc45225f1030ce2dc, qemu-ga selinux policy built inside the guest which require checkmodule command to check and compile selinux policy. Currently RHCOS have checkpolicy package installed by default but it is not part of FCOS. This patch install the checkpolicy package in case of OKD bundle creation.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -209,6 +209,10 @@ EOF
 function prepare_qemu_guest_agent() {
     local vm_ip=$1
 
+    if [[ ${BASE_OS} = "fedora-coreos" ]]; then
+        install_additional_packages ${vm_ip} checkpolicy
+    fi
+
     # f36 default selinux policy blocks usage of qemu-guest-agent over vsock
     ${SCP} qemuga-vsock.te core@${vm_ip}:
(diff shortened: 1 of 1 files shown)


cfergeau:
I'd prefer to switch back to doing this on the host and use this `checkmodule` option:  `  -c VERSION build a policy module targeting a modular policy version (4-21)`

cfergeau:
> I'd prefer to switch back to doing this on the host and use this `checkmodule` option: ` -c VERSION build a policy module targeting a modular policy version (4-21)`

`-c 19` should be fine for rhel8

cfergeau:
> > I'd prefer to switch back to doing this on the host and use this `checkmodule` option: ` -c VERSION build a policy module targeting a modular policy version (4-21)`
> 
> `-c 19` should be fine for rhel8

https://github.com/crc-org/snc/pull/626 implements this suggestion, but I haven't tested yet.

praveenkumar:
/close

#626 have it handled.

PR #626, opened 2022-12-12
qemuga: Ensure selinux policy module is rhel8-compatible

cfergeau:
RHEL8 only supports policy modules up to version 19, while fedora and rhel9
use newer versions. When building on the host, we need to make sure the
policy module we build is compatible with rhel8 which is what the openshift
bundles use as a base OS.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -210,11 +210,13 @@ function prepare_qemu_guest_agent() {
     local vm_ip=$1
 
     # f36 default selinux policy blocks usage of qemu-guest-agent over vsock
-    ${SCP} qemuga-vsock.te core@${vm_ip}:
-    ${SSH} core@${vm_ip} '/usr/bin/checkmodule -M -m -o qemuga-vsock.mod qemuga-vsock.te'
-    ${SSH} core@${vm_ip} '/usr/bin/semodule_package -o qemuga-vsock.pp -m qemuga-vsock.mod'
+    # checkpolicy
+    /usr/bin/checkmodule -M -m -o qemuga-vsock.mod qemuga-vsock.te
+    # policycoreutils
+    /usr/bin/semodule_package -o qemuga-vsock.pp -m qemuga-vsock.mod
 
-    ${SSH} core@${vm_ip} 'sudo semodule -i qemuga-vsock.pp && rm qemuga-vsock.pp qemuga-vsock.mod qemuga-vsock.te'
+    ${SCP} qemuga-vsock.pp core@${vm_ip}:
+    ${SSH} core@${vm_ip} 'sudo semodule -i qemuga-vsock.pp && rm qemuga-vsock.pp'
     ${SCP} qemu-guest-agent.service core@${vm_ip}:
     ${SSH} core@${vm_ip} 'sudo mv -Z qemu-guest-agent.service /etc/systemd/system/'
     ${SSH} core@${vm_ip} 'sudo systemctl daemon-reload'
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -209,10 +209,18 @@ EOF
 function prepare_qemu_guest_agent() {
     local vm_ip=$1
 
-    # f36 default selinux policy blocks usage of qemu-guest-agent over vsock
-    # checkpolicy
-    /usr/bin/checkmodule -M -m -o qemuga-vsock.mod qemuga-vsock.te
-    # policycoreutils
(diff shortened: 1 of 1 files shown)


praveenkumar:
/hold

During testing it for okd bundle where host is RHEL-8.7 it fails with following error. Looks like only new version of checkmodule have option to specifiy the version of policy module. 

```
$ rpm -qa | grep check
checkpolicy-2.9-1.el8.x86_64

$ /usr/bin/checkmodule -c 19 -M -m -o qemuga-vsock.mod qemuga-vsock.te
/usr/bin/checkmodule: invalid option -- 'c'
usage:  /usr/bin/checkmodule [-h] [-V] [-b] [-C] [-U handle_unknown] [-m] [-M] [-o FILE] [INPUT]
Build base and policy modules.
Options:
  INPUT      build module from INPUT (else read from "policy.conf")
  -V         show policy versions created by this program
  -b         treat input as a binary policy file
  -C         output CIL policy instead of binary policy
  -h         print usage
  -U OPTION  How to handle unknown classes and permissions
               deny: Deny unknown kernel checks
               reject: Reject loading of policy with unknowns
               allow: Allow unknown kernel checks
  -m         build a policy module instead of a base module
  -M         enable MLS policy
  -o FILE    write module to FILE (else just check syntax)

```

createdisk-library.sh
```diff
@@ -210,11 +210,15 @@ function prepare_qemu_guest_agent() {
     local vm_ip=$1
 
     # f36 default selinux policy blocks usage of qemu-guest-agent over vsock
-    ${SCP} qemuga-vsock.te core@${vm_ip}:
-    ${SSH} core@${vm_ip} '/usr/bin/checkmodule -M -m -o qemuga-vsock.mod qemuga-vsock.te'
-    ${SSH} core@${vm_ip} '/usr/bin/semodule_package -o qemuga-vsock.pp -m qemuga-vsock.mod'
+    if ! checkmodule -c 19 2>&1 |grep 'invalid option' >/dev/null; then
+	    # RHEL8 checkmodule does not have this arg
+	    MOD_VERSION_ARG="-c 19"
+    fi
```
praveenkumar, line 216: This block not work as expected because as per man page `The  return status of a pipeline is the exit status of the last command, unless the pipefail option is enabled.  If pipefail is enabled, the pipeline's return status is the value
       of the last (rightmost) command to exit with a non-zero status, or zero if all commands exit successfully. `  in our case pipefail option is set for the script.

```
$ cat ./test.sh 
#!/bin/bash

set -eou pipefail

if ! checkmodule -c 19 2>&1 | grep 'invalid option' >/dev/null; then
	    # RHEL8 chAeckmodule does not have this arg
	    MOD_VERSION_ARG="-c 19"
	    echo "added -c 19"
fi

echo ${MOD_VERSION_ARG}

$ ./test.sh 
added -c 19
-c 19
```

praveenkumar (changes requested):

createdisk-library.sh
cfergeau: Not sure what the best way around this is?
Wrap it in `set +o pipefail`?

```
    set +o pipefail
    if ! checkmodule -c 19 2>&1 |grep 'invalid option' >/dev/null; then
	    # RHEL8 checkmodule does not have this arg
	    MOD_VERSION_ARG="-c 19"
    fi
    set -o pipefail
```

praveenkumar: @cfergeau yes I think we should do that and add comment why we are unsetting pipefail for this block of code.

createdisk-library.sh
```diff
@@ -209,12 +209,22 @@ EOF
...
+    #
+    # we need to disable pipefail for the `checkmodule | grep check` as we expect `checkmodule`
+    # to fail on rhel8.
+    set +o pipefail
+    if ! checkmodule -c 19 2>&1 |grep 'invalid option' >/dev/null; then
+	    # RHEL8 checkmodule does not have this arg
+	    MOD_VERSION_ARG="-c 19"
+    fi
+    set -o pipefail
+    /usr/bin/checkmodule ${MOD_VERSION_ARG} -M -m -o qemuga-vsock.mod qemuga-vsock.te
```
praveenkumar, line 223: @cfergeau It is now fail with 
```
createdisk-library.sh: line 223: MOD_VERSION_ARG: unbound variable
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```
So we do need a default value for this variable, like empty string.

praveenkumar (changes requested):

praveenkumar:
working now.

/lgtm

PR #627, opened 2022-12-14
build-patched-kao-kcmo-images.sh improvements

cfergeau:
This PR adds 2 new features to the script:
- it automatically regenerates crc-dnsmasq and crc-routes-controller so that they use the same base image as the rest of the cluster
- it builds a patched release image so that it's only done once

### DIFF
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -44,7 +44,7 @@ else
     if test -n "${OPENSHIFT_RELEASE_VERSION}"; then
         echo "Using release ${OPENSHIFT_RELEASE_VERSION} from the mirror"
     else
-        echo "Unable to determine an OpenShift release version.  You may want to set the OPENSHIFT_VERSION environment variable explicitly."
+        echo "Unable to determine an OpenShift release version. You may want to set the OPENSHIFT_VERSION environment variable explicitly."
         exit 1
     fi
 fi
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -20,6 +20,8 @@ export LANG=C.UTF-8
 
 rm -fr crc-cluster-kube-apiserver-operator
 rm -fr crc-cluster-kube-controller-manager-operator
+rm -fr crc-dnsmasq
+rm -fr crc-routes-controller
 
 function check_pull_secret() {
         if [ -z "${OPENSHIFT_PULL_SECRET_PATH-}" ]; then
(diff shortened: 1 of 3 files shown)


cfergeau:
I added a commit to make use of the pre-built release image in snc.sh instead of rebuilding it, but I haven't tested this commit.

build-patched-kao-kcmo-images.sh
```diff
@@ -90,5 +91,32 @@ function patch_and_push_image() {
...
+    git reset --hard origin/crc-1-rhel-8
+    sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
+    git add Dockerfile
+    git commit -m "Use OpenShift ${openshift_version} base image"
+    git push origin -f HEAD:private-cfergeau
+    rhpkg container-build --scratch --target crc-1-rhel-8-candidate
+    popd
+
+    skopeo copy --dest-authfile ${OPENSHIFT_PULL_SECRET_PATH} --all docker://registry-proxy.engineering.redhat.com/rh-osbs/${brew_repo}:latest docker://quay.io/crcont/${brew_repo#crc-}:${openshift_version}
+    skopeo copy --dest-authfile ${OPENSHIFT_PULL_SECRET_PATH} --all docker://registry-proxy.engineering.redhat.com/rh-osbs/${brew_repo}:latest docker://quay.io/crcont/${brew_repo#crc-}:latest
```
praveenkumar, line 146: Would've been clean solution if `--additional-tag` from `skopeo copy` also support docker transport but it only support `docker-archive` only.

snc.sh
```diff
@@ -75,8 +75,8 @@ fi
 
 if [[ ${USE_PATCHED_RELEASE_IMAGE} == "enabled" ]]
 then
-   echo "Creating a new image with patched KAO/KCMO images"
-   create_new_release_with_patched_images
+   echo "Using release image with patched KAO/KCMO images"
+   OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=quay.io/crcont/ocp-release:${OPENSHIFT_RELEASE_VERSION}-${yq_ARCH}
```
praveenkumar, line 79: If we are using the `${OPENSHIFT_RELEASE_VERSION}-${yq_ARCH}` then what is the use of having single multiple arch image on the quay side? Did you test `openshift-install` make use of this multi arch image instead of arch specific one.

cfergeau: I haven't tested `openshift-install` with an image manifest. Ideally we can delete the -arch images after the manifest is cretaed, but that depends on `openshift-install`. Even if it cannot use it, does not hurt to have the manifest in addition to the individual images.

PR #636, opened 2023-01-03
Use openshift version as image tag for dnsmasq/route-controller if presnt

praveenkumar:
We started to create images for dnsmasq and route-controller as part of patched KAO/KCMO images and those images are tagged with openshift version. It is good to tag dnsmasq and route-controller image also with openshift version if present otherwise use latest.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -57,12 +57,17 @@ fi
 
 prepare_qemu_guest_agent api.${CRC_VM_NAME}.${BASE_DOMAIN}
 
+image_tag="latest"
+if podman manifest inspect quay.io/crcont/dnsmasq:${OPENSHIFT_VERSION} >/dev/null 2>&1; then
+    image_tag=${OPENSHIFT_VERSION}
+fi
+
 # Add gvisor-tap-vsock and crc-dnsmasq services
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
The dnsmasq image has nothing openshift related in it, it's tagged with the openshift version because it was theeasiest way of getting a unique tag for it. But  in general, you want to use the latest one as it will be the one with the latest security fixes. Using `latest` makes more sense to me than hardcoding some openshift version.

routes-controller may end up depending on the openshift version, so why not.

praveenkumar:
> The dnsmasq image has nothing openshift related in it, it's tagged with the openshift version because it was theeasiest way of getting a unique tag for it. But in general, you want to use the latest one as it will be the one with the latest security fixes. Using `latest` makes more sense to me than hardcoding some openshift version.

We wanted to solve 2 problem, one which you mentioned about security fixes and another one is using the same base image as the openshift release have to avoid extra layer for bundle. Let's assume I am trying to build 4.12 bundle and then 4.13 rc/ec comes then our dnsmasq image might be based on RHEL-9 ( as base for 4.13) so either we always use latest RHEL-8 image and don't base it on openshift or we should use same version.

PR #642, opened 2023-01-13
Use src-tls-verify=false for source registry

praveenkumar:
This will resolve the issue on the slave where required trusted cert for
`registry-proxy.engineering.redhat.com` registry is not present.

```
time="2023-01-12T05:34:55-05:00" level=fatal msg="initializing source docker://registry-proxy.engineering.redhat.com/rh-osbs/openshift-crc-cluster-kube-apiserver-operator:v4.12.0-202301052135.p0.g336ffd5.assembly.stream: pinging container registry registry-proxy.engineering.redhat.com: Get \"[https://registry-proxy.engineering.redhat.com/v2/\](https://registry-proxy.engineering.redhat.com/v2//)": x509: certificate signed by unknown authority"
```


cfergeau:
Could we add the needed CA cert and use it?

praveenkumar:
> Could we add the needed CA cert and use it?

@cfergeau In the pipeline those slave are created on demand `dynamic` on PSI infra so either we need to have a way to get the RH cert and add it first or use this `src-tls-verify=false` also since it is mostly internal (brew registry is not public facing) so should not be a issue of using this way?

build-patched-kao-kcmo-images.sh
```diff
@@ -95,7 +95,10 @@ function patch_and_push_image() {
     git checkout --track origin/rhaos-${OCP_VERSION}-rhel-8
     git merge --no-ff -m "Merge commit ${vcs_ref} into rhaos-${OCP_VERSION}-rhel-8" -m "MaxFileSize: 104857600" ${vcs_ref}
     git push origin HEAD
-    rhpkg container-build --target crc-1-rhel-8-candidate
+    # If brew build already exist for the release don't rebuild it again
```
cfergeau, line 98: When does this happen? I thought we'd only run this when there are new cluster-kube-apiserver-operator / cluster-kube-controller-manager-operator builds?

build-patched-kao-kcmo-images.sh
```diff
@@ -95,7 +95,10 @@ function patch_and_push_image() {
...
+    if ! brew list-builds --package=crc-${image_name}-container | grep -q ${release}; then
```
cfergeau, line 99: `brew buildinfo crc-${image_name}-container-${release}` instead of `list-builds`+`grep`? Might be better to return early before the `rhpkg clone` when we don't need to build?

build-patched-kao-kcmo-images.sh
```diff
@@ -139,13 +139,11 @@ function update_base_image() {
...
     git checkout --track origin/crc-1-rhel-8
-    git checkout --track origin/private-cfergeau
-    git reset --hard origin/crc-1-rhel-8
     sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
     git add Dockerfile
     git commit -m "Use OpenShift ${openshift_version} base image"
-    git push origin -f HEAD:private-cfergeau
-    rhpkg container-build --scratch --target crc-1-rhel-8-candidate
+    git push origin
+    rhpkg container-build --target crc-1-rhel-8-candidate
```
cfergeau, line 146: I'm unconvinced it's usefulu to explicitly specify the target, as it's a unusual way of running `rhpkg container-build`

build-patched-kao-kcmo-images.sh
praveenkumar: @cfergeau `rhpkg clone` is needed because that is what dnsmasq/router images consume for updating the base image.

build-patched-kao-kcmo-images.sh
praveenkumar: explicit is always better than implicit.

build-patched-kao-kcmo-images.sh
praveenkumar: it happen when a new version of openshift release (mostly z stream) where these component might not even change so this way when we build the image it should not  fail.

build-patched-kao-kcmo-images.sh
praveenkumar: > brew buildinfo crc-${image_name}-container-${release} instead of list-builds+grep? 

Sure, this we can update.

build-patched-kao-kcmo-images.sh
cfergeau: I disagree. "oh, they specify --target, they must not be using the default target for that branch. That's odd, what is the default target? And why do they need a different one?"

build-patched-kao-kcmo-images.sh
cfergeau: > @cfergeau rhpkg clone is needed because that is what dnsmasq/router images consume for updating the base image.

If we don't need to rebuild crc-cluster-kube-apiserver-operator, this means in particular its base image won't have changed, so we don't need to rebuild the dnsmasq/router images, the ones we have are already up to date.

build-patched-kao-kcmo-images.sh
praveenkumar: Or they are not using any target so what is the default one? both time user need to look for the doc. As new user of brew I don't know the branch is default target because in case of patch container images we are building tagged images from the branches and if we don't specify the target it would fail so I still prefer to have it explicit.

build-patched-kao-kcmo-images.sh
praveenkumar: > If we don't need to rebuild crc-cluster-kube-apiserver-operator, this means in particular its base image won't have changed, so we don't need to rebuild the dnsmasq/router images, the ones we have are already up to date.

Sure, in that case we first check if the respective directory exist and if not then not need to build `dnsmasq/router` images.

build-patched-kao-kcmo-images.sh
cfergeau: But it's not important to know what target is being used, unless you are debugging that very code. 99% of the code, `rhpkg container-build` is good enough.

build-patched-kao-kcmo-images.sh
praveenkumar: And in that case we can push that change before rhpkg so we don't need to pull it.

build-patched-kao-kcmo-images.sh
praveenkumar: > Ah they don't rebuild everything even when it's unchanged? ok!

Yes in the past that what I observed.

build-patched-kao-kcmo-images.sh
praveenkumar: I understand for folks who are familiar with brew/koji have no issue to know that default target is taken as your branch if not provided but for anyone else who are new to it will going to have bit hard time if he is doing something in different branch and build fails because it will as for target and he have to figure out what could be target.

Anyway I have remove it  but I still think explicit is always better (no harm to put extra info even it is default),  but don't want to block the PR because of this.

cfergeau: I'm not going to block this either if you keep the --target, but too much information really is harmful in itself

build-patched-kao-kcmo-images.sh
```diff
@@ -87,16 +87,19 @@ function patch_and_push_image() {
...
-    git remote add upstream git://pkgs.devel.redhat.com/containers/ose-${image_name}
-    # Just fetch the upstream/rhaos-${OCP_VERSION}-rhel-8 instead of all the branches and tags from upstream
-    git fetch upstream rhaos-${OCP_VERSION}-rhel-8 --no-tags
-    git checkout --track origin/rhaos-${OCP_VERSION}-rhel-8
-    git merge --no-ff -m "Merge commit ${vcs_ref} into rhaos-${OCP_VERSION}-rhel-8" -m "MaxFileSize: 104857600" ${vcs_ref}
-    git push origin HEAD
-    rhpkg container-build --target crc-1-rhel-8-candidate
-    popd
+    # If brew build already exist for the release don't rebuild it again
+    if ! brew buildinfo crc-${image_name}-container-${version}-${release}; then
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau As I told you for minor releases when these component doesn't change we still need that release tag on the quay so `skopeo copy` make that happen. Assume 4.12.1 and 4.12.2 both contain same component then on the quay only `4.12.1` is available but we need both so here basically we are just adding tag using `skopeo copy` .

build-patched-kao-kcmo-images.sh
```diff
@@ -154,7 +157,11 @@ patch_and_push_image cluster-kube-apiserver-operator
...
-
-update_base_image crc-dnsmasq "${base_image}"
-update_base_image crc-routes-controller "${base_image}"
+# In case there is no change in the openshift component then the base
+# image is also not changed so no need to build dnsmasq/route images
+if [ -f crc-cluster-kube-apiserver-operator/Dockerfile ]; then
+    base_image=$(grep "^FROM openshift/ose-base" crc-cluster-kube-apiserver-operator/Dockerfile | sed 's/^FROM //')
+    
+    update_base_image crc-dnsmasq "${base_image}"
+    update_base_image crc-routes-controller "${base_image}"
```
cfergeau, line 165: Arguably we also need to call `skopeo copy --dest-authfile ${OPENSHIFT_PULL_SECRET_PATH} --all docker://registry-proxy.engineering.redhat.com/rh-osbs/${brew_repo}:latest docker://quay.io/crcont/${brew_repo#crc-}:${openshift_version}` even if we did not rebuild anything, but we don't use the openshift version tag for these, so we can also remember that if an openshift release is missing, this means it used the previous one.

cfergeau (approved):

build-patched-kao-kcmo-images.sh
praveenkumar: @cfergeau Right we should but the problem is as of now we are using `docker://registry-proxy.engineering.redhat.com/rh-osbs/${brew_repo}:latest` so if I run this script for 4.12.8 and those images never built on the brew then everything works as expected but then even by mistake someone run this script for 4.12.6 where the image already exist so the brew build not happen for KAO/KCMO and just tagging happen (which is again pushing same image on the same tag) but if we add same logic for dnsmasq/router image then latest is 4.12.8 and it will update the same for 4.12.6. Which is also OK as long as there is no base image change but might create issue if image is updated. 

What we need is same logic as we have for KAO/KCMO where we identify build and if present then don't rebuild but if not then build.

PR #643, opened 2023-01-13
baseimage: Use non-scratch builds

cfergeau:
Looks like I kept some testing changes when I pushed the PR to
keep the base container images in sync with the cluster version.
We should be using non-scratch builds as otherwise it's quite hard to
pick the right image from RH registries.

### DIFF
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -135,13 +135,11 @@ function update_base_image() {
     rhpkg clone containers/${brew_repo}
     pushd ${brew_repo}
     git checkout --track origin/crc-1-rhel-8
-    git checkout --track origin/private-cfergeau
-    git reset --hard origin/crc-1-rhel-8
     sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
     git add Dockerfile
     git commit -m "Use OpenShift ${openshift_version} base image"
-    git push origin -f HEAD:private-cfergeau
-    rhpkg container-build --scratch --target crc-1-rhel-8-candidate
+    git push origin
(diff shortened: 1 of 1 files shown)


build-patched-kao-kcmo-images.sh
```diff
@@ -135,13 +135,11 @@ function update_base_image() {
...
     git checkout --track origin/crc-1-rhel-8
-    git checkout --track origin/private-cfergeau
-    git reset --hard origin/crc-1-rhel-8
     sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
     git add Dockerfile
     git commit -m "Use OpenShift ${openshift_version} base image"
-    git push origin -f HEAD:private-cfergeau
-    rhpkg container-build --scratch --target crc-1-rhel-8-candidate
+    git push origin
+    rhpkg container-build
```
praveenkumar, line 142: `container-build` without the target? does it take target `crc-1-rhel-8-candidate` as default?

cfergeau: Since at this point the current branch is `crc-1-rhel-8`, yes, it should use `crc-1-rhel-8-candidate` automatically.

praveenkumar: @cfergeau I added this commit to https://github.com/crc-org/snc/pull/642 and make it explicit, I think it should be ok.

cfergeau: You typically don't specify the target when using `rhpkg  build` (for example when you build a RPM for rhel), adding it could be slightly confusing ("why aren't they using the default target?")

praveenkumar:
This is merged as part of #642 

/close

PR #644, opened 2023-01-16
Some refactor to enable to use createdisk.sh for microshift preset also

praveenkumar:
Initial refactor which help us to allow use of same script for microshift preset. Some more changes required but that will happen as part of #592 . So this initial patch is to make sure everything still works and able to generate the bundle.


tools.sh
```diff
@@ -130,21 +130,22 @@ function get_vm_prefix {
...
-    local vm_prefix=$1
-    retry sudo virsh start ${vm_prefix}-master-0
+    local vm_name=$1
+    local vm_ip=$2
+    retry sudo virsh start ${vm_name}
     # Wait till ssh connection available
-    until ${SSH} core@api.${CRC_VM_NAME}.${BASE_DOMAIN} -- "exit 0" >/dev/null 2>&1; do
-        echo " ${vm_prefix}-master-0 still booting"
+    until ${SSH} core@${vm_ip} -- "exit 0" >/dev/null 2>&1; do
+        echo " ${vm_name} still booting"
```
cfergeau, line 148: Might be worth configuring https://libvirt.org/nss.html at some point so that we don't need the IP in addition to the VM name. With libvirt-nss configured, we can just do `${SSH} core@${CRC_VM_NAME}`

createdisk.sh
```diff
@@ -28,9 +28,10 @@ OPENSHIFT_VERSION=$(${JQ} -r .clusterInfo.openshiftVersion $INSTALL_DIR/crc-bund
 # VM_PREFIX: full VM name with the random string generated by openshift-installer
 CRC_VM_NAME=${CRC_VM_NAME:-crc}
 VM_PREFIX=$(get_vm_prefix ${CRC_VM_NAME})
+VM_NAME="${VM_PREFIX}-master-0"
```
cfergeau, line 31: One of `VM_NAME` or `CRC_VM_NAME` should be renamed, or it's going to be very confusing. `SHORT_VM_NAME` and `FULL_VM_NAME` maybe?
Even `VM_PREFIX=$(get_vm_prefix ${CRC_VM_NAME})` is odd, `VM_PREFIX` is something longer than `CRC_VM_NAME`, I'd expect a prefix of ${CRC_VM_NAME}, ie something shorter.

cfergeau (commented):
Overall looks good!

createdisk.sh
praveenkumar: `CRC_VM_NAME=${CRC_VM_NAME:-crc}`  should be `CRC_PRODUCT_NAME=${CRC_PRODUCT_NAME:-crc}` becaue `crc` is just product name and it happened to part of the VM name.

tools.sh
praveenkumar: I am goingto create a followup issue for that instead of putting that effort right now. It is a good suggestion.

createdisk.sh
cfergeau: Just `PRODUCT_NAME`then if we expect this can be different from `crc`? Or `SNC_PRODUCT_NAME`if you want this namespaced?

praveenkumar: @cfergeau Added a new commit for this since it happen to be this variable consumed in different helper/script.

createdisk.sh
```diff
@@ -24,10 +24,10 @@ fi
 BASE_OS=${BASE_OS:-rhcos}
 OPENSHIFT_VERSION=$(${JQ} -r .clusterInfo.openshiftVersion $INSTALL_DIR/crc-bundle-info.json)
 
-# CRC_VM_NAME: short VM name to use in crc_libvirt.sh
+# SNC_PRODUCT_NAME: short VM name to use in crc_libvirt.sh
```
cfergeau, line 27: It's not really a "short VM name" anymore, and `crc_libvirt.sh` has been gone for a very long time, would be worth updating the comment at the same time.

createdisk.sh
```diff
@@ -24,10 +24,10 @@ fi
...
 # VM_PREFIX: full VM name with the random string generated by openshift-installer
-CRC_VM_NAME=${CRC_VM_NAME:-crc}
-VM_PREFIX=$(get_vm_prefix ${CRC_VM_NAME})
```
cfergeau: "full VM name" for something called "VM_PREFIX" is so confusing :-/  It's is the short VM name + the random string generated by openshift-installer. Then full VM names/disk image names/... can be constructed by appending `-base`, `-master-0`, ... (I know you know that, just writing up explicitly what it is to help update the  comment ^^)

cfergeau (commented):
A few minor comments, apart from this /lgtm /approve

PR #649, opened 2023-01-19
Use global 'MIRROR' variable through out script

praveenkumar:
We don't want to update same value to local and global variable.

### DIFF
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -56,8 +56,7 @@ fi
 
 function release_image_for_arch() {
      local arch=$1
-     local mirror="https://mirror.openshift.com/pub/openshift-v4/${arch}/clients/ocp"
-     curl -L "${mirror}/${OPENSHIFT_RELEASE_VERSION}/release.txt" 2>/dev/null| sed -n 's/^Pull From: //p'
+     curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/release.txt" 2>/dev/null| sed -n 's/^Pull From: //p'
 }
 
 if test -z "${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE-}"; then


build-patched-kao-kcmo-images.sh
```diff
@@ -56,8 +56,7 @@ fi
 
 function release_image_for_arch() {
      local arch=$1
-     local mirror="https://mirror.openshift.com/pub/openshift-v4/${arch}/clients/ocp"
-     curl -L "${mirror}/${OPENSHIFT_RELEASE_VERSION}/release.txt" 2>/dev/null| sed -n 's/^Pull From: //p'
+     curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/release.txt" 2>/dev/null| sed -n 's/^Pull From: //p'
```
praveenkumar, line 59: `MIRROR` is declared in https://github.com/crc-org/snc/pull/649/files#diff-628d5fcbbf9c964c1d78fb50659bd6f0e9b79250ff1ba01b7ebf3a4fac5f3095L41

cfergeau (changes requested):

build-patched-kao-kcmo-images.sh
praveenkumar: Does that mean we are creating different arch image from different arch host?

cfergeau: This function was added in https://github.com/crc-org/snc/commit/39986c909337476d638ebba1d422dec8dd496266 in preparation for https://github.com/crc-org/snc/commit/5d2b5a314b3b0a99ab855af83341748afa2d9a43

Initially build-patched-kao-kcmo-images.sh was looking at the openshift release image content to get to a dist-git commit. The arch of the openshift release image it was looking at was not important as the dist-git commit is the same for all arches.
However when we started patching the release image in  build-patched-kao-kcmo-images.sh we need to patch the release image for the correct arch.
The script is doing
```
    for arch in amd64 arm64; do \
        create_patched_release_image_for_arch ${upstream_registry} ${arch}
```
so the end result won't depend on the arch the script runs on.

praveenkumar: Got it now, completely forgot about it and didn't look for the reference of this function call :(

PR #663, opened 2023-03-09
Fix podman static binary for linux which now have arch info

### LINKED ISSUE #662: podman - 4.4.2 static binary name is changed and now contain arch info
https://github.com/containers/podman/releases/tag/v4.4.1 now have `podman-remote-static-linux_arm64.tar.gz ` and `podman-remote-static-linux_amd64.tar.gz` instead `podman-remote-static.tar.gz` , this need to updated in our podman bundle creation side.

praveenkumar:
podman-remote linux binary now have arch info and binary is placed under bin folder before creating the tarball. We now need to extract it with bin folder and remove it after moving it to correct location.
```
$ curl -L -O https://github.com/containers/podman/releases/download/v4.4.1/podman-remote-static-linux_amd64.tar.gz
$ tar -tf podman-remote-static-linux_amd64.tar.gz
./bin/podman-remote-static-linux_amd64
```

fix #662

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -276,8 +276,9 @@ function download_podman() {
     local arch=$2
 
     mkdir -p podman-remote/linux
-    curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-static.tar.gz | tar -zx -C podman-remote/linux podman-remote-static
-    mv podman-remote/linux/podman-remote-static podman-remote/linux/podman-remote
+    curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-static-linux_${arch}.tar.gz | tar -zx -C podman-remote/linux ./bin/podman-remote-static-linux_${arch}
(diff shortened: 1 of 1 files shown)


gbraad (approved):

praveenkumar:
> > and binary is placed under bin folder before creating the tarball.
> 
> This is only the case for 4.4.1, 4.4.2 does not have a `bin/` directory. A similar patch would be good to have in `master` as well to keep both files in sync.

@cfergeau on the master we will have this patch once RHCOS have 4.4.1, as of now it is 4.2.0 https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/contents.html?stream=prod/streams/4.12&release=412.86.202303071430-0&arch=x86_64

praveenkumar:
> This is only the case for 4.4.1, 4.4.2 does not have a bin/ directory.

@cfergeau yes, I am not sure why they keep changing the file structure for each binary :( will keep #662 issue open for tracking issue.

PR #666, opened 2023-03-13
set acl to allow the qemu user to access snc dir recursively

anjannath:
this fixes the following error on the snc job:
```
load /home/hudson/jenkins/workspace/eng-crc-snc-arm/crc-tmp-install-data/fcos-config.ign: Failed to open file "/home/hudson/jenkins/workspace/eng-crc-snc-arm/crc-tmp-install-data/fcos-config.ign": Permission denied
```

PR #667, opened 2023-03-15
Move ign file to /var/lib/libvirt/images directory

praveenkumar:
This patch fix the acl permission issue which is faced because of directory structure and for CI we run it `$HOME` but for internal CI it is part of jenkins workspace directory to avoid setting up acl on directory level (even recursive) lets use the default `/var/lib/libvirt/images` location which have right acls.


snc.sh
```diff
@@ -44,6 +44,7 @@ ${PODMAN} run --pull=always --rm -i quay.io/coreos/ignition-validate:release - <
 mkdir ${PWD}/${CRC_INSTALL_DIR}/tmp/
 ${PODMAN} run --pull=always --rm -v ${PWD}/${CRC_INSTALL_DIR}/tmp/:/data:Z -w /data quay.io/coreos/coreos-installer:release download -a ${ARCH} -s stable -p qemu -f qcow2.xz --decompress
 sudo mv ${CRC_INSTALL_DIR}/tmp/fedora-coreos-*-qemu.${ARCH}.qcow2 /var/lib/libvirt/images/fedora-coreos-qemu.${ARCH}.qcow2
+sudo mv -Z ${CRC_INSTALL_DIR}/fcos-config.ign /var/lib/libvirt/images/
 rmdir ${PWD}/${CRC_INSTALL_DIR}/tmp/
 
 sudo setfacl -m u:qemu:rx $HOME
```
anjannath, line 50: This `setfacl` call can be removed now i think

praveenkumar: Ah yes.

cfergeau:
> This patch fix the acl permission issue which is faced because of directory structure and for CI we run it `$HOME` but for internal CI it is part of jenkins workspace directory to avoid setting up acl on directory level (even recursive) lets use the default `/var/lib/libvirt/images` location which have right acls.

But the .ign file is not an image, `/var/lib/libvirt-images` is for virtual machine images. It might be convenient to put the file there, but if we can understand what's going on with the ACLs in the home dir, it's useful knowledge.

snc.sh
```diff
@@ -44,17 +44,17 @@ ${PODMAN} run --pull=always --rm -i quay.io/coreos/ignition-validate:release - <
 mkdir ${PWD}/${CRC_INSTALL_DIR}/tmp/
 ${PODMAN} run --pull=always --rm -v ${PWD}/${CRC_INSTALL_DIR}/tmp/:/data:Z -w /data quay.io/coreos/coreos-installer:release download -a ${ARCH} -s stable -p qemu -f qcow2.xz --decompress
 sudo mv ${CRC_INSTALL_DIR}/tmp/fedora-coreos-*-qemu.${ARCH}.qcow2 /var/lib/libvirt/images/fedora-coreos-qemu.${ARCH}.qcow2
+sudo mv -Z ${CRC_INSTALL_DIR}/fcos-config.ign /var/lib/libvirt/images/
```
cfergeau, line 47: If we go this way (I'd rather not), please prefix it with `crc` or `snc`, and make sure it's cleaned up when no longer needed.

praveenkumar: > I'd rather not

@cfergeau is there any implication ?

 > please prefix it with crc or snc, and make sure it's cleaned up when no longer needed.

Added `snc` prefix and you want to cleanup only the `.ign` file or also the `qcow2` which is used?

cfergeau: `virsh undefine` should be removing the disk images associated with the VM,, but I don't think it's going to do this for the .ign file.

cfergeau:
I think Anjan has been making progress with fixing the permissions in the other PR.

praveenkumar:
/close

since #666 is working with the right acls.

anjannath:
> I think Anjan has been making progress with fixing the permissions in the other PR.

now i am thinking this is better then #666 see: https://github.com/crc-org/snc/pull/666#discussion_r1138158338

anjannath:
LGTM

anjannath (approved):

PR #668, opened 2023-03-16
Use ${SNC_PRODUCT_NAME}.${BASE_DOMAIN} instead of crc.testing

eranco74:
This will create the currect registry-certs configmap incase the user changed the default SNC_PRODUCT_NAME or BASE_DOMAIN

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -229,7 +229,7 @@ ${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- 'sudo mv /home/core/kubeco
 
 # Add exposed registry CA to VM
 retry ${OC} extract secret/router-ca --keys=tls.crt -n openshift-ingress-operator --confirm
-retry ${OC} create configmap registry-certs --from-file=default-route-openshift-image-registry.apps-crc.testing=tls.crt -n openshift-config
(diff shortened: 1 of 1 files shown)


cfergeau:
A similar change will be needed in `microshift.sh` once it lands. Otherwise, this makes sense to me!

praveenkumar:
> A similar change will be needed in microshift.sh once it lands. Otherwise, this makes sense to me!

@cfergeau you mean as part of hostname `sed -i '/--bootproto=dhcp/a\network  --hostname=api.crc.testing' scripts/image-builder/config/kickstart.ks.template`  or some other place?

cfergeau:
> > A similar change will be needed in microshift.sh once it lands. Otherwise, this makes sense to me!
> 
> @cfergeau you mean as part of hostname `sed -i '/--bootproto=dhcp/a\network --hostname=api.crc.testing' scripts/image-builder/config/kickstart.ks.template` or some other place?

yes this one

eranco74:
e2e-snc failed due to:
`
Suite run returned error: AlertErr: services "prometheus-k8s" not found
`
 which is a bit strange since we ddon't expect prometheus to run on CRC, perhaps this test should be excluded from the test suite?

praveenkumar:
> which is a bit strange since we ddon't expect prometheus to run on CRC, perhaps this test should be excluded from the test suite?

@eranco74 yes some of the test cases are obsolete and need to be removed from CI side.

PR #673, opened 2023-03-24
Update gvisor-tap service by using nmcli and binary from container image

praveenkumar:
This patch do following
- Create a tap device using nmcli with a hard coded mac address, this mac address is mapped in the crc daemon lease which provide `192.168.127.2` IP address to the VM.
- copy the `vm` binary from container image and put it to /usr/local/bin
- modify gvisor-tap-vsock service to use the `vm` binary instead running the container using podman

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -92,10 +92,32 @@ if podman manifest inspect quay.io/crcont/dnsmasq:${OPENSHIFT_VERSION} >/dev/nul
     image_tag=${OPENSHIFT_VERSION}
 fi
 
+# create the tap device interface with specified mac address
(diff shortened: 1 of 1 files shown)


cfergeau (commented):
Had these comments around, never posted them apparently :(

cfergeau:
https://github.com/containers/gvisor-tap-vsock/pull/202 is exploring something similar.

cfergeau:
I've done some work on this in https://github.com/cfergeau/snc/tree/gvisor_service
However for testing I used the podman branch and https://github.com/cfergeau/snc/tree/podman-gvisor-service , which worked fine. I'll have to test the openshift changes ^^

PR #675, opened 2023-03-31
Use podman machine marker instead deprecated enabled machine config

praveenkumar:
With https://github.com/containers/common/commit/0db57c8ff1272546f13100d464674366d98323a9 `machine_enabled=true/false` setting in the containers.conf is deprecated and now there is `podman-machine` marker file, which is used for this setting.

- https://github.com/containers/common/blob/main/pkg/machine/machine.go#L21

### DIFF
--- a/fcos-config.yaml
+++ b/fcos-config.yaml
@@ -112,13 +112,10 @@ storage:
          [containers]
          netns="bridge"
          rootless_networking="cni"
-      # Set machine_enabled to true to indicate we're in VM
-    - path: /etc/containers/containers.conf
+      # Create empty file as machine marker for podman
+      # https://github.com/containers/common/blob/main/pkg/machine/machine.go
+    - path: /etc/containers/podman-machine
       mode: 0644
-      contents:
-        inline: |
-          [engine]
-          machine_enabled=true
     - path: /etc/NetworkManager/dispatcher.d/pre-up.d/gvisor-tap-vsock-tap-device.sh
       mode: 0755
       contents:


createdisk-library.sh
```diff
@@ -183,7 +183,7 @@ function downgrade_kernel() {
     local arch=$2
     case $arch in
          amd64)
-            ${SSH} core@${vm_ip} "curl -L -O https://kojipkgs.fedoraproject.org/packages/kernel/5.19.17/300.fc37/x86_64/kernel-5.19.17-300.fc37.x86_64.rpm -L -O https://kojipkgs.fedoraproject.org/packages/kernel/5.19.17/300.fc37/x86_64/kernel-core-5.19.17-300.fc37.x86_64.rpm -L -O https://kojipkgs.fedoraproject.org/packages/kernel/5.19.17/300.fc37/x86_64/kernel-modules-5.19.17-300.fc37.x86_64.rpm"
+            ${SSH} core@${vm_ip} "curl -L -O https://kojipkgs.fedoraproject.org//packages/kernel/5.19.16/301.fc37/x86_64/kernel-5.19.16-301.fc37.x86_64.rpm -L -O https://kojipkgs.fedoraproject.org//packages/kernel/5.19.16/301.fc37/x86_64/kernel-core-5.19.16-301.fc37.x86_64.rpm -L -O https://kojipkgs.fedoraproject.org//packages/kernel/5.19.16/301.fc37/x86_64/kernel-modules-5.19.16-301.fc37.x86_64.rpm"
```
cfergeau, line 186: Can this be moved to its own PR? Feels better to unbreak podman bundle generation as soon as possible in git, while keeping the option of delaying machine marker landing if desirable.

praveenkumar: @cfergeau I need to test the bundle with marker file so put in same PR but I can have separate PR for kernel stuff.

praveenkumar: https://github.com/crc-org/snc/pull/678 here.

praveenkumar:
tested with the bundle and it works as expected.
```
$ ./crc start -b ~/Downloads/crc_podman_libvirt_4.4.1_amd64.pr675.crcbundle
[...]
$ eval $(./crc podman-env)
$ podman-remote run -d -p 8080:80 docker.io/httpd:2.4
Storing signatures
412ca013e63ff3f890529837bb54326399108871b216da9e989e3f874573e1fd
$ curl -Ik localhost:8080
HTTP/1.1 200 OK
```

PR #676, opened 2023-03-31
Use podman machine marker instead deprecated enabled machine config

praveenkumar:
With https://github.com/containers/common/commit/0db57c8ff1272546f13100d464674366d98323a9 `machine_enabled=true/false` setting in the containers.conf is deprecated and now there is `podman-machine` marker file, which is used for this setting.

- https://github.com/containers/common/blob/main/pkg/machine/machine.go#L21

### DIFF
--- a/podman_changes.ks
+++ b/podman_changes.ks
@@ -15,10 +15,7 @@ rootless_networking="cni"
 EOF
 chown -R core:core /home/core/.config/containers
 
-tee /etc/containers/containers.conf <<EOF
-[engine]
-machine_enabled=true
-EOF
+touch /etc/containers/podman-machine
 
 tee /etc/containers/registries.conf.d/999-podman-machine.conf <<EOF
 unqualified-search-registries=["docker.io"]


praveenkumar:
/hold

need to test the bundle.

podman_changes.ks
```diff
@@ -15,10 +15,7 @@ rootless_networking="cni"
 EOF
 chown -R core:core /home/core/.config/containers
 
-tee /etc/containers/containers.conf <<EOF
-[engine]
-machine_enabled=true
-EOF
+touch /etc/containers/podman-machine
```
cfergeau, line 18: It's expected to contain `qemu` (possible values are `qemu` or `wsl`)

cfergeau:
Support for this landed in podman 4.2.0

podman_changes.ks
praveenkumar: @cfergeau no that is not required, only having this marker file is enough `qemu` and `wsl` only require to run gvisor stack or not, for us we want to run gvisor stack for both mac/windows.
- https://github.com/containers/common/blob/main/pkg/machine/machine.go#L67
- https://github.com/containers/common/blob/main/pkg/machine/machine_test.go#L28-L34 (mostly this test)

cfergeau: Hmm the implementation supports empty files indeed, though podman-machine does not use an empty file: https://github.com/containers/podman/blob/3e44a7afed7a5751ef700729f97b9f6e955fcc5e/pkg/machine/ignition.go#L421-L435

Probably best to stay as close as possible to podman-machine?

praveenkumar: > podman-machine does not use an empty file

@cfergeau it is not using empty file because for them they have different implementation for qemu and wsl side, for us we don't have anything different where we dont' want to run the gvisor vsock. So `machineEnabled` as of now just this maker file. In future if we need different implementation then we might need to add the respective provider info.

cfergeau: > In future if we need different implementation then we might need to add the respective provider info.

We are not the ones parsing this file, the file is something intended for podman use. And in the current podman code, I don't think podman ever creates an empty `podman-machine` file, hence my suggestion to stay as close as possible to podman.
Sure, nothing breaks **now** because podman is always checking against the `wsl` value. Hopefully they'll never start testing for `qemu` in other code.

cfergeau (approved):
Not going to fight for `qemu`, even if the PR could be more future-proof.

podman_changes.ks
praveenkumar: Having `qemu` in the marker file is wrong for us because we are not even using `qemu` for mac/windows. What we want is just letting podman know if it have machine enabled or not and having this file without any content make more sense to me to replicate what is is deprecated now.

PR #683, opened 2023-04-04
Revert "aarch64: Use latest released rhel9 kernel"

praveenkumar:
This reverts commit 19372d4e919d713bbd735aec807042c9704a9e97.

OCP-4.13 will have RHEL-9.2 content by default so we don't need to update the kernel for aarch64.


praveenkumar:
/hold

until we create and  test arm bundle using this PR.

praveenkumar:
/close 

we are going to downgrade rhel9.

PR #685, opened 2023-04-06
podman: Fix extraction of podman 4.4.2 binary

cfergeau:
Its path in the archive changed again with the 4.4.2 release which
landed in FCOS 37.20230322.3.0 stable


gbraad (approved):
`Its path in the archive changed again`
can be simplified with less opinion:
`Path changed to exclude ./bin`

PR #689, opened 2023-04-17
CI: add script to create bundles for microshift

praveenkumar:


praveenkumar:
Required for https://github.com/openshift/release/pull/38200 one.

ci_microshift.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -exuo pipefail
+
+sudo yum install -y podman make golang rsync
```
anjannath, line 5: podman is also installed by the `microshift.sh` script, might not be needed here? https://github.com/crc-org/snc/blob/cfe2c7dc246fd8efb769a6cd0bdbb80d7280d3b0/microshift.sh#L57

anjannath (approved):

ci_microshift.sh
anjannath: also `rsync` is not being used by the script?

ci_microshift.sh
```diff
@@ -0,0 +1,28 @@
...
+
+./shellcheck.sh
+./microshift.sh
+
+# Run createdisk script
```
cfergeau, line 10: If you add a comment, it's more useful to document why we use these zstd flag

ci_microshift.sh
```diff
@@ -0,0 +1,30 @@
...
+set -exuo pipefail
+
+sudo yum install -y make golang
+
+./shellcheck.sh
+./microshift.sh
+
+# Run createdisk script
+# Set the zstd compression level to 10 to have faster
+# compression.
```
cfergeau, line 12: There's a `--fast` option if the goal is to get fast compression. You could also drop `--long`?

ci_microshift.sh
```diff
@@ -0,0 +1,30 @@
...
+export CRC_ZSTD_EXTRA_FLAGS="-10 --long"
+./createdisk.sh crc-tmp-install-data
+
+git clone https://github.com/code-ready/crc.git
```
gbraad, line 16: Perhaps add comment why we run with latest of crc.

gbraad (approved):
Latest is a clear choice. Worth adding comment.
Otherwise /lgtm

ci_microshift.sh
cfergeau: *grin* `/code-ready/`?

ci_microshift.sh
praveenkumar: with fast the bundle size would be not that optimized so using this is kind of proper balance the size of bundle and time taken.

PR #692, opened 2023-04-17
build-patch-script: Update the arch info using sed instead duplicate

praveenkumar:
As of now we use `MIRROR` which can be changed using this env variable or it has default value but then we have a local mirror where we want to create custom images for different arch and for that we are again using `mirror` local variable which need to updated. With this PR we just use `MIRROR` variable and consume it both the place.

### DIFF
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -37,8 +37,8 @@ function check_pull_secret() {
 
 check_pull_secret
 
-ARCH=$(uname -m)
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+HOST_ARCH=$(uname -m)
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$HOST_ARCH/clients/ocp}
 
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 if test -n "${OPENSHIFT_VERSION-}"; then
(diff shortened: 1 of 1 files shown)


build-patched-kao-kcmo-images.sh
```diff
@@ -56,7 +56,7 @@ fi
 
 function release_image_for_arch() {
      local arch=$1
-     local mirror="https://mirror.openshift.com/pub/openshift-v4/${arch}/clients/ocp"
+     local mirror=$(echo ${MIRROR} | sed "s/$ARCH/$arch/g")
```
cfergeau, line 59: Can you add a `mirror_for_arch` helper instead, I think it would be a lot easier to understand what it's doing.

praveenkumar: you mean something like this?
```bash
function release_image_for_arch() {
     local arch=$1
     local mirror=mirror_for_arch
     curl -L "${mirror}/${OPENSHIFT_RELEASE_VERSION}/release.txt" 2>/dev/null| sed -n 's/^Pull From: //p'
}

function mirror_for_arch() {
    echo ${MIRROR} | sed "s/$ARCH/$arch/g"
}
```

cfergeau:
```
diff --git a/build-patched-kao-kcmo-images.sh b/build-patched-kao-kcmo-images.sh
index 8d0358a..be372ab 100755
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -35,10 +35,15 @@ function check_pull_secret() {
         fi
 }
 
+function mirror_for_arch() {
+	local arch=$1
+	echo "https://mirror.openshift.com/pub/openshift-v4/$arch/clients/ocp"
+}
+
 check_pull_secret
 
 ARCH=$(uname -m)
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+MIRROR=$(mirror_for_arch $ARCH)
 
 # If user defined the OPENSHIFT_VERSION environment variable then use it.
 if test -n "${OPENSHIFT_VERSION-}"; then
@@ -56,7 +61,7 @@ fi
 
 function release_image_for_arch() {
      local arch=$1
-     local mirror="https://mirror.openshift.com/pub/openshift-v4/${arch}/clients/ocp"
+     local mirror=$(mirror_for_arch $arch)
      curl -L "${mirror}/${OPENSHIFT_RELEASE_VERSION}/release.txt" 2>/dev/null| sed -n 's/^Pull From: //p'
 }

 ```

praveenkumar: @cfergeau this patch remove the `MIRROR` env so in the downstream CI, I can choose MIRROR to `ocp-dev-preview` and `ocp` when creating the patch images it will be then not possible.

cfergeau: But `release_image_for_arch` already did not use `MIRROR` from the environment, is this PR fixing an issue related to this?

praveenkumar: @cfergeau we want to consume `MIRROR` on both the place so it we can avoid duplication and also able to use different URL by using this `MIRROR` env.

cfergeau, line 59: Maybe rename `ARCH` to `HOST_ARCH` for more clarity? `s/$ARCH/$arch` is not obvious at all :)

I'd include the `/` around `$ARCH` to limit the risk of unexpected matches: `sed "s;/$ARCH/;/$arch/;g"`

cfergeau (commented):
After further discussion on slack, the real problem (and the motivation for this PR) is this:

release_image_for_arch hardcodes "https://mirror.openshift.com/pub/openshift-v4/$%7Barch%7D/clients/ocp" and does not use MIRROR if set. This will be causing issues in 4.13 where the script needs to be run with MIRROR="https://mirror.openshift.com/pub/openshift-v4/$%7Barch%7D/clients/ocp-dev-preview"

With that in mind, this PR looks good to me, though there could be some minor readability improvements

cfergeau (approved):

Issue #700, opened 2023-04-25
[podman] Downgrade kernel is failing because of  `kernel-modules-core` package


praveenkumar:
We again have the issue during downgrade of kernel, so with latest fedora37 following are the kernel packages installed

```
$ rpm -qa | grep kernel
kernel-modules-core-6.2.9-300.fc38.x86_64
kernel-core-6.2.9-300.fc38.x86_64
kernel-modules-6.2.9-300.fc38.x86_64
kernel-6.2.9-300.fc38.x86_64
```

As as per our downgrade script we use https://kojipkgs.fedoraproject.org/packages/kernel/5.19.16/301.fc37/x86_64/ which doesn't contain any package for `kernel-modules-core` so during downgrade we get following error

```
ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o IdentitiesOnly=yes -i id_ecdsa_crc core@192.168.122.43 'sudo rpm-ostree override -C replace *.rpm'
Warning: Permanently added '192.168.122.43' (ECDSA) to the list of known hosts.
Checking out tree 441aeb0...done
Resolving dependencies...done
[0m[31merror: [0mCould not depsolve transaction; 1 problem detected:
 Problem: package kernel-modules-core-6.2.8-200.fc37.x86_64 requires kernel-uname-r = 6.2.8-200.fc37.x86_64, but none of the providers can be installed
  - cannot install both kernel-core-5.19.16-301.fc37.x86_64 and kernel-core-6.2.8-200.fc37.x86_64
  - conflicting requests
 ```

PR #701, opened 2023-04-26
podman: Fix kernel downgrade

cfergeau:
This currently fails with dependency issues related to 
missing kernel-uname-r dependencies. 
This PR should fix that. During my testing, first try failed, and second
attempt succeeded, so this must go through CI first :)

Fixes: https://github.com/crc-org/snc/issues/700

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -178,20 +178,22 @@ function install_additional_packages() {
 }
 
 function downgrade_kernel() {
-    # workaround https://github.com/crc-org/vfkit/issues/11 on macOS
+    # workaround https://github.com/crc-org/vfkit/issues/11 on macOS 11/12
     local vm_ip=$1
     local arch=$2
+    local bodhi_url
     case $arch in
          amd64)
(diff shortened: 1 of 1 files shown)


createdisk-library.sh
```diff
@@ -193,7 +193,11 @@ function downgrade_kernel() {
 	    ;;
     esac
 
-    ${SSH} core@${vm_ip} "sudo rpm-ostree override -C replace ${bodhi_url}"
+    ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora.repo'
+    ${SSH} core@${vm_ip} -- 'sudo sed -i -z s/enabled=0/enabled=1/ /etc/yum.repos.d/fedora-updates.repo'
+    ${SSH} core@${vm_ip} "sudo rpm-ostree override replace ${bodhi_url}"
```
praveenkumar, line 198: for consistency can use `--` here also `${SSH} core@${vm_ip} -- "sudo .."`

praveenkumar:
>This PR should fix that. During my testing, first try failed, and second
attempt succeeded, so this must go through CI first :)

CI only create the bundle for podman and doesn't run it using CRC so may be even CI is green, created bundle need manual testing. I will create an issue to atleast test that created bundle on linux.

cfergeau:
> > first try failed, and second
> 
> Can you be more specific what/how/when this failed? The resulting image failed on start up the first time, but a restart worked? or was a new build needed?

This was a failure during build, but after investigating and trying a second build, it worked. CI is green now, so this seems fine.

createdisk-library.sh
cfergeau: Force-pushed some changes to address this, but they are untested, need to wait for CI.

cfergeau: CI is green.

praveenkumar:
/lgtm
/approve


Tested with https://github.com/crc-org/snc/pull/704

PR #715, opened 2023-05-08
Issue 711 - macos: Install older kernel

praveenkumar:
Because of https://github.com/crc-org/vfkit/issues/11 we need to use an older kernel with our macOS bundles.
This is fixed in macOS 13, but we want to keep compatibility with macOS 12 for a while.
This uses the kernel which is part of RHEL-9.0 by replacing 9.2 kernel.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -114,9 +114,6 @@ cat crio-wipe.service | ${SSH} core@${VM_IP} "sudo tee -a /etc/systemd/system/cr
 ${SSH} core@${VM_IP} -- "sudo podman pull quay.io/crcont/routes-controller:${image_tag}"
 
 if [ "${ARCH}" == "aarch64" ] && [ ${BUNDLE_TYPE} != "okd" ]; then
-   # aarch64 support is mainly used on Apple M1 machines which can't run a rhel8 kernel
-   # https://access.redhat.com/solutions/6545411
-   install_rhel9_kernel ${VM_IP}
    # Install qemu-user-static-x86 packaage from fedora koji to run x86 image on M1
    # Not supported by RHEL https://access.redhat.com/solutions/5654221 and not included
    # in any subscription repo.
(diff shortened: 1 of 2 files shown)


createdisk.sh
```diff
@@ -106,6 +106,16 @@ cat crio-wipe.service | ${SSH} core@${VM_IP} "sudo tee -a /etc/systemd/system/cr
 
 # Preload routes controller
 ${SSH} core@${VM_IP} -- "sudo podman pull quay.io/crcont/routes-controller:${image_tag}"
+
+if [ ${BUNDLE_TYPE} == "snc" ]; then
```
cfergeau, line 110: Not sure what "move ... to kubelet config" means in the commit log.

createdisk.sh
```diff
@@ -106,17 +106,14 @@ cat crio-wipe.service | ${SSH} core@${VM_IP} "sudo tee -a /etc/systemd/system/cr
...
-   # aarch64 support is mainly used on Apple M1 machines which can't run a rhel8 kernel
-   # https://access.redhat.com/solutions/6545411
-   install_rhel9_kernel ${VM_IP}
-fi
-
 cleanup_vm_image ${VM_NAME} ${VM_IP}
 
 # Only used for macOS bundle generation
 if [ -n "${SNC_GENERATE_MACOS_BUNDLE}" ]; then
+    # workaround https://github.com/crc-org/vfkit/issues/11 on macOS 12
```
cfergeau, line 123: You are only telling part of the story here, the rhel 9.2 kernel *should* be working according to its versioning, and there are different issues with aarch64 and amd64 kernels. Can you add the non-working RHEL kernel versions/arch, and the version you are downgrading to either to the issue or the commit log?

cfergeau (commented):
I'll need to take a closer look, some early feedback related to the commit logs.

cfergeau (approved):

PR #719, opened 2023-05-10
createdisk.sh: Don't change dest dir if already set

praveenkumar:
As part of https://github.com/crc-org/snc/commit/d90d53d22d51fdf406e265ac964206f46d2badd0 we override the `destDirPrefix` even it is set for `okd` bundle. With this commit we don't override it and use existing value if present.


createdisk.sh
```diff
@@ -24,7 +24,7 @@ BASE_OS=${BASE_OS:-rhcos}
 OPENSHIFT_VERSION=$(${JQ} -r .clusterInfo.openshiftVersion $INSTALL_DIR/crc-bundle-info.json)
 BASE_DOMAIN=$(${JQ} -r .clusterInfo.baseDomain $INSTALL_DIR/crc-bundle-info.json)
 BUNDLE_TYPE=$(${JQ} -r .type $INSTALL_DIR/crc-bundle-info.json)
-destDirPrefix="crc"
+destDirPrefix=${destDirPrefix:-crc}
```
cfergeau, line 27: BUNDLE_TYPE will be `okd` for an okd bundle, you can move the `if [[ ${OKD_VERSION} != "none" ]]` test here, check for BUNDLE_TYPE instead (possibly with a `case`) and make the okd/microshift code consistent. You can then have a `default` case which sets `BASE_OS` to `rhcos` and `destDirPrefix` to `crc`.

cfergeau (changes requested):

cfergeau (approved):
Looks good to me, code flow is more straightforward this way! Minor comment regarding an error message.

PR #726, opened 2023-05-23
Enable user to generate mac, windows bundles or not

praveenkumar:
Right now if user want to generate or not generate  mac/win bundles for
any specific arch or in general then manual changes in the script required.

With this PR user can use following way in case don't want to generate any of those bundles.
```
SNC_GENERATE_MACOS_BUNDLE=0 ./snc.sh
SNC_GENERATE_MACOS_BUNDLE=0 ./createdisk.sh
```
And in case want to generate bundle then
```
SNC_GENERATE_MACOS_BUNDLE=1 ./snc.sh
SNC_GENERATE_MACOS_BUNDLE=1 ./createdisk.sh
```


tools.sh
```diff
@@ -22,18 +22,18 @@ ARCH=$(uname -m)
 case "${ARCH}" in
     x86_64)
         yq_ARCH="amd64"
-        SNC_GENERATE_MACOS_BUNDLE=1
-        SNC_GENERATE_WINDOWS_BUNDLE=1
+        SNC_GENERATE_MACOS_BUNDLE=${SNC_GENERATE_MACOS_BUNDLE:-1}
```
cfergeau, line 25: I'd take a slightly different approach which only needs changes in tools.sh:  `SNC_DISABLE_MACOS_BUNDLE=1 ./snc.sh` and then
```
if [ -z ${SNC_DISABLE_MACOS_BUNDLE-} ]; then
    SNC_GENERATE_MACOS_BUNDLE=1
fi
```

gbraad:
And this does not generate Linux bundles? It conflicts with expectation...

praveenkumar:
> And this does not generate Linux bundles? It conflicts with expectation...

@gbraad Linux bundles are generated for all platform and doesn't have any variable for not to generate it. Are you saying we should also have ` SNC_GENERATE_LINUX_BUNDLE` variable?

tools.sh
praveenkumar: this mean it would be looks like in tools.sh? Also how do you set the default value like for `arm64` where we don't want to generate the windows bundle by default?

```
case "${ARCH}" in
    x86_64)
            if [ -z ${SNC_DISABLE_MACOS_BUNDLE-} ]; then
                   SNC_GENERATE_MACOS_BUNDLE=1
            fi
            if [ -z ${SNC_DISABLE_WINDOWS_BUNDLE-} ]; then
                   SNC_GENERATE_WINDOWS_BUNDLE=1
            fi
    [...]
          same for all the arch            
```

cfergeau: I'd use something like this:
```
diff --git a/tools.sh b/tools.sh
index 5893404..44183a9 100755
--- a/tools.sh
+++ b/tools.sh
@@ -20,20 +20,25 @@ PATCH=${PATCH:-patch}
 ARCH=$(uname -m)
 
 case "${ARCH}" in
+    SNC_GENERATE_MACOS_BUNDLE=
+    SNC_GENERATE_WINDOWS_BUNDLE=
     x86_64)
         yq_ARCH="amd64"
-        SNC_GENERATE_MACOS_BUNDLE=1
-        SNC_GENERATE_WINDOWS_BUNDLE=1
+	if [ -z ${SNC_DISABLE_MACOS_BUNDLE-} ]; then
+            SNC_GENERATE_MACOS_BUNDLE=1
+	fi
+	if [ -z ${SNC_DISABLE_WINDOWS_BUNDLE-} ]; then
+            SNC_GENERATE_WINDOWS_BUNDLE=1
+	fi
 	;;
     aarch64)
         yq_ARCH="arm64"
-        SNC_GENERATE_MACOS_BUNDLE=1
-        SNC_GENERATE_WINDOWS_BUNDLE=
+	if [ -z ${SNC_DISABLE_MACOS_BUNDLE-} ]; then
+            SNC_GENERATE_MACOS_BUNDLE=1
+	fi
 	;;
     *)
         yq_ARCH=${ARCH}
-        SNC_GENERATE_MACOS_BUNDLE=
-        SNC_GENERATE_WINDOWS_BUNDLE=
 	;;
 esac
 ```

Comment (praveenkumar) [reply tools.sh:25]: @cfergeau even with this if now I need to generate the windows bundle for `aarch64` I need to make change in the script because the by default only I can only enable or disable macos bundle not windows one. The current PR actually allow user to enable/disable win/mac bundles for any arch without losing the default values.

Comment (cfergeau) [reply tools.sh:25]: Grmble, the PR description is:
> Right now if user don't want to generate mac/win bundles for x86_64 arch then changes in the script needed. With this PR user can use following way in case don't want to generate any of those bundles.

This says nothing about force-enabling generation of disabled bundles. My reason for suggesting alternatives is that one must remember that the magic values to use are `0` and `1`, not `yes` and `no`, or `enabled` and `disabled`.

Do you expect to often need to enable/disable windows bundle generation on the fly on aarch64 that changing the code is going to get in the way?

Comment (praveenkumar) [reply tools.sh:25]: > Grmble, the PR description is:

My bad, will update the description and not to use any specific arch. I used x86_64 as example but it created the confusion :(

> My reason for suggesting alternatives is that one must remember that the magic values to use are 0 and 1, not yes and no, or enabled and disabled.

Yes, I also thought that but again these exposed variable we should document so user can look beforehand https://github.com/crc-org/snc/issues/431 (old issue)

> Do you expect to often need to enable/disable windows bundle generation on the fly on aarch64 that changing the code is going to get in the way?

Think about if I want to enable the CI to get some experimental bundles of windows for aarch64, I can just set the variable accordingly and don't care about changing the code. If I want to generate something manually I can always change the code according.

**Comment by gbraad** [comment]
I would expect to be able to disable ANY bundle for generation...

**Comment by cfergeau** [comment]
> > I would expect to be able to disable ANY bundle for generation...
> 
> In that case I would handle the linux bundle generation in a different PR if that is OK for you?

This should do the trick, no?
```
diff --git a/createdisk.sh b/createdisk.sh
index ded2e6f..34671f9 100755
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -177,7 +177,9 @@ else
     sparsify_lvm "${libvirtDestDir}"
 fi
 copy_additional_files "$INSTALL_DIR" "$libvirtDestDir" "${VM_NAME}"
-create_tarball "$libvirtDestDir"
+if [ "${SNC_GENERATE_LINUX_BUNDLE}" == "1" ]; then
+    create_tarball "$libvirtDestDir"
+fi
 
 # vfkit image generation
 # This must be done after the generation of libvirt image as it reuses some of
```

File: tools.sh [hunk 2]
```diff
@@ -22,18 +22,18 @@ ARCH=$(uname -m)
...
+        SNC_GENERATE_WINDOWS_BUNDLE=${SNC_GENERATE_WINDOWS_BUNDLE:-1}
 	;;
     aarch64)
         yq_ARCH="arm64"
-        SNC_GENERATE_MACOS_BUNDLE=1
-        SNC_GENERATE_WINDOWS_BUNDLE=
+        SNC_GENERATE_MACOS_BUNDLE=${SNC_GENERATE_MACOS_BUNDLE:-1}
+        SNC_GENERATE_WINDOWS_BUNDLE=${SNC_GENERATE_WINDOWS_BUNDLE:-}
```
cfergeau, line 31: Should be set to `0`, not kept empty.

tools.sh
cfergeau: > Yes, I also thought that but again these exposed variable we should document so user can look beforehand

The latest version of the PR is missing the documentation.

ci.sh
```diff
@@ -19,7 +19,7 @@ cat > /tmp/ignoretests.txt << EOF
 EOF
 
 ./shellcheck.sh
-./snc.sh
+SNC_GENERATE_MACOS_BUNDLE=0 ./snc.sh
```
cfergeau, line 22: Dunno if we can achieve the downgrade using public repos (assuming this runs on rhel9).

cfergeau (approved):

praveenkumar:
/hold

CI still try to create mac bundle but it shouldn't.

cfergeau (approved):

ci.sh
```diff
@@ -30,7 +30,7 @@ sudo mv /tmp/os-test/openshift-tests /usr/local/bin/
 
 # Run createdisk script
 export CRC_ZSTD_EXTRA_FLAGS="-10 --long"
-./createdisk.sh crc-tmp-install-data
+SNC_GENERATE_MACOS_BUNDLE=0 ./createdisk.sh crc-tmp-install-data
```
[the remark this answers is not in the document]
praveenkumar: @cfergeau I am thinking around https://github.com/crc-org/snc/pull/726#discussion_r1205310183 one if we can stop using internal repo if our CI images are migrated to RHEL-9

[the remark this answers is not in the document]
praveenkumar: So I am going ahead with making change in the `ci_microshift.sh` script also. Since our CI nodes are still not on RHEL-9, which might take some time. Once we move the CI nodes to RHEL-9 then might be reconsider it.

cfergeau (changes requested):
https://github.com/crc-org/snc/pull/726#discussion_r1202239045  is still pending, apart from this looks fine to me!

README.md
```diff
@@ -79,4 +79,12 @@ Host bootstrap
     UserKnownHostsFile /dev/null
 ```
 
+## ENV
+
```
cfergeau, line 83: -> Environment Variables

README.md
```diff
@@ -79,4 +79,12 @@ Host bootstrap
...
+Following environment variables can be used to change default value of bundle generation.
```
cfergeau, line 84: "The following environment variables can be used to change the default values of bundle generation"

cfergeau (approved):

PR #740, opened 2023-06-16
Remove lvm system device file

praveenkumar:
As of now this file contain the device name and disk ID associated with a physical volume identifier (PVI) and we noticed that device name is different for different hypervisor like for hyperV it is `/dev/sda` and for libvirt it is `/dev/vda` [0].

In this PR we follow `man lvmdevices` which states remove of this file means lvm will not use a devices file.
```
The  LVM  devices file lists devices that lvm can use.
The default file is /etc/lvm/devices/system.devices, and
the lvmdevices(8) command is used to add or remove device
entries.  If the file does not exist, or if lvm.conf includes
use_devicesfile=0, then lvm will not use a devices file.
```

With this patch I don't see any degrade in boot time
```
<=== with the patch ===>
INFO CRC instance is running with IP 127.0.0.1
INFO CRC VM is running
^C
real	0m13.727s
user	0m0.142s

<=== without the patch ===>
INFO CRC instance is running with IP 127.0.0.1
INFO CRC VM is running
INFO Updating authorized keys...
^C

real	0m14.216s
user	0m0.158s

```

[0] https://www.baeldung.com/linux/vda-vs-sda

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -81,6 +81,10 @@ EOF
     # Copy the sample microshift config and update the base domain with crc base domain
     ${SSH} core@${VM_IP} -- sudo cp /etc/microshift/config.yaml.default /etc/microshift/config.yaml
     ${SSH} core@${VM_IP} -- "sudo sed -i 's/#baseDomain: .*/baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}/g' /etc/microshift/config.yaml"
+    # Remove the lvm system.device file since it have diskID and deviceName which changes
+    # for different hypervisor and as per `man lvmdevices` if the file does not exist, or if lvm.conf
+    # includes use_devicesfile=0, then lvm will not use a devices file.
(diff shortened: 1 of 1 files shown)


cfergeau:
> FYI our crc disk resize code, older microshift bundle is not going to work on windows and only bundle with this change going to work.

Define "not going to work"? On Windows, it won't be possible to resize the disk image, but apart from this, the bundle will work fine? Or there will be bigger issues with older bundles/newer crc on Windows?

gbraad:
the comment is unnecessary. no need to refer to other bundlesnhere. this comment would be better suited for the CRC PR

praveenkumar:
> Define "not going to work"? On Windows, it won't be possible to resize the disk image, but apart from this, the bundle will work fine? Or will there be bigger issues with older bundles/newer crc on Windows?

It will work on windows without failure if we have addition changes in the crc codebase.

gbraad (approved):

PR #637, opened 2023-01-04
macos: Install older kernel

cfergeau:
Because of https://github.com/crc-org/vfkit/issues/11 we need to use an 
older kernel with our macOS bundles. This is fixed in macOS 13, but we want
to keep compatibility with macOS 12 for a while.
However, we want to keep using the latest kernel for linux and Windows.
This PR starts by refactoring the bundle generation. Before this PR, the bundle generation process is roughly:
- do various modifications in the VM
- shut down the VM
- use the same VM image for all 3 platforms

After this PR, it will become:
- do various modifications in the VM
- shut it down
- generate the linux/Windows bundles
- restart the VM
- downgrade the kernel
- shut the VM down
- generate the macOS bundle

This makes the bundle generation slower as making the VM image
sparse/converting it/... can be slow, but this allows to keep compatibility
with macOS 12.

This PR uses the latest 5.18 fedora kernel on M1, and latest 5.19 fedora kernel
on Intel mac.

PR #749, opened 2023-06-28
Downgrade kernel only for mac

praveenkumar:
68c6383 moved the kernel changes from an aarch64-only block to a block which is run when the macOS bundle is enabled. However, the kernel downgrade is done before any bundle is generated, so the kernel will be downgraded for all our bundles, linux, macos, windows.

This PR fixes this and use https://github.com/crc-org/snc/pull/637 logic to only downgrade the kernel for macOS.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -112,6 +112,20 @@ function create_qemu_image {
     rm -fr ${destDir}/${base}
 }
 
+function create_bundle_qemu_image() {
+  local libvirtDestDir="$1"
+  local VM_PREFIX="$2"
+  local VM_NAME="$3"
+
+  if [ "${BUNDLE_TYPE}" != "microshift" ]; then
+    create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
+    mv "${libvirtDestDir}/${VM_NAME}" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
+  else
+    create_qemu_image "$libvirtDestDir" "${VM_NAME}.qcow2" "microshift.qcow2"
+    mv "${libvirtDestDir}/microshift.qcow2" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
+  fi
+}
+
 function update_json_description {
     local srcDir=$1
     local destDir=$2
@@ -431,3 +445,4 @@ function remove_pull_secret_from_disk() {
 	;;
     esac
 }
+
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -138,33 +138,6 @@ fi
 
 cleanup_vm_image ${VM_NAME} ${VM_IP}
 
-# Only used for macOS bundle generation
-if [ "${SNC_GENERATE_MACOS_BUNDLE}" != "0" ]; then
-    if [ ${BUNDLE_TYPE} != "okd" ]; then
-        # workaround https://github.com/crc-org/vfkit/issues/11 on macOS 12
-        downgrade_rhel9_kernel ${VM_IP}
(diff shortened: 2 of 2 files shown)


praveenkumar:
/hold 

bundles need to generated downstream and then tested on each platform.

cfergeau (commented):
Looks good to me, there's an extra blank line at the end of the file. This could be made even more similar to the code in the podman branch.
CI is green.

createdisk.sh
```diff
@@ -181,26 +154,20 @@ libvirtDestDir="${destDirPrefix}_libvirt_${destDirSuffix}"
...
-fi
+function copy_qemu_image {
+    if [ $BUNDLE_TYPE != "microshift" ]; then
+        create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
+        mv "${libvirtDestDir}/${VM_NAME}" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
+    else
+        sparsify_lvm "${libvirtDestDir}"
+    fi
+}
+copy_qemu_image
```
cfergeau, line 165: There's already a `create_qemu_image` which copies and sparsifies, please don't add an additional `copy_qemu_image` which calls `create_qemu_image` and also sparsifies in other cases. Let's use this opportunity to group all this code :)

cfergeau (changes requested):

createdisk.sh
praveenkumar: @cfergeau you mean `create_qemu_image` should also work with LVM image type?

gbraad (approved):

createdisk-library.sh
```diff
@@ -275,6 +275,9 @@ function generate_vfkit_bundle {
 
     generate_macos_bundle "vfkit" "$@"
 
+    create_qemu_image "$srcDir" "${SNC_PRODUCT_NAME}.qcow2" "microshift"
+    mv "${srcDir}/microshift" "${srcDir}/${SNC_PRODUCT_NAME}.qcow2"
```
cfergeau, line 279: Is this correct for okd/openshift?

praveenkumar: @cfergeau (Good catch) no, It is only going to work with microshift, we might need another helper which execute this with `BUNDLE_TYPE` something `copy_qemu_image` like https://github.com/crc-org/snc/pull/749#discussion_r1245390062 or do you have a better helper name in mind?

cfergeau: Maybe it's `create_qemu_image` which needs to be renamed to be more descriptive?

praveenkumar: or may be we should handle that bundle type as part of `create_qemu_image` function.

cfergeau: `create_qemu_image` is relatively generic, adding some bundle type checks there would remove some of this genericity.

praveenkumar: > `create_qemu_image` is relatively generic, adding some bundle type checks there would remove some of this genericity.

@cfergeau agree, what we need a helper which encapsulate following code, because if we rename `create_qemu_image` and create 2 separate helper `create_qemu_image_snc` and `create_qemu_image_microshift` and call `create_qemu_image` internally for both helper then also this conditional block is required.
```bash
if [ $BUNDLE_TYPE != "microshift" ]; then
    create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
else  
    create_qemu_image "$libvirtDestDir" "${VM_NAME}.qcow2" "microshift"
    mv "${libvirtDestDir}/microshift" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
fi  
```

cfergeau (approved):

PR #750, opened 2023-07-03
Createdisk: update sparsify helper to use for microshift bundle also

praveenkumar:
During microshift bundle creation d90d53d22d51fdf406e265ac964206f46d2badd0 we created `sparsify_lvm` to move fast to have microshift bundle sooner. With this PR we are going to use sparsify helper for microshift bundle and remove the `sparsify_lvm` helper.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -46,6 +46,10 @@ function sparsify {
     export LIBGUESTFS_BACKEND=direct
     # Check which partition is labeled as `root`
     partition=$(${VIRT_FILESYSTEMS} -a $baseDir/$srcFile -l --partitions | sort -rk4 -n | sed -n 1p | cut -f1 -d' ')
+    # check if the base image has the lvm named as `rhel/root`
+    if ${VIRT_FILESYSTEMS} --lvs -a ${baseDir}/${srcFile}  | grep -q "rhel/root"; then
+      partition="/dev/rhel/root"
+    fi
 
     # https://bugzilla.redhat.com/show_bug.cgi?id=1837765
     export LIBGUESTFS_MEMSIZE=2048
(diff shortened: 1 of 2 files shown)


createdisk.sh
```diff
@@ -185,7 +185,8 @@ if [ $BUNDLE_TYPE != "microshift" ]; then
     create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
     mv "${libvirtDestDir}/${VM_NAME}" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
 else
-    sparsify_lvm "${libvirtDestDir}"
+    create_qemu_image "$libvirtDestDir" "${VM_NAME}.qcow2" "microshift"
+    mv "${libvirtDestDir}/microshift" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
```
cfergeau, line 189: if `VM_NAME` is different from `SNC_PRODUCT_NAME`, it's probably possible to directly pass `SNC_PRODUCT_NAME.qcow2` to `create_qemu_image`

createdisk-library.sh
```diff
@@ -87,15 +91,18 @@ function create_qemu_image {
 
     if [ -f /var/lib/libvirt/images/${overlay} ]; then
       sudo cp /var/lib/libvirt/images/${overlay} ${destDir}
```
cfergeau, line 93: I'd add a comment either here or before `local overlay=$3` to indicate that `${overlay}` won't exist in some cases, for example when using microshift.

cfergeau (commented):
Nice to see these 2 codepaths can be so similar!

createdisk.sh
praveenkumar: @cfergeau in case of microshift it is same https://github.com/crc-org/snc/pull/750/files#diff-030b04caaa878905ffc5b7caa7e11fbae90df58669834ebe42fbbe907bfb6db7R41-R46 so I used `${VM_NAME}`

createdisk-library.sh
praveenkumar: @cfergeau updated.

PR #754, opened 2023-07-05
Drop the use of internal repo to download kernel package for downgrade 

praveenkumar:
After upgrading our CI to RHEL-9, We can use https://access.redhat.com/solutions/134403
to downgrade the kernel instead of depending on the internal repo which also affect us to
perform generate the macOS bundle on CI.


praveenkumar:
/hold

Dependency: 
- https://github.com/openshift/release/pull/40922
- https://github.com/ironcladlou/openshift4-libvirt-gcp/pull/46/

createdisk-library.sh
```diff
@@ -219,9 +219,10 @@ function install_additional_packages() {
 function downgrade_rhel9_kernel {
     local vm_ip=$1
     local pkgDir=$(mktemp -d tmp-rpmXXX)
+    local ker_version="5.14.0-70.30.1.el9_0.${ARCH}"
```
cfergeau, line 222: We can't hardcode the kernel version, we'll forget to update it :)

praveenkumar: @cfergeau on flip side we know which kernel version we are downgrading to. what if an `el9_0` kernel again become unsupported for older mac because it updated.

cfergeau: **if** there are issues with newer `el9_0` kernels, then we can hardcode it. Before this change, we were using latest 9.0 kernel. if there's an upgrade to a 9.0 kernel, this means bugs/CVEs fixed, so without a strong reason for it, it's much preferrable to use the latest.

praveenkumar: @cfergeau Alright so I used something like this now, which should provide us latest version of kernel for `el9_0`.
```
$ sudo dnf repoquery-na --quiet --queryformat '%{version}.%{release}.%{arch}' kernel.x86_64
5.14.0.162.12.1.el9_1.x86_64
5.14.0.162.18.1.el9_1.x86_64
5.14.0.162.22.2.el9_1.x86_64
5.14.0.162.23.1.el9_1.x86_64
5.14.0.162.6.1.el9_1.x86_64
5.14.0.284.11.1.el9_2.x86_64
5.14.0.284.18.1.el9_2.x86_64
5.14.0.70.13.1.el9_0.x86_64
5.14.0.70.17.1.el9_0.x86_64
5.14.0.70.22.1.el9_0.x86_64
5.14.0.70.26.1.el9_0.x86_64
5.14.0.70.30.1.el9_0.x86_64

$ $ sudo dnf --quiet repoquery-na --queryformat '%{version}.%{release}.%{arch}' kernel.x86_64 | grep el9_0 | tail -n 1
5.14.0.70.30.1.el9_0.x86_64
```

createdisk-library.sh
```diff
@@ -219,9 +219,10 @@ function install_additional_packages() {
 function downgrade_rhel9_kernel {
     local vm_ip=$1
     local pkgDir=$(mktemp -d tmp-rpmXXX)
+    ker_version=$(sudo dnf --quiet repoquery-na --queryformat '%{version}.%{release}.%{arch}' kernel.${ARCH} | grep el9_0 | tail -n 1)
```
cfergeau, line 222: Maybe "kernel_version" ?

cfergeau (approved):

PR #756, opened 2023-07-06
microshift: Drop RH internal 'brew' requirement

praveenkumar:
microshift is officially released in RH repos, and microshift.sh already requires a subscription, so we can use `yum download` instead of `brew download-build`.
I've checked that setting `MICROSHIFT_NVR` works for 4.13 releases.


praveenkumar:
Once this is merged I will work on https://github.com/openshift/release/pull/38200 one again.

PR #759, opened 2023-07-06
Rename route_controller.yaml.in to router-controller.yaml.in

praveenkumar:


cfergeau (commented):
Maybe document that:
- this is more consistent with the naming of the container image
- this allows to workaround a problem with bundle v4.13.xx which had an invalid route_controller.json file

praveenkumar:
@cfergeau updated the commit log.

cfergeau (approved):
And I just noticed the new name is wrong in the shortlog :-/ 
`Rename route_controller.yaml.in to router-controller.yaml.in` (it's 'routes-controller.yaml.in')

Apart from that, ACK.

PR #761, opened 2023-07-06
Use sudo for virsh commands as part of preflight checks

praveenkumar:
openshift/okd bundles use `qemu+tcp` connection so that doesn't need the 'sudo' to get the details about network, uri and capabilities but for microshift bundle we use same preflight checks and use `qemu:///system` uri which need `sudo` to get those details otherwise the CI fails with following error:
```
+ '[' microshift == okd ']'
+ virsh -c qemu:///system uri
error: failed to connect to the hypervisor
```

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -62,18 +62,18 @@ function run_preflight_checks() {
 	fi
 
         # check if we can connect to ${LIBVIRT_URI}
-        if ! virsh -c ${LIBVIRT_URI} uri >/dev/null; then
+        if ! sudo virsh -c ${LIBVIRT_URI} uri >/dev/null; then
               preflight_failure  "libvirtd is not listening for ${LIBVIRT_URI}, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
         fi
 
-	if ! virsh -c ${LIBVIRT_URI} net-info default &> /dev/null; then
+	if ! sudo virsh -c ${LIBVIRT_URI} net-info default &> /dev/null; then
 		echo "Installing libvirt default network configuration"
 		sudo dnf install -y libvirt-daemon-config-network || exit 1
 	fi
 	echo "default libvirt network is available"
 
 	#Check if default libvirt network is Active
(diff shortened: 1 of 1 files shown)


praveenkumar:
It should fix https://console.build01.ci.openshift.org/k8s/ns/ci-op-2n5ihvjm/pods/e2e-microshift-code-ready-snc-microshift-test/logs ( which is job run from https://github.com/openshift/release/pull/38200 )

cfergeau:
An alternative is to make sure the CI user is part of the libvirt group. This avoids the need for sudo.

praveenkumar:
> An alternative is to make sure the CI user is part of the libvirt group. This avoids the need for sudo.

Yes, either we make this change here (snc side) or on CI side but if we make it here then some user try to create microshift bundle then it prompt for password and they don't have to add the user to libvirt group. We already using `sudo` https://github.com/crc-org/snc/blob/master/microshift.sh#L138 when creating the VM so IMO it is better to have this change here.

gbraad:
> either we make this change here (snc side) or on CI side but if we make it here

This would both be possible, and perhaps a good fallback. However, when a user is added to the group, you need to `newgrp` before it takes effect.

cfergeau:
> > An alternative is to make sure the CI user is part of the libvirt group. This avoids the need for sudo.
> 
> Yes, either we make this change here (snc side) or on CI side but if we make it here then some user try to create microshift bundle then it prompt for password and they don't have to add the user to libvirt group. We already using `sudo` https://github.com/crc-org/snc/blob/master/microshift.sh#L138 when creating the VM so IMO it is better to have this change here.

The less `sudo` usage  we have/the more fine grained we are in giving the CI user additional permissions, the better. Ideally, our CI nodes would not even need sudo at all. To push this to the extreme, why do we bother when we could just run `sudo microshift.sh` ;)

Looks like `sudo virt-install` could also be replaced with `virt-install --connect qemu:///system` if the user is in the `libvirt` group.
This leaves the access to `/var/lib/libvirt/images` which is owned  by root. If we wanted to go very far with sudo removal, maybe `virsh download` would allow to replace it, but that's definitely some hypothetical long term change ;)

cfergeau (approved):
Reluctant approval to unblock CI work, but to me, most of the time `sudo` is not a good solution, it's just a quick and easy workaround. I'd aim at having less `sudo` calls, not more.

PR #769, opened 2023-07-07
Rename router-controller.yaml.in to routes-controller.yaml.in

praveenkumar:
With 3edc3ceb4db9f9ba81eb7e5d80467acbc34d4d89 we want to make change the file with `routes-controller.yaml.in` but by mistake file is renamed to `router-controller.yaml.in` and the uses of the file named correctly with `routes-*`. This PR fix that file rename mistake.


praveenkumar:
/hold

Till CI have bundle created.

cfergeau (approved):

PR #776, opened 2023-07-12
ci_microshift: Just undefine the crc domain

praveenkumar:
During creation of bundle `crc` vm is already in shutdown state so `virsh destroy` commands fails. We only need to undefine the VM.

### DIFF
--- a/ci_microshift.sh
+++ b/ci_microshift.sh
@@ -14,7 +14,6 @@ export CRC_ZSTD_EXTRA_FLAGS="-10"
 
 # Delete the crc domain which created by snc so it can created
 # for crc test
-sudo virsh destroy crc
 sudo virsh undefine crc
 
 git clone https://github.com/crc-org/crc.git


ci_microshift.sh
```diff
@@ -14,7 +14,6 @@ export CRC_ZSTD_EXTRA_FLAGS="-10"
 
 # Delete the crc domain which created by snc so it can created
 # for crc test
-sudo virsh destroy crc
```
anjannath, line 17: did you consider ignoring the error (`|| true`)?, so in case of an unsuccessful run and if the VM stays running, it'll stop that

praveenkumar: if it unsuccessful then bundle will be not generated so there is no point to run next steps.

anjannath: got it, if creatdisk.sh fails then bundle will not be generated and this steps will not even run

PR #779, opened 2023-07-18
Build crc-dnsmasq and crc-route-controller image even KAO image already present

praveenkumar:
Recently there was issue with brew and jobs are stuck to `free` state for a day and after that it succeed but our jenkins jobs fails to wait that long and when we rebuild the job since KAO image already built on the brew it will not fetch the repo and routes and dnsmasq images never built for that specific version of openshift.

This PR make sure we always built those images as long as the `From` section from Dockerfile changes.

### DIFF
--- a/build-patched-kao-kcmo-images.sh
+++ b/build-patched-kao-kcmo-images.sh
@@ -139,11 +139,14 @@ function update_base_image() {
     rhpkg clone containers/${brew_repo}
     pushd ${brew_repo}
     git checkout --track origin/crc-1-rhel-8
-    sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
-    git add Dockerfile
-    git commit -m "Use OpenShift ${openshift_version} base image"
-    git push origin
-    rhpkg container-build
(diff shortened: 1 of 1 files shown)


build-patched-kao-kcmo-images.sh
```diff
@@ -139,11 +139,14 @@ function update_base_image() {
...
     pushd ${brew_repo}
     git checkout --track origin/crc-1-rhel-8
-    sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
-    git add Dockerfile
-    git commit -m "Use OpenShift ${openshift_version} base image"
-    git push origin
-    rhpkg container-build
+    base_image_of_repo=$(grep "^FROM openshift/ose-base" Dockerfile | sed 's/^FROM //')
+    if [ ${base_image} != ${base_image_of_repo} ]; then
+      sed -i "s!^FROM openshift/ose-base.*!FROM $base_image!" Dockerfile
```
cfergeau, line 144: you probably could do the `sed` unconditionally, and ask git if there were any changes to the work tree instead of getting `base_image_of_repo` and making a comparison.

PR #781, opened 2023-07-19
Enable linger for core user

praveenkumar:
For the podman-env setting we use an SSH connection to start the container. It seems that containers get killed after a timeout. The way it is handled in podman-machine side is to enable lingering for `core` user and this PR is suppose to do same.


podman_changes.ks
```diff
@@ -21,3 +21,5 @@ tee /etc/containers/registries.conf.d/999-podman-machine.conf <<EOF
 unqualified-search-registries=["docker.io"]
 EOF
 
+# Enable linger for core user to make sure podman socket work when user not logged in
+sudo touch /var/lib/systemd/linger/core
```
cfergeau, line 25: If the system has loginctl installed, can we use this?

praveenkumar: @cfergeau Can we use loginctl command in kickstart file?

cfergeau: You can use `sudo touch` in the kickstart file, is running `loginctl` any different?

gbraad:
How about rootful containers (using root) ? I actually haven't tested this, or looked into it, so curious how this is done.

podman_changes.ks
gbraad: `loginctl enable-linger core` is what I used to set this.

praveenkumar:
> How about rootful containers (using root) ? I actually haven't tested this, or looked into it, so curious how this is done.

On podman machine side I am not seeing the root user as part of linger. I also didn't test `root` user.

podman_changes.ks
praveenkumar: When using `loginctl enable-linger core` to kickstart file I am getting following error from kickstart post installation
```
kernel boot argument 'ks' detected. Did you want to use 'inst.ks' for the installer instead?
12:26:36 All Anaconda kernel boot arguments are now required to use 'inst.' prefix!
12:26:36 
############## Installer errors encountered during boot ##############
'ks' is deprecated and has been removed.
All usage of Anaconda boot arguments without 'inst.' prefix was removed. Please use inst.ks instead.

Running post-installation scripts
================================================================================
================================================================================
Error

  There was an error running the kickstart script at line 36.  This is a fatal
  error and installation will be aborted.  The details of this error are:
  
  success
  success
  [containers]
  netns="bridge"
  rootless_networking="cni"
  unqualified-search-registries=["docker.io"]
  Could not enable linger: No such file or directory
  

Press ENTER to exit: 
```
I think during the post installation there is no directory `/var/lib/systemd/linger/` so I am trying with following as part of post installation script of kickstart to see if that works.
```
mkdir -p /var/lib/systemd/linger
loginctl enable-linger core
```

praveenkumar: Tested following and installation succeed and core user have linger enabled.
```
mkdir -p /var/lib/systemd/linger
touch /var/lib/systemd/linger/core
```

gbraad: weird, but seems reported something similar a long time ago: https://tracker.pureos.net/T887

gbraad (approved):

PR #790, opened 2023-08-10
[release-4.14] CI: use MICROSHIFT_PRERELEASE for creating pre release bundles of microshift

praveenkumar:


ci_microshift.sh
```diff
@@ -5,7 +5,7 @@ set -exuo pipefail
 sudo yum install -y make golang
 
 ./shellcheck.sh
-./microshift.sh
+MICROSHIFT_PRERELEASE=yes ./microshift.sh
```
cfergeau, line 8: `MICROSHIFT_PRERELASE=1` is imo more common to enable an env var

PR #793, opened 2023-08-23
gen-bundle-image: refactor to accomodate microshift preset

praveenkumar:
This PR refactor around set_bundle_variables, generate_image and sign_bundle_files functions to make sure image can be generated for microshift bundle also.

### DIFF
--- a/gen-bundle-image.sh
+++ b/gen-bundle-image.sh
@@ -57,7 +57,7 @@ function generate_manifest {
    local preset=$2
    podman manifest rm ${preset}-bundle:${version} || true
    podman manifest create ${preset}-bundle:${version}
-   if [[ ${preset} = "podman" ]]; then
+   if [[ ${preset} != "okd" ]]; then
       podman manifest add ${preset}-bundle:${version} containers-storage:localhost/${preset}-bundle:darwin-arm64
    fi
    podman manifest add ${preset}-bundle:${version} containers-storage:localhost/${preset}-bundle:darwin-amd64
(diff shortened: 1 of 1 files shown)


gen-bundle-image.sh
```diff
@@ -8,22 +8,18 @@ function set_bundle_variables {
...
+   if [ ${PRESET} != 'okd' ]; then
+       vfkit_bundle_arm64=crc${bundlePreset}_vfkit_${version}_arm64.crcbundle
    fi
 
-   vfkit_bundle=crc${preset}_vfkit_${version}_amd64.crcbundle
-   libvirt_bundle=crc${preset}_libvirt_${version}_amd64.crcbundle
-   hyperv_bundle=crc${preset}_hyperv_${version}_amd64.crcbundle
+   vfkit_bundle=crc${bundlePreset}_vfkit_${version}_amd64.crcbundle
+   libvirt_bundle=crc${bundlePreset}_libvirt_${version}_amd64.crcbundle
+   hyperv_bundle=crc${bundlePreset}_hyperv_${version}_amd64.crcbundle
```
cfergeau, line 22: This looks related to supporting `microshift` in addition to `okd`, `podman` and `openshift`...

gen-bundle-image.sh
```diff
@@ -57,7 +53,7 @@ function generate_manifest {
    local preset=$2
    podman manifest rm ${preset}-bundle:${version} || true
    podman manifest create ${preset}-bundle:${version}
-   if [[ ${preset} = "podman" ]]; then
+   if [[ ${preset} != "okd" ]]; then
```
cfergeau, line 56: ... but this part seems to be fixing a slightly different issue, the arm64 bundle was only added to a manifest/signed for the podman bundle, even if arm64 bundles are also generated for openshift since there is this code before:
```
   if [[ ${preset} = "openshift" ]]; then
       vfkit_bundle_arm64=crc_vfkit_${version}_arm64.crcbundle
   fi
```

Can this be moved to a different commit, or at least described in the commit log?

cfergeau (changes requested):

gen-bundle-image.sh
praveenkumar: Separate commit added for it.

gen-bundle-image.sh
```diff
@@ -69,7 +69,7 @@ function generate_manifest {
 function sign_bundle_files {
   local preset=$1
   rm -fr *.sig
-  if [[ ${preset} = "podman" ]]; then
+  if [[ ${preset} != "okd" ]]; then
```
cfergeau, line 68: Fwiw, I would change this logic to check if `vfkit_bundle_arm64` is empty or not, this way we don't have multiple places in this script which need to know which bundles support arm64, we set `vfkit_bundle_arm64` according to preset, and the rest of the code does the right thing once this is set or not.

cfergeau (approved):

gen-bundle-image.sh
praveenkumar: @cfergeau is it still unresolved? or this comment is for commit log change?

PR #797, opened 2023-08-25
snc-library: Fix app route for microshift preset

praveenkumar:
In microshift we don't configure app route different than base domain like we can do in OCP using ingress configuration, so app route for microshift become `apps.crc.testing` but for OCP it is `apps-crc.testing`. This patch make sure it is updated to the bundle metadata correctly.

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -143,11 +143,13 @@ function create_json_description {
             | ${JQ} ".clusterInfo.openshiftVersion = \"${OPENSHIFT_RELEASE_VERSION}\"" \
             | ${JQ} ".clusterInfo.clusterName = \"${SNC_PRODUCT_NAME}\"" \
             | ${JQ} ".clusterInfo.baseDomain = \"${BASE_DOMAIN}\"" \
-            | ${JQ} ".clusterInfo.appsDomain = \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
+            | ${JQ} ".clusterInfo.appsDomain = \"apps.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
     if [ ${bundle_type} == "snc" ] || [ ${bundle_type} == "okd" ]; then
         openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
         tmp=$(mktemp)
-        cat ${INSTALL_DIR}/crc-bundle-info.json | ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\"" \
+        cat ${INSTALL_DIR}/crc-bundle-info.json \
+            | ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\"" \
+            | ${JQ} ".clusterInfo.appsDomain = \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" \
             > ${tmp} && mv ${tmp} ${INSTALL_DIR}/crc-bundle-info.json
     fi
 }


snc-library.sh
```diff
@@ -143,11 +143,12 @@ function create_json_description {
...
             | ${JQ} ".clusterInfo.clusterName = \"${SNC_PRODUCT_NAME}\"" \
             | ${JQ} ".clusterInfo.baseDomain = \"${BASE_DOMAIN}\"" \
-            | ${JQ} ".clusterInfo.appsDomain = \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
+            | ${JQ} ".clusterInfo.appsDomain = \"apps.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
     if [ ${bundle_type} == "snc" ] || [ ${bundle_type} == "okd" ]; then
         openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
         tmp=$(mktemp)
-        cat ${INSTALL_DIR}/crc-bundle-info.json | ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\"" \
+        cat ${INSTALL_DIR}/crc-bundle-info.json | \
+            ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\" | .clusterInfo.appsDomain |= \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" \
```
cfergeau, line 151: Don't you need a `${JQ}` after the `|`? And can you split this line in 2?

cfergeau (commented):
There were some DNS recursion issues with the default apps domain, which why we changed it to `apps-crc`. Hopefully these issues are gone now :)

snc-library.sh
praveenkumar: > Don't you need a ${JQ} after the |?

No 

> And can you split this line in 2?

`".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\" | .clusterInfo.appsDomain |= \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\""` is a single JQ query so can't split in 2 different line.

The CI is failed not during bundle creation but because of https://github.com/crc-org/crc/pull/3806 one

cfergeau: The block above is using `JQ` after every pipe, I'd prefer we keep this consistent.

praveenkumar: @cfergeau updated.

praveenkumar:
The error message from microshift CI is expected and depend on https://github.com/crc-org/crc/pull/3806
```
level=info msg="Downloading bundle: crc_microshift_libvirt_4.13.9_amd64.pr797.crcbundle..."
level=info msg="Uncompressing crc_microshift_libvirt_4.13.9_amd64.pr797.crcbundle"
unexpected bundle, it must have .apps-crc.testing apps domain
```

snc-library.sh
```diff
@@ -143,11 +143,13 @@ function create_json_description {
...
             | ${JQ} ".clusterInfo.baseDomain = \"${BASE_DOMAIN}\"" \
-            | ${JQ} ".clusterInfo.appsDomain = \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
+            | ${JQ} ".clusterInfo.appsDomain = \"apps.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" >${INSTALL_DIR}/crc-bundle-info.json
     if [ ${bundle_type} == "snc" ] || [ ${bundle_type} == "okd" ]; then
         openshiftInstallerVersion=$(${OPENSHIFT_INSTALL} version)
         tmp=$(mktemp)
-        cat ${INSTALL_DIR}/crc-bundle-info.json | ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\"" \
+        cat ${INSTALL_DIR}/crc-bundle-info.json \
+            | ${JQ} ".buildInfo.openshiftInstallerVersion = \"${openshiftInstallerVersion}\"" \
+            | ${JQ} ".clusterInfo.appsDomain |= \"apps-${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"" \
```
cfergeau, line 152: is |= mandatory? In my testing `=` seems to be enough? (`echo {} | jq ".clusterInfo.appsDomain=\"apps.crc.testing\"" |  jq ".clusterInfo.appsDomain=\"apps-crc.testing\""`

praveenkumar: As per man page of `jq` . For simple operation like this one `=` and `|=` works as same but update assignment is more precise for updating the existing value by keeping the same json structure. 
```
Update-assignment: |=
       This  is the "update" operator ´|=´. It takes a filter on the right-hand side and works out the new value for the property of . being assigned to by running the old value through this expression. For instance, (.foo, .bar) |= .+1 will build an object with the "foo" field set to the input´s "foo" plus 1, and the "bar" field set to the input´s "bar" plus 1.
```

cfergeau: We simply want to overwrite the value, I don't think `|=` is needed, and it's not used anywhere else

praveenkumar: > We simply want to overwrite the value, I don't think |= is needed

As I said with current situation `=` and `|=` works same.

>and it's not used anywhere else

Doesn't mean we can't use it now. If that is adding confusion and block PR merge then I will update it.

praveenkumar: Updated.

PR #802, opened 2023-09-06
Use GPT partition table instead msdos

praveenkumar:
This PR adds `clearpart` option `--disklabel` to use `gpt` as default disk label and to allow this we also need to add a partition of `biosboot` so that it will not error out with following error.
```
your bios-based system needs a special partition to boot from a gpt disk label
```

- https://pykickstart.readthedocs.io/en/latest/kickstart-docs.html#clearpart
- https://access.redhat.com/solutions/3370891

### DIFF
--- a/microshift.sh
+++ b/microshift.sh
@@ -103,6 +103,7 @@ EOF
     sed -i "/--bootproto=dhcp/a\network --hostname=api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}" scripts/image-builder/config/kickstart.ks.template
     sed -i 's/clearpart --all --initlabel/clearpart --all --disklabel gpt/g' scripts/image-builder/config/kickstart.ks.template
     sed -i "/clearpart --all/a\part biosboot --fstype=biosboot --size=1" scripts/image-builder/config/kickstart.ks.template
+    sed -i '$i\grub2-install --target=i386-pc /dev/vda' scripts/image-builder/config/kickstart.ks.template
     sed -i '$e cat podman_changes.ks' scripts/image-builder/config/kickstart.ks.template
     scripts/image-builder/cleanup.sh -full
     # The home dir and files must have read permissions to group
--- a/microshift.sh
+++ b/microshift.sh
@@ -158,4 +158,5 @@ sudo virt-install \
     --cdrom /var/lib/libvirt/images/microshift-installer.iso \
     --events on_reboot=restart \
     --autoconsole none \
+    --boot uefi \
     --wait 5
--- a/microshift.sh
+++ b/microshift.sh
@@ -101,6 +101,8 @@ version = "*"
 EOF
     sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
     sed -i "/--bootproto=dhcp/a\network --hostname=api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}" scripts/image-builder/config/kickstart.ks.template
+    sed -i 's/clearpart --all --initlabel/clearpart --all --disklabel gpt/g' scripts/image-builder/config/kickstart.ks.template
+    sed -i "/clearpart --all/a\part biosboot --fstype=biosboot --size=1" scripts/image-builder/config/kickstart.ks.template
     sed -i '$e cat podman_changes.ks' scripts/image-builder/config/kickstart.ks.template
     scripts/image-builder/cleanup.sh -full
     # The home dir and files must have read permissions to group
--- a/ci_microshift.sh
+++ b/ci_microshift.sh
@@ -14,7 +14,7 @@ export CRC_ZSTD_EXTRA_FLAGS="-10"
 
 # Delete the crc domain which created by snc so it can created
 # for crc test
-sudo virsh undefine crc
+sudo virsh undefine crc --nvram
 
 git clone https://github.com/crc-org/crc.git
 pushd crc


praveenkumar:
/hold

need to test it on arm64 also.

microshift.sh
```diff
@@ -101,6 +101,8 @@ version = "*"
 EOF
     sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
     sed -i "/--bootproto=dhcp/a\network --hostname=api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}" scripts/image-builder/config/kickstart.ks.template
+    sed -i 's/clearpart --all --initlabel/clearpart --all --initlabel --disklabel gpt/g' scripts/image-builder/config/kickstart.ks.template
+    sed -i "/clearpart --all/a\part biosboot --fstype=biosboot --size=1" scripts/image-builder/config/kickstart.ks.template
```
cfergeau, line 105: Why the `\` before `part`?

cfergeau (commented):
This seems like a generic change ,this can go together with an upstream (microshift) issue/PR

microshift.sh
```diff
@@ -101,6 +101,8 @@ version = "*"
 EOF
     sed -i 's/redhat/core/g' scripts/image-builder/config/kickstart.ks.template
     sed -i "/--bootproto=dhcp/a\network --hostname=api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}" scripts/image-builder/config/kickstart.ks.template
+    sed -i 's/clearpart --all --initlabel/clearpart --all --initlabel --disklabel gpt/g' scripts/image-builder/config/kickstart.ks.template
```
cfergeau, line 104: is `--initlabel` still required?
> `--initlabel`: Initializes the disk label to the default for your architecture
> `--disklabel DISKLABEL` Set the default disklabel to use.

The latter seems to override the former?

microshift.sh
praveenkumar: Just to explicitly separate out `/a` (append part).

microshift.sh
praveenkumar: Yes as per docs it looks like but I did check one of the kickstart file which uses both the option, can try without `initlabel` and see if that works.

- https://dl.acronis.com/u/storage2/html/AcronisStorage_2_installation_pxe_guide_en-US/creating-a-kickstart-file/kickstart-file-example.html

cfergeau:
Before this PR (4.13.9 bundle):
```
GPT fdisk (gdisk) version 1.0.7

Partition table scan:
  MBR: MBR only
  BSD: not present
  APM: not present
  GPT: not present


***************************************************************
Found invalid GPT and valid MBR; converting MBR to GPT format
in memory. 
***************************************************************


Warning! Secondary partition table overlaps the last partition by
33 blocks!
You will need to delete this partition or resize it in another utility.
Disk /dev/vda: 65011712 sectors, 31.0 GiB
Sector size (logical/physical): 512/512 bytes
Disk identifier (GUID): C145A918-55D2-491B-AB5E-6B07D9C2D5BA
Partition table holds up to 128 entries
Main partition table begins at sector 2 and ends at sector 33
First usable sector is 34, last usable sector is 65011678
Partitions will be aligned on 2048-sector boundaries
Total free space is 2014 sectors (1007.0 KiB)

Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048         1640447   800.0 MiB   8300  Linux filesystem
   2         1640448         2050047   200.0 MiB   0700  Microsoft basic data
   3         2050048        65011711   30.0 GiB    8E00  Linux LVM
```

After this PR:
```
$ sudo gdisk -l /dev/vda
GPT fdisk (gdisk) version 1.0.7

Partition table scan:
  MBR: protective
  BSD: not present
  APM: not present
  GPT: present

Found valid GPT with protective MBR; using GPT.
Disk /dev/vda: 65011712 sectors, 31.0 GiB
Sector size (logical/physical): 512/512 bytes
Disk identifier (GUID): 090A7736-7EEE-45AC-9187-9AE9D7291080
Partition table holds up to 128 entries
Main partition table begins at sector 2 and ends at sector 33
First usable sector is 34, last usable sector is 65011678
Partitions will be aligned on 2048-sector boundaries
Total free space is 4029 sectors (2.0 MiB)

Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048            4095   1024.0 KiB  EF02  
   2            4096         1642495   800.0 MiB   8300  
   3         1642496         2052095   200.0 MiB   EF00  
   4         2052096        65009663   30.0 GiB    8E00  
```

The partition layout and sizes look fine, and the partition table is GPT. Not sure why we lost the code names, and the code for the 200 MiB partition is different, but this might not be a problem.

gbraad:
The Podman bundle (FCOS) works as a HyperV Gen2 VM
It might be something specific to RHCOS

microshift.sh
```diff
@@ -103,6 +103,7 @@ EOF
     sed -i "/--bootproto=dhcp/a\network --hostname=api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}" scripts/image-builder/config/kickstart.ks.template
     sed -i 's/clearpart --all --initlabel/clearpart --all --disklabel gpt/g' scripts/image-builder/config/kickstart.ks.template
     sed -i "/clearpart --all/a\part biosboot --fstype=biosboot --size=1" scripts/image-builder/config/kickstart.ks.template
+    sed -i '$e grub2-install --target=i386-pc /dev/vda' scripts/image-builder/config/kickstart.ks.template
```
cfergeau, line 106: Have you tried using https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/performing_an_advanced_rhel_9_installation/kickstart-commands-and-options-reference_installing-rhel-as-an-experienced-user#bootloader-required_kickstart-commands-for-handling-storage ?

gbraad:
Works as gen1 and gen2

gbraad (approved):

microshift.sh
cfergeau: Praveen tried adding `bootloader` to the kickstart file, but for some reason this was not equivalent to manuallly running `grub2-install`.

cfergeau (changes requested):
https://github.com/crc-org/snc/pull/802#pullrequestreview-1612784700

praveenkumar:
> This seems like a generic change ,this can go together with an upstream (microshift) issue/PR

@cfergeau https://github.com/openshift/microshift/pull/2331 created PR for generic part, installation of bootloader doesn't feel generic because on microshift side the outcome is an ISO which can work on but uefi and legacy mode.

cfergeau:
> @cfergeau [openshift/microshift#2331](https://github.com/openshift/microshift/pull/2331) created PR for generic part,

Can you add this to the commit logs? We will need to revert some of these commits when/if this is merged.

praveenkumar:
> Can you add this to the commit logs? We will need to revert some of these commits when/if this is merged.

updated the commit logs with upstream link.

microshift.sh
```diff
@@ -103,6 +103,7 @@ EOF
     sed -i "/--bootproto=dhcp/a\network --hostname=api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}" scripts/image-builder/config/kickstart.ks.template
     sed -i 's/clearpart --all --initlabel/clearpart --all --disklabel gpt/g' scripts/image-builder/config/kickstart.ks.template
     sed -i "/clearpart --all/a\part biosboot --fstype=biosboot --size=1" scripts/image-builder/config/kickstart.ks.template
+    sed -i '$i\grub2-install --target=i386-pc /dev/vda' scripts/image-builder/config/kickstart.ks.template
```
cfergeau, line 106: Still disappointed that the KS `bootloader` directive does not work as expected :)

cfergeau (commented):
The log for ` microshift: Boot the ISO in uefi mode` is hard to understand, I'd phrase it as "If we don't boot the ISO in uefi mode, then the UEFI firmware files won't get installed in the generated qcow2 image. This would prevent this image from booting with UEFI."

Apart from this, looks good to me.

PR #812, opened 2023-11-02
Switch from libvirt IPI to UPI with SNO capability

praveenkumar:
This PR is going to use https://docs.openshift.com/container-platform/4.14/installing/installing_sno/install-sno-installing-sno.html#install-sno-installing-sno-manually
docs to install single node openshift cluster manually.
It does the following.
- Update the install-config as per doc
- Create the manifest and then single node ignition using this
  install-config
- Generate the installation iso which have ignition embeded as per doc
- Create the infra using libvirt
   - storage pool
   - network
   - VM using that ISO
- Wait for the cluster to come up and watch it with openshift-install
  binary.

Why?
- Right now the resource utilization is high because IPI doesn't support
  bootstrap in place functionailty and bootstrap + controller node run
  in parallel until bootstrap is finish.
- As per https://issues.redhat.com/browse/OCPBUGS-5945 looks like OVN-K
  have reduced footprint in case of SNO and with PR we are going to be
  true SNO so we can make use of it and have OVN-K as default network
  plugin.
- Till now we were using libvirt IPI which is only for dev preview so
  every single time when we face OCP releated issue it become low
  priority. But with this PR we follow official openshift doc around
  creating the bootstrap in place iso and then using it so it would
  become easy to have issue with supported steps.


snc-library.sh
```diff
@@ -184,22 +184,18 @@ function create_pvs() {
 
     # Add registry storage to pvc
     retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim": "crc-image-registry-storage"}}]' --type=json
-    # Remove emptyDir as storage for registry
-    retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "remove", "path": "/spec/storage/emptyDir"}]' --type=json
```
praveenkumar, line 188: This is removed because as part of SNO config image registry doesn't have `/spec/storage/emptyDir` object.

praveenkumar:
/hold

Bundle generated from it need to be tested and also OKD bits can be created.

cfergeau:
Some explanations in the PR  why we want to do this, and why this is a good change for us are needed. I can only see explanations about what the PR is doing, but "what" can be answered by looking at the PR, "why" cannot be answered by looking at the PR.

praveenkumar:
> Some explanations in the PR why we want to do this, and why this is a good change for us are needed. I can only see explanations about what the PR is doing, but "what" can be answered by looking at the PR, "why" cannot be answered by looking at the PR.

@cfergeau I added some details to commit logs, let me know if it still need improvements.

createdisk-library.sh
```diff
@@ -114,16 +110,9 @@ function create_qemu_image {
...
-
-  if [ "${BUNDLE_TYPE}" != "microshift" ]; then
-    create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
-    mv "${libvirtDestDir}/${VM_NAME}" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
-  else
-    create_qemu_image "$libvirtDestDir" "${VM_NAME}.qcow2" "microshift.qcow2"
-    mv "${libvirtDestDir}/microshift.qcow2" "${libvirtDestDir}/${SNC_PRODUCT_NAME}.qcow2"
-  fi
+  local VM_NAME="$2"
+  create_qemu_image "$libvirtDestDir" "${VM_NAME}.qcow2" "${VM_NAME}"
```
cfergeau, line 114: Are we still in the situation where the 3rd arg ("$VM_NAME") to `create_qemu_image` is an existing overlay file for openshift bundles, and a non-existent file for microshift? Or can `create_qemu_image` be simplified further? Looks like `create_bundle_qemu_image` and `create_qemu_image` could even become just one function?

net.xml.template
```diff
@@ -0,0 +1,24 @@
+<network>
```
cfergeau, line 1: `host-libvirt-net.xml.template` would be a more descriptive name

install-config.yaml
```diff
@@ -4,33 +4,26 @@ apiVersion: v1
...
 metadata:
-  creationTimestamp: null
   name: crc
 networking:
   clusterNetwork:
   - cidr: 10.217.0.0/22
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OpenShiftSDN
```
cfergeau, line 26: I assume the default networkType will be OVNKubernetes? Were there any measurements made with the new bundles ? https://github.com/crc-org/snc/pull/851#issuecomment-1942017621

snc.sh
```diff
@@ -69,8 +69,8 @@ echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RE
 # Extract openshift-install binary if not present in current directory
 if test -z ${OPENSHIFT_INSTALL-}; then
     echo "Extracting OpenShift baremetal installer binary"
```
cfergeau, line 71: `s/baremetal//`

snc.sh
```diff
@@ -86,7 +86,7 @@ if test -z "${SNC_OPENSHIFT_INSTALL_NO_DEBUG-}"; then
 else
         OPENSHIFT_INSTALL_EXTRA_ARGS=""
 fi
-
+BAREMETAL_RUNTIMECFG=$(${OC} adm release info -a ${OPENSHIFT_PULL_SECRET_PATH} ${OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE} --image-for=baremetal-runtimecfg)
```
cfergeau, line 89: In the hunk just before this one, we switched away from `openshift-baremetal-install` to `openshift-install`, and here we set `BAREMETAL_RUNTIMECFG`? Maybe we should keep using the baremetal installer after all?

snc.sh
```diff
@@ -164,11 +159,37 @@ DYNAMIC_DATA=$(base64 -w0 node-sizing-enabled.env) envsubst < 99_master-node-siz
...
-OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create ignition-configs ${OPENSHIFT_INSTALL_EXTRA_ARGS}
+OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create single-node-ignition-config ${OPENSHIFT_INSTALL_EXTRA_ARGS}
 # mask the chronyd service on the bootstrap node
-cat <<< $(${JQ} '.systemd.units += [{"mask": true, "name": "chronyd.service"}]' ${INSTALL_DIR}/bootstrap.ign) > ${INSTALL_DIR}/bootstrap.ign
-
-${OPENSHIFT_INSTALL} --dir ${INSTALL_DIR} create cluster ${OPENSHIFT_INSTALL_EXTRA_ARGS} || ${OC} adm must-gather --dest-dir ${INSTALL_DIR}
+cat <<< $(${JQ} '.systemd.units += [{"mask": true, "name": "chronyd.service"}]' ${INSTALL_DIR}/bootstrap-in-place-for-live-iso.ign) > ${INSTALL_DIR}/bootstrap-in-place-for-live-iso.ign
+
+# Download the image
+ISO_URL=$(OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=$OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE ${OPENSHIFT_INSTALL} coreos print-stream-json | grep location | grep $ARCH | grep iso | cut -d\" -f4)
```
cfergeau, line 167: Can we use `jq` to find the right image? I think I've asked already, but  I forgot the answer :-/

snc.sh
```diff
@@ -164,11 +159,37 @@ DYNAMIC_DATA=$(base64 -w0 node-sizing-enabled.env) envsubst < 99_master-node-siz
...
+curl -L ${ISO_URL} -o ${INSTALL_DIR}/rhcos-live.iso
+
+podman run --privileged --pull always --rm \
+      -v /dev:/dev -v /run/udev:/run/udev -v $PWD:/data \
+      -w /data quay.io/coreos/coreos-installer:release \
+      iso ignition embed --force \
+      --ignition-file ${INSTALL_DIR}/bootstrap-in-place-for-live-iso.ign \
+      ${INSTALL_DIR}/rhcos-live.iso
+
+sudo cp -Z ${INSTALL_DIR}/rhcos-live.iso /var/lib/libvirt/${SNC_PRODUCT_NAME}/rhcos-live.iso
```
cfergeau, line 177: `mv` instead of `cp`?

snc.sh
cfergeau: This variable seems to be used once at the other end of the file, maybe we can keep both definitions together?

snc.sh
```diff
@@ -262,6 +283,12 @@ while retry ${OC} get mcp master -ojsonpath='{.status.conditions[?(@.type!="Upda
     echo "Machine config still in updating/degrading state"
 done
 
+# Create a container from baremetal-runtimecfg image which consumed by nodeip-configuration service so it is
+# not deleted by `crictl rmi --prune` command
+${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- "sudo podman create --name baremetal_runtimecfg ${BAREMETAL_RUNTIMECFG}"
+
```
cfergeau, line 289: Can we re-pull the image after pruning? This container creation/removal feels hackish.

snc.sh
```diff
@@ -107,6 +107,14 @@ if [ -f /etc/NetworkManager/dnsmasq.d/openshift.conf ]; then
     sudo rm /etc/NetworkManager/dnsmasq.d/openshift.conf
 fi
 
+destroy_libvirt_resource
+sed -e "s|NETWORK_NAME|${SNC_PRODUCT_NAME}|" \
+    -e "s|CLUSTER_NAME|${SNC_PRODUCT_NAME}|" \
+    -e "s|BASE_DOMAIN|${BASE_DOMAIN}|" \
+    net.xml.template > net.xml
```
cfergeau, line 114: This could be moved to `create_libvirt_resource`?

tools.sh
```diff
@@ -159,6 +165,24 @@ function start_vm {
     wait_for_ssh ${vm_name} ${vm_ip}
 }
 
+function destroy_libvirt_resource {
```
cfergeau, line 168: I'd add use plural `resources` (same for `create`)

snc.sh
```diff
@@ -164,11 +159,37 @@ DYNAMIC_DATA=$(base64 -w0 node-sizing-enabled.env) envsubst < 99_master-node-siz
 export OPENSHIFT_INSTALL_INVOKER="codeReadyContainers"
```
cfergeau, line 159: Good opportunity to change this?

snc.sh
```diff
@@ -164,11 +159,37 @@ DYNAMIC_DATA=$(base64 -w0 node-sizing-enabled.env) envsubst < 99_master-node-siz
...
+    --vcpus ${SNC_CLUSTER_CPUS} \
+    --os-variant=rhel9-unknown \
+    --network=network:"${SNC_PRODUCT_NAME}",mac=52:54:00:ee:42:e1 \
+    --nographics \
+    --events on_reboot=restart \
+    --cdrom /var/lib/libvirt/${SNC_PRODUCT_NAME}/rhcos-live.iso \
+    --disk path=/var/lib/libvirt/${SNC_PRODUCT_NAME}/${SNC_PRODUCT_NAME}.qcow2,size=${CRC_VM_DISK_SIZE} \
+    --boot uefi \
+    --noautoconsole \
+    --wait
```
cfergeau, line 190: Can we share the virt-install cmdline with `microshift.sh`? They are very similar, but at the same time with unneeded (I think) differences.

createdisk.sh
```diff
@@ -53,9 +53,6 @@ VM_IP=$(sudo virsh domifaddr ${VM_NAME} | tail -2 | head -1 | awk '{print $4}' |
 wait_for_ssh ${VM_NAME} ${VM_IP}
```
cfergeau, line 49: >  we need to cached a specific image after prune (baremetal-runtimecfg) to run node-ip command.

What is this baremetal-runtimecfg/node-ip command? What do they achieve, when do they run?

cfergeau (commented):
Some questions/comments/minor improvements, but overall looks fine.
Bigger question is what the difference is wrt resource usage compared to what we currently are doing?

cfergeau:
Forgot to ask, does the SNO bundle work as is with crc? Or are there required changes in crc?

createdisk-library.sh
praveenkumar: @cfergeau yes it can simplified further but I want to have this PR as simple as possible initially and then put another one around refactoring/simplification.

snc.sh
praveenkumar: > In the hunk just before this one, we switched away from openshift-baremetal-install to openshift-install, and here we set BAREMETAL_RUNTIMECFG? Maybe we should keep using the baremetal installer after all?

We used `openshift-baremetal-install` because it had libvirt provider option not because we were doing some baremetal installation.  So even we keep using `baremetal` installer we still need this image to be part of cluster.

> This variable seems to be used once at the other end of the file, maybe we can keep both definitions together?

Yes it is just to pull that image again after prune,  sure we can move it together.

snc.sh
praveenkumar: Sure we can move it to `create_libvirt_resource helper.

snc.sh
praveenkumar: Sure we can change it but this is mostly on openshift telemetry side and this is what is reserved for us, have to check about it.

snc.sh
praveenkumar: There is some difference but can try to share but not part of this PR. (some refactor PR).

snc.sh
praveenkumar: I didn't try to use `jq` just put it here what is mention in the document https://docs.openshift.com/container-platform/4.14/installing/installing_sno/install-sno-installing-sno.html#install-sno-installing-sno-manually around grabbing the ISO_URL.

praveenkumar:
> Forgot to ask, does the SNO bundle work as is with crc? Or are there required changes in crc?

Only required changes would be having more resource because of OVNK, other than that it is going to work as expected.

createdisk-library.sh
cfergeau: I have to disagree with the PR being "as simple as possible", you already did 80% of the simplification work in create_bundle_qemu_image/create_qemu_image, and this makes it even harder than before to understand the `overlay` argument. 
If you want to keep things as simple as possible, I think this should be enough changes (disk image on the virt-install command line needs to be put in /var/libvirt/images/${SNC_PRODUCT_NAME}.qcow2)
```
diff --git a/createdisk-library.sh b/createdisk-library.sh
index 2b630bc..c80fd33 100755
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -118,7 +118,7 @@ function create_bundle_qemu_image() {
   local VM_NAME="$3"
 
   if [ "${BUNDLE_TYPE}" != "microshift" ]; then
-    create_qemu_image "$libvirtDestDir" "${VM_PREFIX}-base" "${VM_NAME}"
+    create_qemu_image "$libvirtDestDir" "${SNC_PRODUCT_NAME}.qcow2" "${VM_NAME}"
   else
     create_qemu_image "$libvirtDestDir" "${VM_NAME}.qcow2" "microshift.qcow2"
```

install-config.yaml
cfergeau: We need to remove the OVNKubernetes mentions from the PR/logs/..., or reword it :-/

snc.sh
cfergeau: Ah might be worth linking to https://docs.openshift.com/container-platform/4.14/installing/installing_sno/install-sno-installing-sno.html#generating-the-install-iso-manually_install-sno-installing-sno-with-the-assisted-installer to remember where it comes from if this ever changes?

snc.sh
cfergeau: We can/should minimize the differences, and use the same order for the arguments, as many identical arguments as possible, ...

snc.sh
cfergeau: I know why we used `openshift-baremetal-install` before, but now we switch away from it, and we set a `BAREMETAL` variable, as if we were doing a baremetal install even if we stopped using the baremetal installer. Weird to see both changes at the same time :)

install-config.yaml
praveenkumar: I think I removed it from the commit message will remove it from this PR description also.

cfergeau: Ah ok  good then, I haven't double checked

createdisk.sh
```diff
@@ -53,9 +53,6 @@ VM_IP=$(sudo virsh domifaddr ${VM_NAME} | tail -2 | head -1 | awk '{print $4}' |
...
 
 if [ ${BUNDLE_TYPE} != "microshift" ]; then
-    # Remove unused images from container storage
-    ${SSH} core@${VM_IP} -- 'sudo crictl rmi --prune'
-    
     # Disable kubelet service
     ${SSH} core@${VM_IP} -- sudo systemctl disable kubelet
```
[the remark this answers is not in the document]
praveenkumar: >isn't the kubelet service strongly tied to the kubeapi server?

@cfergeau if you mean when stopping this service should also stop the apiserver then no,  because pods are still running.

snc.sh
cfergeau: For `Download the image`, imo it's useful to note where the non-trivial sed/awk/... is coming from in case it needs to be changed.

createdisk-library.sh
praveenkumar: @cfergeau updated the PR and now have different commit to simplify it.

snc.sh
praveenkumar: Added as comment also.

snc.sh
praveenkumar: Added as different commit to consume it.

microshift.sh
```diff
@@ -122,7 +122,7 @@ microshift_pkg_dir=$(mktemp -p /tmp -d tmp-rpmXXX)
 chmod 0755 ${microshift_pkg_dir}
 download_microshift_rpm ${microshift_pkg_dir}
 create_iso ${microshift_pkg_dir}
-sudo cp -Z microshift/_output/image-builder/microshift-installer-*.iso /var/lib/libvirt/images/microshift-installer.iso
+sudo cp -Z microshift/_output/image-builder/microshift-installer-*.iso /var/lib/libvirt/${SNC_PRODUCT_NAME}/microshift-installer.iso
```
cfergeau, line 125: `microshift-installer.iso` might be leaked? `destroy_libvirt_resources` explicitly removes rhcos-live.iso, but not `microshift-installer.iso`.

createdisk-library.sh
```diff
@@ -86,33 +86,19 @@ EOF
...
-    local base=$2
-    local overlay=$3
+    local tempFile="temp_${SNC_PRODUCT_NAME}"
 
-    if [ -f /var/lib/libvirt/images/${base} ]; then
-      sudo cp /var/lib/libvirt/images/${base} ${destDir}
-    else
-      sudo cp /var/lib/libvirt/${SNC_PRODUCT_NAME}/${base} ${destDir}
-    fi
+    sudo cp /var/lib/libvirt/${SNC_PRODUCT_NAME}/${SNC_PRODUCT_NAME}.qcow2 ${destDir}
```
cfergeau, line 91: Makes more sense to me to tell `sparsify` to create the final file, rather than creating to a temp file we rename right away.
```
sudo cp /var/lib/libvirt/${SNC_PRODUCT_NAME}/${SNC_PRODUCT_NAME}.qcow2 ${destDir}/${tempFile}
sparsify ${destDir}  ${tempFile} ${SNC_PRODUCT_NAME}.qcow2
```
Working directly from `/var/lib/libvirt` would even avoid the copy, but that's trickier as this requires changes in `sparsify`, there will be permission problems, ... so not required right now.

cfergeau (commented):
looking good! a few minor comments

microshift.sh
praveenkumar: updated, now destroy_libvirt_resources need iso name as arg.

PR #814, opened 2023-11-03
createdisk: fix basedomain regex

praveenkumar:
Till 4.13 `basedomain` string was commented[0] but with 4.14 looks like it is not commented anymore and default basedomain used as `example.com`

[0] https://github.com/openshift/microshift/blob/release-4.13/packaging/microshift/config.yaml#L3
[1] https://github.com/openshift/microshift/blob/release-4.14/packaging/microshift/config.yaml#L16

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -78,9 +78,12 @@ EOF
     # also in case of microshift the ports like 2222, 443, 80 ..etc need to be manually added
     # and OCP/OKD/podman bundles have it disabled by default.
     ${SSH} core@${VM_IP} -- sudo systemctl disable firewalld
-    # Copy the sample microshift config and update the base domain with crc base domain
-    ${SSH} core@${VM_IP} -- sudo cp /etc/microshift/config.yaml.default /etc/microshift/config.yaml
-    ${SSH} core@${VM_IP} -- "sudo sed -i 's/#baseDomain: .*/baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}/g' /etc/microshift/config.yaml"
+    ${SSH} core@${VM_IP} -- cat /etc/microshift/config.yaml.default > config.yaml
+    ${YQ} eval --inplace ".dns.baseDomain = \"${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\""  config.yaml
+    ${SCP} config.yaml core@${VM_IP}:/home/core
+    ${SSH} core@${VM_IP} -- 'sudo mv /home/core/config.yaml /etc/microshift/config.yaml'
+    # Make sure `baseDomain` is set to crc.testing
+    ${SSH} core@${VM_IP} -- "grep '^\s\+baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}' /etc/microshift/config.yaml"
     # Remove the lvm system.device file since it have diskID and deviceName which changes
     # for different hypervisor and as per `man lvmdevices` if the file does not exist, or if lvm.conf
     # includes use_devicesfile=0, then lvm will not use a devices file.


adrianriobo (approved):
Commit message would require still a bit on thinking on how this fixes the problem the 
 non  working due to # is a bit hidden, may explain previously there was no basedomain so we have to set it up, now we need to override... but anway if it fine to me

createdisk.sh
```diff
@@ -80,7 +80,7 @@ EOF
     ${SSH} core@${VM_IP} -- sudo systemctl disable firewalld
     # Copy the sample microshift config and update the base domain with crc base domain
     ${SSH} core@${VM_IP} -- sudo cp /etc/microshift/config.yaml.default /etc/microshift/config.yaml
-    ${SSH} core@${VM_IP} -- "sudo sed -i 's/#baseDomain: .*/baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}/g' /etc/microshift/config.yaml"
+    ${SSH} core@${VM_IP} -- "sudo sed -i 's/baseDomain: .*/baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}/g' /etc/microshift/config.yaml"
```
cfergeau, line 83: Change this to `^baseDomain` or `^#?baseDomain` (not sure about the syntax) so that we don't regress again if this gets commented again. And I'd add a check/error if `baseDomain` was not modified as expceted.

cfergeau: or just use yq/jq to edit the file and ensure we set the right value in the right place?

praveenkumar: > Change this to ^baseDomain or ^#?baseDomain (not sure about the syntax) so that we don't regress again if this gets commented again. 

`^baseDomain` means line start with `baseDomain` but that is not the case

> or just use yq/jq to edit the file and ensure we set the right value in the right place?

Using `yq` seems better option but VM doesn't contain yq binary it have `jq` but with `jq` we can't make modification in yaml file right?

>  And I'd add a check/error if baseDomain was not modified as expceted
yes I will add that.

cfergeau: > [^baseDomain](url) means line start with baseDomain but that is not the case

https://www.gnu.org/software/sed/manual/html_node/Character-Classes-and-Bracket-Expressions.html

> Using yq seems better option but VM doesn't contain yq binary it have jq but with jq we can't make modification in yaml file right?

but we have yq on the host, could pipe guest `cat config.yaml.example` through host yq and write that to the guest

praveenkumar: used container image of yq so it should be good.

createdisk.sh
```diff
@@ -78,9 +78,9 @@ EOF
     # also in case of microshift the ports like 2222, 443, 80 ..etc need to be manually added
     # and OCP/OKD/podman bundles have it disabled by default.
     ${SSH} core@${VM_IP} -- sudo systemctl disable firewalld
-    # Copy the sample microshift config and update the base domain with crc base domain
-    ${SSH} core@${VM_IP} -- sudo cp /etc/microshift/config.yaml.default /etc/microshift/config.yaml
-    ${SSH} core@${VM_IP} -- "sudo sed -i 's/#baseDomain: .*/baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}/g' /etc/microshift/config.yaml"
+    ${SSH} core@${VM_IP} -- "sudo podman run --rm  -v /etc/microshift:/workdir:Z docker.io/mikefarah/yq:4.35.2 eval '.dns.baseDomain = \"${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\"'  config.yaml.default | sudo tee /etc/microshift/config.yaml"
```
cfergeau, line 81: Do we need to run `podman` as root? I don't think `--rm` will remove the container image which was just pulled.
I'd prefer that we don't add a new external dependency to our build process, mostly for security reasons.

createdisk.sh
```diff
@@ -78,9 +78,9 @@ EOF
...
+    # Make sure `baseDomain` is set to crc.testing
+    ${SSH} core@${VM_IP} -- "grep 'baseDomain: ${SNC_PRODUCT_NAME}.${BASE_DOMAIN}' /etc/microshift/config.yaml"
```
cfergeau, line 83: This will match `#baseDomain: ...`

createdisk.sh
praveenkumar: ah I need to remove the image, will update but if we remove it then there is no external dep left around also on the host we download the `yq` binary from github.com/mikefarah/yq and same we are doing as container image.

createdisk.sh
praveenkumar: @cfergeau how will this match `#baseDomain: example.com` doesn't it match only `baseDomain: crc.testing` or `#baseDomain: crc.testing` (which is not possible)?

createdisk.sh
praveenkumar: > Do we need to run podman as root?

Does it even matter in this command context? root or rootless at the end we want to have update the config file and delete it and then remove the image.

praveenkumar: Also with `rooless` it is not work because we are mounting `/etc/microshift` which have `root` ownership so it fail with `lsetxattr /etc/microshift/config.yaml.default: operation not permitted`

createdisk.sh
praveenkumar: yes but `#baseDomain: crc.testing` why this will even there in this file the commented one is `#baseDomain: example.com`?

createdisk.sh
cfergeau: > also on the host we download the yq binary from github.com/mikefarah/yq and same we are doing as container image.

The `yq` binary and this container image are 2 different "binaries", yes we already use a linux binary from a given github.com/mikefarah/yq release, but docker.io/mikefarah/yq:4.35.2 is another different binary that we decide to use and trust. 

> > Do we need to run podman as root?
> 
> Does it even matter in this command context? root or rootless at the end we want to have update the config file and delete it and then remove the image.

We run an untrusted container image inside the VM we are going to ship to customers. Imo yes it matters, we should think twice before running untrusted binaries as root inside this VM. And the less we do this, the better.

createdisk.sh
cfergeau: > yes but #baseDomain: crc.testing why this will even there in this file,

I don't know how it would get there, but the aim of this check is to catch future regressions around domain name setting. If we can avoid false negatives, even if they can't happen now, it's better :)

createdisk.sh
praveenkumar: > The `yq` binary and this container image are 2 different "binaries", yes we already use a linux binary from a given github.com/mikefarah/yq release, but docker.io/mikefarah/yq:4.35.2 is another different binary that we decide to use and trust.

If that is the concern then I can use same latest release for both the places.

> We run an untrusted container image inside the VM we are going to ship to customers. Imo yes it matters, we should think twice before running untrusted binaries as root inside this VM. And the less we do this, the better.

https://github.com/mikefarah/yq/blob/master/Dockerfile is what this image built, I am not sure when you say untrusted what you mean, some random image which we don't even sure how it built or even the images which we have access to container file and the source also?

cfergeau: We have no control over docker.io/mikefarah, I don't think we have a way of verifying that the container that you pull was really built using the docker file you linked to, `FROM alpine:3` in the file that you linked is also questionable, no way of knowing what exactly the base image was.

In short, this is one new piece of untrusted code that we are going to use in our build process.

praveenkumar: I don't wanted to block this PR over this dockerfile argument and since it really need for release I went with using YQ binary over host.

createdisk.sh
praveenkumar: Again not want to block this PR on something might happen in future. Added the check false negative.

PR #816, opened 2023-11-06
CI: use golang-1.20 image from openshift release to build crc binary

praveenkumar:
Our CI runs on GCP which have RHEL9.2 and go-1.19 so our jobs are failing, This PR use registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.20-openshift-4.15 to build the crc binary and we will revert this once RHEL-9.3 released which have gotool-1.20.x


ci.sh
```diff
@@ -47,7 +47,7 @@ sudo systemctl reload NetworkManager
 
 git clone https://github.com/code-ready/crc.git
 pushd crc
-make cross
+podman run --rm -v ${PWD}:/data:Z registry.ci.openshift.org/openshift/release:rhel-8-release-golang-1.20-openshift-4.15 /bin/bash -c "cd /data && GO_COMPLIANCE_INFO=0 GO_COMPLIANCE_DEBUG=0 make cross"
```
cfergeau, line 50: GO_COMPLIANCE_DEBUG does not need to be explicitly set, it's disabled by default.
If we start silencing the messages from the go wrapper in these images ourselves, we also need to do it in crc. We implicitly decided not to do it there.

Issue #826, opened 2023-11-27
Kernel used on macos is an old one


cfergeau:
Because of virtualization framework bugs, on macos we use older kernel versions otherwise the virtual machine fails to boot. However, the [`dnf repoquery`](https://github.com/crc-org/snc/blob/ae8d1350624fd7c8d8337dd371fa022bdd9795f6/createdisk-library.sh#L236) used to lookup the latest rhel 9.0 kernel does not work as expected, it picks up the `5.14.0-70.30.1.el9_0` kernel while there were newer 9.0 release https://access.redhat.com/errata/RHSA-2023:7382

PR #827, opened 2023-11-29
Use latest released rhel-9.0.z kernel

cfergeau:
The kernel we used so far was from October 2022, which is unexpected, we
intended to pick up the latest 9.0.z kernel.

This hardcodes the kernel version to the latest 9.0.z kernel released as
of today, need to find a better way to get this...

This fixes https://github.com/crc-org/snc/issues/826

This also fixes macOS x86_64 microshift bundles which fail to boot
without this. The kernel modules can't be loaded.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -233,10 +233,9 @@ function install_additional_packages() {
 function downgrade_rhel9_kernel {
     local vm_ip=$1
     local pkgDir=$(mktemp -d tmp-rpmXXX)
-    kernel_version=$(sudo dnf --quiet repoquery-na --queryformat '%{version}-%{release}.%{arch}' kernel.${ARCH} | grep el9_0 | tail -n 1)
 
     mkdir -p ${pkgDir}/packages
-    sudo yum download --downloadonly --downloaddir ${pkgDir}/packages kernel-${kernel_version} kernel-modules-extra-${kernel_version} kernel-core-${kernel_version} kernel-modules-${kernel_version} --resolve
+    podman run --rm -v "./${pkgDir}/packages":/packages:z registry.access.redhat.com/ubi9 yum --releasever=9.0 --repo="rhel-9-for-x86_64-baseos-eus-rpms" download --downloaddir /packages kernel kernel-modules-extra kernel-core kernel-modules
     ${SCP} -r ${pkgDir}/packages core@${vm_ip}:/home/core/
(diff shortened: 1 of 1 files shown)


adrianriobo:
Tested on lab machine x86 mac OS `12.7.1` and cluster now starts as expected

anjannath:
Not sure if its worth waiting for the openshift ci tests, as those have not been running in the last few days due to missing resources

cfergeau:
I might have a better solution in the next hour, as `yum --releasever=9.0  --enablerepo="rhel-9-for-x86_64-baseos-eus-rpms" repoquery kernel` gives me the kernel version I expect

gbraad (approved):

gbraad:
> I might have a better solution in the next hour

It has been an hour ago... can we consider that now a follow-up refinement?

cfergeau:
> > I might have a better solution in the next hour
> 
> It has been an hour ago... can we consider that now a follow-up refinement?

I pushed my changes at the same time as your comment ;)

gbraad (approved):
Changes came a minute after my comment; so these are now considered as 'packages' and will install as part of the follow-up process.

createdisk-library.sh
```diff
@@ -233,10 +233,9 @@ function install_additional_packages() {
 function downgrade_rhel9_kernel {
     local vm_ip=$1
     local pkgDir=$(mktemp -d tmp-rpmXXX)
-    kernel_version=$(sudo dnf --quiet repoquery-na --queryformat '%{version}-%{release}.%{arch}' kernel.${ARCH} | grep el9_0 | tail -n 1)
 
     mkdir -p ${pkgDir}/packages
-    sudo yum download --downloadonly --downloaddir ${pkgDir}/packages kernel-${kernel_version} kernel-modules-extra-${kernel_version} kernel-core-${kernel_version} kernel-modules-${kernel_version} --resolve
+    podman run --rm -v "./${pkgDir}/packages"):/packages:z registry.access.redhat.com/ubi9 yum --releasever=9.0 --repo="rhel-9-for-x86_64-baseos-eus-rpms" download --downloaddir /packages kernel kernel-modules-extra kernel-core kernel-modules
```
anjannath, line 238: there's a typo, an extra `)` after `"./${pkgDir}/packages"`

adrianriobo:
QE passed with bundles generated with the last push https://crcqe-asia.s3.amazonaws.com/nightly/microshift/4.14.3/20231129/13.0.1/qe-results/e2e-non-ux.xml

adrianriobo (approved):
QE PASS

anjannath:
we need to switch the repository to `rhel-9-for-aarch64-baseos-rpms` for the aarch64 packages :(

PR #828, opened 2023-11-30
use arch specific repository id for kernel package

anjannath:
to install the latest rhel-9 kernel we need to use the rhel-9-for-aarch64-baseos-eus-rpms for arm64 and for amd64 the rhel-9-for-x86_64-baseos-eus-rpms repo

this makes use of the ${ARCH} variable to form the repo id

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -235,7 +235,7 @@ function downgrade_rhel9_kernel {
     local pkgDir=$(mktemp -d tmp-rpmXXX)
 
     mkdir -p ${pkgDir}/packages
-    podman run --rm -v "./${pkgDir}/packages":/packages:z registry.access.redhat.com/ubi9 yum --releasever=9.0 --repo="rhel-9-for-x86_64-baseos-eus-rpms" download --downloaddir /packages kernel kernel-modules-extra kernel-core kernel-modules
+    podman run --rm -v "./${pkgDir}/packages":/packages:z registry.access.redhat.com/ubi9 yum --releasever=9.0 --repo="rhel-9-for-${ARCH}-baseos-eus-rpms" download --downloaddir /packages kernel kernel-modules-extra kernel-core kernel-modules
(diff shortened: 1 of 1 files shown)


gbraad (approved):

gbraad:
However, the idea was that ONLY the x86 image needed customization, as it otherwise wouldn't work. @cfergeau consistency between the images is probably preferred. WDYT?

anjannath:
also without this, the downstream bundle job fails, as when running the script in an aarch64 machine that repository is missing..
```
+ podman run --rm -v ./tmp-rpmRHj/packages:/packages:z registry.access.redhat.com/ubi9 yum --releasever=9.0 --repo=rhel-9-for-x86_64-baseos-eus-rpms download --downloaddir /packages kernel kernel-modules-extra kernel-core kernel-modules
Trying to pull registry.access.redhat.com/ubi9:latest...
Getting image source signatures
Checking if image destination supports signatures
Copying blob sha256:19dd821b7a83bf1ad388914394494ea983308667df914c57a315c52e312eb2c4
Copying config sha256:5e30e9cab1592c10f4ef3ac0513077a109771570b9bf456fbf17b335b387432f
Writing manifest to image destination
Storing signatures
Updating Subscription Management repositories.
Unable to read consumer identity
subscription-manager is operating in container mode.

This system is not registered with an entitlement server. You can use subscription-manager to register.

Error: Unknown repo: 'rhel-9-for-x86_64-baseos-eus-rpms'
Build step 'Execute shell' marked build as failure
```

cfergeau:
> However, the idea was that ONLY the x86 image needed customization, as it otherwise wouldn't work. @cfergeau consistency between the images is probably preferred. WDYT?

macOS 12 + arm64 also needed a downgraded kernel, see https://github.com/crc-org/snc/commit/68c6383de78d658b2581fb0f5097dbf34618befc

cfergeau (approved):

gbraad:
> macOS 12 + arm64 also needed

... so this is more of a macOS 12 specific issue, right?

PR #829, opened 2023-12-18
Remove crc-dnsmasq container and run dnsmasq directly on vm

praveenkumar:
Right now we use crc-dnsmasq container to provide the dns service for different `.testing` subdomain locally and till now it worked as expected when we were using the openshift-sdn but with OVN-K it is not going to work because this container use new podman bridge and we have fix IP address `10.88.0.8` for it. In case of of OVN-K this IP is not resolvable by dns pod of openshift and dns resolution for those `.testing` subdomain going to fail.

With this removal we need to adjust the code on crc side to put the configuration to VM dnsmasq service and use it.

```
=== using openshift-sdn ===
$ oc rsh busybox-sleep-pod
sh-5.1# ping 10.88.0.8
PING 10.88.0.8 (10.88.0.8) 56(84) bytes of data.
64 bytes from 10.88.0.8: icmp_seq=1 ttl=63 time=0.878 ms
64 bytes from 10.88.0.8: icmp_seq=2 ttl=63 time=0.068 ms

=== using ovn-k ===
sh-5.1# ping 10.88.0.8
PING 10.88.0.8 (10.88.0.8) 56(84) bytes of data.
^C
--- 10.88.0.8 ping statistics ---
15 packets transmitted, 0 received, 100% packet loss, time 14368ms
```

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -99,17 +99,14 @@ fi
 prepare_qemu_guest_agent ${VM_IP}
 
 image_tag="latest"
-if podman manifest inspect quay.io/crcont/dnsmasq:${OPENSHIFT_VERSION} >/dev/null 2>&1; then
+if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION} >/dev/null 2>&1; then
     image_tag=${OPENSHIFT_VERSION}
 fi
 
-# Add gvisor-tap-vsock and crc-dnsmasq services
+# Add gvisor-tap-vsock service
 ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
   podman create --name=gvisor-tap-vsock --privileged --net=host -v /etc/resolv.conf:/etc/resolv.conf -it quay.io/crcont/gvisor-tap-vsock:latest
   podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
-  touch /var/srv/dnsmasq.conf
-  podman create --ip 10.88.0.8 --name crc-dnsmasq -v /var/srv/dnsmasq.conf:/etc/dnsmasq.conf -p 53:53/udp --privileged quay.io/crcont/dnsmasq:${image_tag}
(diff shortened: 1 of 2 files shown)


praveenkumar:
/hold

I need to test the bundle generated from this to crc after making dnsmasq config change.

cfergeau:
>  In case of of OVN-K this IP is not resolvable by dns pod of openshift 

No way to ask OVN-K to setup an IP for the container so that the dns pod of openshift can access it?

praveenkumar:
> > In case of of OVN-K this IP is not resolvable by dns pod of openshift
> 
> No way to ask OVN-K to setup an IP for the container so that the dns pod of openshift can access it?

No, I did asked on internal channel and it can only allow network from the host so either you run that crc-dnsmasq container using host network which means it will not going to have container IP and which is kind of similar to running dnsmasq as service directly on the host.

createdisk.sh
```diff
@@ -99,17 +99,14 @@ fi
 prepare_qemu_guest_agent ${VM_IP}
 
 image_tag="latest"
-if podman manifest inspect quay.io/crcont/dnsmasq:${OPENSHIFT_VERSION} >/dev/null 2>&1; then
+if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION} >/dev/null 2>&1; then
     image_tag=${OPENSHIFT_VERSION}
 fi
 
-# Add gvisor-tap-vsock and crc-dnsmasq services
+# Add gvisor-tap-vsock services
```
cfergeau, line 106: `service`, no need for a plural anymore

cfergeau (commented):
As far as I can tell, this PR has no code for "and run dnsmasq directly on vm" which is mentioned in the commit log. I *assume* dnsmasq is already installed in the host and we don't need to do it ourselves? Is dnsmasq already running as a standalone service in the host? Or do we also need to start it ourselves?

Apart from that, looks fine.

createdisk.sh
praveenkumar: updated.

PR #832, opened 2024-01-02
Microshift: Update repo to generate 4.15 microshift bundle

praveenkumar:


microshift.sh
```diff
@@ -53,7 +53,25 @@ ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
...
 # yum-utils package is required for reposync utility to synchronize packages of a remote DNF repository to a local directory
 function configure_host {
-    sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
+    # https://github.com/openshift/microshift/blob/release-4.15/scripts/image-builder/configure.sh#L19-L35
+    source /etc/os-release
+    if [ "${ID}" == "rhel" ] && [ "${VERSION_ID}" == "9.3" ]; then
+        # osbuild from COPR to install version that:
+        # 1. can build rhel-93 images
+        # 2. doesn't have issues with unexpected mirror's RPM verification
+        #    (like the one in 9.3 beta at the moment of writing this comment)
```
cfergeau, line 62: This comment mentions 9.3 beta, but 9.3 has been released 2 months ago. Is this still an issue?

praveenkumar: @cfergeau yes it is still same, I just took copy same content so can update it.

praveenkumar: I can check microshift team if it is there and link it to this one.

praveenkumar: @cfergeau added the link.

microshift.sh
```diff
@@ -53,7 +53,26 @@ ssh-keygen -t ecdsa -b 521 -N "" -f id_ecdsa_crc -C "core"
 # createrepo package is required to create localrepo for microshift and it's dependenices
 # yum-utils package is required for reposync utility to synchronize packages of a remote DNF repository to a local directory
 function configure_host {
-    sudo dnf install -y git osbuild-composer composer-cli ostree rpm-ostree \
```
cfergeau, line 56: Looks like this is fixed in osbuild-composer 88.3-1.el9_3 https://access.redhat.com/errata/RHBA-2024:0475

praveenkumar: @cfergeau yes, I removed that commit now.

ci_microshift.sh
```diff
@@ -5,7 +5,8 @@ set -exuo pipefail
 sudo yum install -y make golang
 
 ./shellcheck.sh
-MICROSHIFT_PRERELEASE=yes ./microshift.sh
```
cfergeau: This line was changed in the first commit in this series, should they be squashed together?

praveenkumar: These are different commit the `MICROSHIFT_PRERELEASE` suppose to get the rpm packages from the mirror with `ocp-dev-preview` but since the mirror location is changed with `ocp` when they start rolling the rc bits so meanwhile we also need to change the `MIRROR` link.

cfergeau: Shouldn't the code handling `MICROSHIFT_PRERELEASE` be reworked then? Looking at microshift.sh, this is also not correct as there is a `release-4.15` branch in microshift repo
```
    if [ -n "${MICROSHIFT_PRERELEASE-}" ]; then
       git clone -b main https://github.com/openshift/microshift.git
    else
       git clone -b release-4.14 https://github.com/openshift/microshift.git
    fi
```

praveenkumar: Right, I created https://github.com/crc-org/snc/pull/844 one and will rebase accordingly.

cfergeau (approved):

microshift.sh
```diff
@@ -74,7 +74,7 @@ function download_microshift_rpm {
     elif [ -n "${MICROSHIFT_NVR-}" ]; then
         nvr_suffix="-${MICROSHIFT_NVR-}"
     fi
-    sudo yum download ${extra_opts} --downloaddir ${pkgDir} --downloadonly microshift${nvr_suffix} microshift-networking${nvr_suffix} microshift-release-info${nvr_suffix} microshift-selinux${nvr_suffix} microshift-greenboot${nvr_suffix}
+    sudo yum download ${extra_opts} --downloaddir ${pkgDir} --downloadonly microshift${nvr_suffix} microshift-networking${nvr_suffix} microshift-release-info${nvr_suffix} microshift-selinux${nvr_suffix} microshift-greenboot${nvr_suffix} microshift-olm${nvr_suffix}
```
cfergeau, line 77: The `microshift-olm` RPM is apparently optional now?
> To simplify the use of operators, we enabled the operator lifecycle manager (OLM) as an optional component for MicroShift (“dnf install microshift-olm”).

praveenkumar: @cfergeau yes but I think for crc point of view better to have it default. So user doesn't need to manually install it if required since it is still under the resource limit.

PR #833, opened 2024-01-04
Remove hack around MCO .orig file

praveenkumar:
https://issues.redhat.com/browse/OCPBUGS-11437 is now fixed which means we no longer need to use the workaround to update .orig file of pull secret to have it replace with a dummy one.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -440,10 +440,6 @@ function remove_pull_secret_from_disk() {
       "microshift")
         ${SSH} core@${VM_IP} -- sudo rm -f /etc/crio/openshift-pull-secret
 	;;
-      "snc")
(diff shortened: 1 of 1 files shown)


gbraad (approved):

praveenkumar:
@cfergeau with this fix there is no `.orig` file generated for pull secret and we are going to test the bundle using this PR if there is still leak of the pull secret.

cfergeau:
> @cfergeau with this fix there is no `.orig` file generated for pull secret and we are going to test the bundle using this PR if there is still leak of the pull secret.

If this does not reintroduce a pull secret leak, looks good to me.

praveenkumar:
> If this does not reintroduce a pull secret leak, looks good to me.

I run the rh-leaks on this bundle and didn't see any pull secret leak so merging it now.

PR #836, opened 2024-01-18
tools: Don't generate linux arm64 bundle

praveenkumar:
As of now there is no support on crc side to test and verify Linux arm64 bundles so better to not generate those and if user really want it to generate the can use `SNC_GENERATE_LINUX_BUNDLE=1 ./createdisk.sh`.

Also in case of downstream CI it will remove addition time of generation that linux bundle.


cfergeau:
@Prashanth684 since you added arm64 support, but mostly with an Apple Silicon focus, any comments on the removal on linux arm64 bundles?

praveenkumar:
> @Prashanth684 since you added arm64 support, but mostly with an Apple Silicon focus, any comments on the removal on linux arm64 bundles?

Support was added for mac side https://github.com/crc-org/snc/pull/454  ,  this changes were added by us.

cfergeau (approved):

Issue #838, opened 2024-01-23
Review capabilites for 4.15.X


adrianriobo:
We already seen changes within the capabilities lets track them on this issue:

adrianriobo:
With 4.15.0-rc3 when building installing the cluster we get an error within capabilities config:

```bash
+ ./openshift-baremetal-install --dir crc-tmp-install-data create manifests
level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: invalid "install-config.yaml" file: capabilities: Invalid value: types.Capabilities{BaselineCapabilitySet:"None", AdditionalEnabledCapabilities:[]v1.ClusterVersionCapability{"openshift-samples", "marketplace", "Console", "MachineAPI", "ImageRegistry", "DeploymentConfig", "Build", "OperatorLifecycleManager"}}: disabling CloudController capability available only for baremetal platforms
```

praveenkumar:
We have generated the bundle successfully so it is fixed.

PR #841, opened 2024-01-24
install-config: Add CloudCredential capability to enabled capabilities

praveenkumar:
From 4.15 CloudCredential is not part of core capability and it should be enabled if the cluster is not provision on baremetal (we are using libvirt provider).

- https://github.com/openshift/api/blob/2054c69227a1/config/v1/types_cluster_version.go#L488-L503

As part of libvirt provider we really don't need it and removing is as part of cvo-override.

This pr will fix following issue as part of snc run.
```
+ ./openshift-baremetal-install --dir crc-tmp-install-data create manifests
level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: invalid "install-config.yaml" file: capabilities: Invalid value: types.Capabilities{BaselineCapabilitySet:"None", AdditionalEnabledCapabilities:[]v1.ClusterVersionCapability{"openshift-samples", "marketplace", "Console", "MachineAPI", "ImageRegistry", "DeploymentConfig", "Build", "OperatorLifecycleManager"}}: disabling CloudController capability available only for baremetal platforms
```

fixes: #838

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -42,4 +42,5 @@ capabilities:
   - DeploymentConfig
   - Build
   - OperatorLifecycleManager
+  - CloudCredential
 publish: External


adrianriobo (approved):

adrianriobo:
Bundle has been created with the new capability added to the config. Cluster is starting properly now

cfergeau:
Bleh, can't parse `disabling CloudController capability available only for baremetal platforms`
/lgtm

PR #844, opened 2024-01-30
microshift: Add `OPENSHIFT_MINOR_VERSION` env variable to track branch

praveenkumar:
Currently `MICROSHIFT_PRERELEASE` is broken because for a unreleased version of microshift there might be already branch which is not track with `main` (eg. release-4.15) and bits are part of `ocp` instead of `ocp-dev-preview`. It happens when a specific openshift/microshift release goes to rc from ec. With this PR we are only using `MICROSHIFT_PRERELEASE` for download packages and
`MICROSHIFT_BRANCH` is used to track the branch info for microshift.

### DIFF
--- a/microshift.sh
+++ b/microshift.sh
@@ -12,11 +12,8 @@ BUNDLE_TYPE="microshift"
 INSTALL_DIR=crc-tmp-install-data
 SNC_PRODUCT_NAME=${SNC_PRODUCT_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
-if [ -n "${MICROSHIFT_PRERELEASE-}" ]; then
-    MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview}
-else
-    MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
-fi
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+OPENSHIFT_MINOR_VERSION=${OPENSHIFT_MINOR_VERSION:-4.14}
 
 if ! grep -q -i "release 9" /etc/redhat-release
 then
@@ -64,7 +61,7 @@ function configure_host {
 
 function enable_repos {
     sudo subscription-manager repos \
-       --enable rhocp-4.14-for-rhel-9-$(uname -i)-rpms \
+       --enable rhocp-${OPENSHIFT_MINOR_VERSION}-for-rhel-9-$(uname -i)-rpms \
        --enable fast-datapath-for-rhel-9-$(uname -i)-rpms
 }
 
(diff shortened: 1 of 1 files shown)


microshift.sh
```diff
@@ -12,11 +12,8 @@ BUNDLE_TYPE="microshift"
 INSTALL_DIR=crc-tmp-install-data
 SNC_PRODUCT_NAME=${SNC_PRODUCT_NAME:-crc}
 BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
-if [ -n "${MICROSHIFT_PRERELEASE-}" ]; then
-    MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview}
-else
-    MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
-fi
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
+MICROSHIFT_BRANCH=${MICROSHIFT_BRANCH:-4.14}
```
cfergeau, line 16: `4.14` is not a microshift branch, it's a microshift version.

fwiw, instead of manually setting both `MICROSHIFT_PRERELEASE` and `MICROSHIFT_BRANCH`, we can get `MICROSHIFT_BRANCH`  from `snc` git branch name (`release-${MICROSHIFT_BRANCH}`), and if it's `main`, then set it to 4.14.

praveenkumar: > 4.14 is not a microshift branch, it's a microshift version.

It is not specific version also (but just major) I put it because can consume 2 different place but if I use `release-4.14` then have to break it.

>fwiw, instead of manually setting both MICROSHIFT_PRERELEASE and MICROSHIFT_BRANCH, we can get MICROSHIFT_BRANCH from snc git branch name (release-${MICROSHIFT_BRANCH}), and if it's main, then set it to 4.14.

I though about it but then what if I am testing on test branch then it will take that branch and there is no mapping on microshift around it. Better to explict then implicit I think.

cfergeau: > It is not specific version also (but just major) I put it because can consume 2 different place but if I use release-4.14 then have to break it.

I'm not objecting to setting a variable to `4.14`, I'm objecting to naming this a "microshift branch" - it is not a microshift branch.

> I though about it but then what if I am testing on test branch then it will take that branch and there is no mapping on microshift around it. Better to explict then implicit I think.

If no mapping, it will fallback to the hardcoded 4.14, just like in this PR.
The more parameters we have to explicitly pass to tweak the behaviour of the script so that it does the right thing, the harder it gets to properly use it/... It quickly becomes a game of knowing exactly which combination of arguments is supported, because we did not test/don't support at all the other variations.

praveenkumar: >I'm not objecting to setting a variable to 4.14, I'm objecting to naming this a "microshift branch" - it is not a microshift branch.

Yes, naming is hard I choose branch which is not prefect but if I choose version that also not perfect :) , Do you have anything in mind or think version is much better than branch ( I don't have any strong opinion around it) ?

> The more parameters we have to explicitly pass to tweak the behaviour of the script so that it does the right thing, the harder it gets to properly use it/... It quickly becomes a game of knowing exactly which combination of arguments is supported, because we did not test/don't support at all the other variations.

Yes, so either we just make sure the script works only the hardcoded version in this case `4.14` and release-4.15 one works only for that branch and version. The whole point for that variable `MICROSHIFT_PRERELEASE` to build different version of microshift bundle even it is not released so to still tackle it either we introduce new variable or not let user to build bundles for non-release content.

cfergeau: > Yes, naming is hard I choose branch which is not prefect but if I choose version that also not perfect :) , Do you have anything in mind or think version is much better than branch ( I don't have any strong opinion around it) ?

`OPENSHIFT_MINOR_VERSION`? This is the naming used in https://access.redhat.com/support/policy/updates/openshift

praveenkumar: > > Yes, naming is hard I choose branch which is not prefect but if I choose version that also not perfect :) , Do you have anything in mind or think version is much better than branch ( I don't have any strong opinion around it) ?
> 
> `OPENSHIFT_MINOR_VERSION`? This is the naming used in https://access.redhat.com/support/policy/updates/openshift

👍🏼 , updated.

cfergeau (approved):

PR #846, opened 2024-01-31
Update oc binary download url for arm64 binary

praveenkumar:
Looks like there is confusion around the oc download path url to grab platfrom specific binary. Ideally if we use arch specific url to download the oc binary it should have by default for that specific org which is not the case so need to explicit specify the arch info in the url.

- https://issues.redhat.com/browse/OCPBUGS-28720

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -35,7 +35,11 @@ function download_oc() {
 
     if [ "${SNC_GENERATE_MACOS_BUNDLE}" != "0" ]; then
         mkdir -p openshift-clients/mac
-        curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/mac oc
+        if [ "${yq_ARCH}" == "arm64" ]; then
+            curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac-${yq_ARCH}-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/mac oc
+        else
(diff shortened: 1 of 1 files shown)


praveenkumar:
Only did for mac because that only the bundle we support for CRC side on arm64 arch.

snc-library.sh
```diff
@@ -35,7 +35,11 @@ function download_oc() {
 
     if [ "${SNC_GENERATE_MACOS_BUNDLE}" != "0" ]; then
         mkdir -p openshift-clients/mac
-        curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/mac oc
+        if [ "${yq_ARCH}" == "arm64" ]; then
+            curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac-${yq_ARCH}-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/mac oc
+        else
+            curl -L "${MIRROR}/${OPENSHIFT_RELEASE_VERSION}/openshift-client-mac-${OPENSHIFT_RELEASE_VERSION}.tar.gz" | tar -zx -C openshift-clients/mac oc
+        fi
```
cfergeau, line 42: Having a check that the oc binary matches what we expect (darwin/linux/windows x86_64/arm64) could be useful to avoid such issues in the future?

cfergeau (commented):
/lgtm
/approve

snc-library.sh
praveenkumar: @cfergeau you mean kind of run `file` command and see if it contain the arch which we expect?

PR #847, opened 2024-02-01
Change reclaim policy of default storage class to Retain

### LINKED ISSUE #835: should we change storage class reclaim policy to `Retain`
Prior to 77aa37ee5a74b3cff7506bae3b53292d028073f9 when we pre-created `pvs` the reclaim policy was set to `Recycle` but since there wasn't any volume re-cycler configured, the effect was similar to having reclaim policy set to `Retain`

it'd be better to set the same `reclaimPolicy: Retain` for the hostpath-provider storage class (currently it is not set and default is Delete), as this will allow users to recover any data after deleting the pvc

this came up in the issue: https://github.com/crc-org/crc/issues/3944

anjannath:
this restore to the original behavior we had for persistent volumes when they were manually created and had reclaimPolicy as Recycle

with this pvs dynamically created by the hostpath-provisioner won't be deleted allowing users to recover any data after deleting a pod/pv

fixes #835

### DIFF
--- a/kubevirt-hostpath-provisioner-csi/csi-sc.yaml
+++ b/kubevirt-hostpath-provisioner-csi/csi-sc.yaml
@@ -8,3 +8,4 @@ provisioner: kubevirt.io.hostpath-provisioner
 parameters:
   storagePool: local
 volumeBindingMode: WaitForFirstConsumer
+reclaimPolicy: Retain


anjannath:
/hold

need to test the bundles from CI

praveenkumar:
@anjannath let me know once bundle testing succeed.

praveenkumar:
you need to remove `on-hold` in that case.

/cherry-pick release-4.15

PR #851, opened 2024-02-06
Switch Network type from OpenShiftSDN to OVNKubernetes

praveenkumar:
With 4.15, OpenShiftSDN network plugin removed and OVNKubernetes is only option which is supported.

- https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html-single/release_notes/index#ocp-4-14-future-deprecation-sdn

### DIFF
--- a/install-config.yaml
+++ b/install-config.yaml
@@ -23,7 +23,7 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OpenShiftSDN
+  networkType: OVNKubernetes
   serviceNetwork:
   - 10.217.4.0/23
 platform:
--- a/snc.sh
+++ b/snc.sh
@@ -31,7 +31,7 @@ BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
 CERT_ROTATION=${SNC_DISABLE_CERT_ROTATION:-enabled}
 USE_PATCHED_RELEASE_IMAGE=${SNC_USE_PATCHED_RELEASE_IMAGE:-disabled}
 HTPASSWD_FILE='users.htpasswd'


snc.sh
```diff
@@ -31,7 +31,7 @@ BASE_DOMAIN=${CRC_BASE_DOMAIN:-testing}
 CRC_PV_DIR="/mnt/pv-data"
 SSH="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
 SCP="scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc"
-MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview}
+MIRROR=${MIRROR:-https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp}
```
cfergeau, line 34: Why?

install-config.yaml
```diff
@@ -23,7 +23,7 @@ networking:
     hostPrefix: 23
   machineNetwork:
   - cidr: 192.168.126.0/24
-  networkType: OpenShiftSDN
+  networkType: OVNKubernetes
```
cfergeau, line 26: >  you might need to increase memory resource to 10G

If there are concerns about resource usage, can this be mentioned/tracked in this PR?

snc.sh
praveenkumar: Because now rc bits of 4.15 available and it is from `https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp` instead `https://mirror.openshift.com/pub/openshift-v4/$ARCH/clients/ocp-dev-preview`

install-config.yaml
praveenkumar: This anyway need to propagate to crc codebase, but I can mention it here also.

snc.sh
cfergeau: This makes reviewing a lot smoother/easier if this is included in the commit log from the start, no need to go back and forth in comments

install-config.yaml
cfergeau: Imo it's better to make people aware of the impact of this PR *before* we take it in, rather than learning about it after the fact through a crc PR.

adrianriobo:
Test passed for the bundle with OVNKubernets, also I notice due to some code it set the memory to 12G, I will run it again with defaults to ensure it also passes with default values.

adrianriobo:
This has been tested and cluster eventually comes up with default resources. But it shows several issues during start, bit more of tested is required.

https://crcqe-asia.s3.ap-south-1.amazonaws.com/nightly/ocp/4.15.0-rc.5/qe-results/e2e-non-ux.xml

adrianriobo:
Indeed with current defaults I am not able to properly start the cluster :( even after 30 minutes operators are not able to stabilize)

adrianriobo:
I need to setup

`crc config set memory 11000`

otherwise I am not able to get a running state for the cluster :(

cfergeau:
https://github.com/crc-org/snc/pull/812 has:
> As per https://issues.redhat.com/browse/OCPBUGS-5945 looks like OVN-K
have reduced footprint in case of SNO and with PR we are going to be
true SNO so we can make use of it and have OVN-K as default network
plugin.

Would #812 help with the resource issues seen in this PR?

praveenkumar:
> #812 has:
> 
> > As per https://issues.redhat.com/browse/OCPBUGS-5945 looks like OVN-K
> > have reduced footprint in case of SNO and with PR we are going to be
> > true SNO so we can make use of it and have OVN-K as default network
> > plugin.
> 
> Would #812 help with the resource issues seen in this PR?

No, it is not help with resource utilization because it have same memory/cpu footprint for ovnk pods.

cfergeau:
> > #812 has:
> > > As per https://issues.redhat.com/browse/OCPBUGS-5945 looks like OVN-K
> > > have reduced footprint in case of SNO and with PR we are going to be
> > > true SNO so we can make use of it and have OVN-K as default network
> > > plugin.
> > 
> > 
> > Would #812 help with the resource issues seen in this PR?
> 
> No, it is not help with resource utilization because it have same memory/cpu footprint for ovnk pods.

This can be removed from the rationale for using SNO then:

> As per https://issues.redhat.com/browse/OCPBUGS-5945 looks like OVN-K
have reduced footprint in case of SNO and with PR we are going to be
true SNO so we can make use of it and have OVN-K as default network
plugin.

praveenkumar:
Closing since we are moving from libvirt IPI to SNO where it is by default OVN-K

PR #856, opened 2024-03-04
Stop accessing libvirt over unauthenticated qemu+tcp connection 

praveenkumar:
As part of libvirt IPI installation, it was required to have tcp socket enabled for cluster-api to work. We shifted to use SNO and now we don't require it anymore.

- https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections

### DIFF
--- a/snc-library.sh
+++ b/snc-library.sh
@@ -66,13 +66,9 @@ function run_preflight_checks() {
         echo "Checking libvirt and DNS configuration"
 
 	LIBVIRT_URI=qemu:///system
-	if [ ${bundle_type} == "snc" ] || [ ${bundle_type} == "okd" ]; then
-           LIBVIRT_URI=qemu+tcp://localhost/system
-	fi
-
         # check if we can connect to ${LIBVIRT_URI}
         if ! sudo virsh -c ${LIBVIRT_URI} uri >/dev/null; then
-              preflight_failure  "libvirtd is not listening for ${LIBVIRT_URI}, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
+              preflight_failure  "libvirtd is not accessible over ${LIBVIRT_URI}, check if libvirt daemon is running https://libvirt.org/daemons.html"
         fi
 
 	if ! sudo virsh -c ${LIBVIRT_URI} net-info default &> /dev/null; then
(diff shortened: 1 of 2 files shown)


cfergeau:
This is unrelated to libvirt socket activation (as in 'systemd socket activation'), this commit stops accessing libvirt over unauthenticated qemu+tcp (which was needed by cluster-api-provider-libvirt), and uses the local/default libvirt connection instead.

praveenkumar:
> accessing libvirt over unauthenticated qemu+tcp

Thanks, updated with the info.

snc-library.sh
```diff
@@ -66,13 +66,9 @@ function run_preflight_checks() {
...
 
 	LIBVIRT_URI=qemu:///system
-	if [ ${bundle_type} == "snc" ] || [ ${bundle_type} == "okd" ]; then
-           LIBVIRT_URI=qemu+tcp://localhost/system
-	fi
-
         # check if we can connect to ${LIBVIRT_URI}
         if ! sudo virsh -c ${LIBVIRT_URI} uri >/dev/null; then
-              preflight_failure  "libvirtd is not listening for ${LIBVIRT_URI}, see https://github.com/openshift/installer/tree/master/docs/dev/libvirt#configure-libvirt-to-accept-tcp-connections"
+              preflight_failure  "libvirtd is not listening for ${LIBVIRT_URI}, check if libvirt daemon is running https://libvirt.org/daemons.html"
```
cfergeau, line 71: I'd phrase this as "libvirtd is not accessible over qemu:///system, check if the libvirt daemon is running"

cfergeau (commented):
```
        # check if firewalld is configured to allow traffic from 192.168.126.0/24 to 192.168.122.1
```
can be removed (it's already commented out anyway)

```
        # check that api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} either can't be resolved, or resolves to 192.168.126.1[01]
```
This also can be reworked to only check for 192.168.126.11

praveenkumar:
> ```
>         # check if firewalld is configured to allow traffic from 192.168.126.0/24 to 192.168.122.1
> ```
> 
> can be removed (it's already commented out anyway)
> 
> ```
>         # check that api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} either can't be resolved, or resolves to 192.168.126.1[01]
> ```
> 
> This also can be reworked to only check for 192.168.126.11

Should we completely remove this because this was commented out for a long period and we didn't hit any issue around firewalld stuff?

snc-library.sh
praveenkumar: @cfergeau you don't want to have link in the failure message?

cfergeau:
> > ```
> >         # check if firewalld is configured to allow traffic from 192.168.126.0/24 to 192.168.122.1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > can be removed (it's already commented out anyway)
> > ```
> >         # check that api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} either can't be resolved, or resolves to 192.168.126.1[01]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This also can be reworked to only check for 192.168.126.11
> 
> Should we completely remove this because this was commented out for a long period and we didn't hit any issue around firewalld stuff?

We no longer need the VMs to be able to connect to the host libvirtd, so this can be removed. The fact that it was removed is one more reason for the removal :) 
The DNS resolution check should be kept.

snc-library.sh
cfergeau: I did not c&p it, but fine for me with or without, it's the beginning I wanted to rephrase

cfergeau (approved):

Issue #861, opened 2024-03-07
Spike: Can we use swap for OCP-4.15 so we don't have to increase the memory resource


praveenkumar:
As per https://docs.openshift.com/container-platform/4.15/nodes/nodes/nodes-nodes-managing.html#nodes-nodes-swap-memory_nodes-nodes-managing it is possible to use swap but this is in Tech preview. I was trying it out and see how reliable we can start the cluster without increase the resources on crc side. Because for 4.15, OVN-K is default and require more memory resource (~1.5G) then SDN for network operator. 

1. Enable Tech- preview feature: Can be done using install-config or as day-2 https://docs.openshift.com/container-platform/4.15/nodes/clusters/nodes-cluster-enabling-features.html#nodes-cluster-enabling-features-cli_nodes-cluster-enabling
2. Add kernel arg `swapaccount=1` which can be done with following machine config
```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: master
  name: 05-kernelarg-swapon
spec:
  kernelArguments:
    - swapaccount=1
``` 
3. Have a custom kubelet setting to enable swap
```bash
oc label machineconfigpool master kubelet-swap=enabled
```
```yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: KubeletConfig
metadata:
  name: swap-config
spec:
  machineConfigPoolSelector:
    matchLabels:
      kubelet-swap: enabled
  kubeletConfig:
    failSwapOn: false 
    memorySwap:
      swapBehavior: UnlimitedSwap
```
4. Have a swap partition in the VM
 - https://github.com/coreos/docs/blob/master/os/adding-swap.md#managing-swap-with-systemd
 

After all those steps swap is used for the the workload and take care of all the extra mem requirement but it has some caveats which is part of https://kubernetes.io/blog/2023/08/24/swap-linux-beta/ one. On openshift side since we enable Techpreview feature gate which means anything behind this gate is enabled automatic which are lot of things mentioned in the doc.

Node resources when swap is on (you can see memory is over committed because of swap is taking hit) and I started this cluster with default mem setting (9G)
```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource           Requests       Limits
  --------           --------       ------
  cpu                3245m (85%)    0 (0%)
  memory             9967Mi (117%)  0 (0%)
[...]

PR #864, opened 2024-03-15
Enable swap on the node

praveenkumar:
After switching to OVN-K we observed increase in memory resource and one
way to tackle it to have swap enabled on the node.

This PR is try to do following.
- Enable only `NodeSwap` feature gate instead complete Tech preview
  features.
     - https://docs.openshift.com/container-platform/4.15/rest_api/config_apis/featuregate-config-openshift-io-v1.html
- Add `swapaccount=1` to kernel argument which is required for the swap
    - https://docs.openshift.com/container-platform/4.15/nodes/nodes/nodes-nodes-managing.html#nodes-nodes-swap-memory_nodes-nodes-managing
- Add swap feature to kubelet config
- Add systemd unit file to create swap and enable it
- Extend the disk size by 4GB because that much of space used for swap

### DIFF
--- a/99-openshift-machineconfig-master-swap-service.yaml
+++ b/99-openshift-machineconfig-master-swap-service.yaml
@@ -1,0 +1,44 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-openshift-machineconfig-master-swap-service
+spec:
+  config:
+    ignition:
+      version: 3.4.0
+    systemd:
+      units:
+      - contents: |
+          [Unit]
+          Description=Create 4GB swap partition
+          Before=var-crc-swapfile1.swap
+          
+          [Service]
+          Type=oneshot
+          RemainAfterExit=yes
+          ExecStart=/usr/bin/mkdir -p /var/crc
+          ExecStart=/usr/bin/fallocate -l 4096m /var/crc/swapfile1
+          ExecStart=/usr/bin/chmod 600 /var/crc/swapfile1
+          ExecStart=/usr/sbin/mkswap /var/crc/swapfile1
+          
+          [Install]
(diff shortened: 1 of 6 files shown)


praveenkumar:
I am going to update the bundle size and resource consumption after testing the bundle. 

/hold till then.

99-openshift-machineconfig-master-swap-service.yaml
```diff
@@ -0,0 +1,44 @@
...
+          What=/var/crc/swapfile1
+          
+          [Install]
+          WantedBy=multi-user.target
+        enabled: false
+        name: var-crc-swapfile1.swap
+    networkd: {}
+    passwd: {}
+    storage: {}
+    osImageURL: ""
```
anjannath, line 44: removing these empty keys: `networkd, passwd...` from the ignition config causes errors?

99_feature-gate.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: config.openshift.io/v1
+kind: FeatureGate
+metadata:
+  name: cluster
+spec:
+  customNoUpgrade:
+    enabled:
+    - NodeSwap
+    - BuildCSIVolumes
```
anjannath, line 9: is the `BuildCSIVolumes` feature gate needed by `NodeSwap`?

praveenkumar: No, but is default enabled so I don't want to disable it.

99-openshift-machineconfig-master-swap-service.yaml
praveenkumar: Not tested, I just copied it from the doc so put as it is.

cfergeau:
Missing `Fixes: https://github.com/crc-org/snc/issues/861` line (had to go search for it  for some additional context I knew was there)

gbraad:
Hold?

praveenkumar:
close same as #924

PR #866, opened 2024-03-21
Add kubelet config file to update containerLogMaxSize and containerLogMaxFiles setting

praveenkumar:
…gMaxFiles setting

By default log size is set to 50Mi and number of log file is set to 5 but for local developement we might not need those setting where disk space is also critical. In this PR we are changing it to 5Mi and 2 respectivily.

During the testing (generation of 4.15.3 bundle) I can see reduction of ~ 200 MB to final bundle size.

### DIFF
--- a/99-kubelet-log-size-config.yaml
+++ b/99-kubelet-log-size-config.yaml
@@ -1,0 +1,11 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: container-logs-limit
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      pools.operator.machineconfiguration.openshift.io/master: ""
+  kubeletConfig:
+    containerLogMaxSize: "2Mi"
+    containerLogMaxFiles: "2"
--- a/snc.sh
+++ b/snc.sh
@@ -147,6 +147,8 @@ cp cluster-network-03-config.yaml ${INSTALL_DIR}/manifests/
 cp 99_master-chronyd-mask.yaml $INSTALL_DIR/openshift/
 # Add dummy network unit file
 cp 99-openshift-machineconfig-master-dummy-networks.yaml $INSTALL_DIR/openshift/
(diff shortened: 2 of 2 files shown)


cfergeau:
Is there a way to delete these logs after we've done the install?

praveenkumar:
> Is there a way to delete these logs after we've done the install?

It is not only we delete after done with install but also have same setting for end users which might don't know why the disk is taking that much space (if default is 50MiB which means every container can able to put that much logs before rotate).

cfergeau:
> but also have same setting for end users which might don't know why the disk is taking that much space (if default is 50MiB which means every container can able to put that much logs before rotate).

Hard to tell how many logs users will expect/need to debug their apps. Once the bundle is downloaded, in my opinion disk space is less of a problem. Is 5MiB / 2 log files something which is recommended somewhere?

praveenkumar:
> > but also have same setting for end users which might don't know why the disk is taking that much space (if default is 50MiB which means every container can able to put that much logs before rotate).
> 
> Is 5MiB / 2 log files something which is recommended somewhere?

No, It is not someone recommended, default for k8s is `10MiB` and 5 logs file. This is just provide user a better disk management when using CRC (since it is development use case).

cfergeau:
> This is just provide user a better disk management when using CRC (since it is development use case).

For a development use case, you **want** logs.

praveenkumar:
> > This is just provide user a better disk management when using CRC (since it is development use case).
> 
> For a development use case, you **want** logs.

We are not removing the **logs** just reducing size of the file. The question here is what would be right size one which provided by openshift 50MiB or one which we think should be good enough to get all the details.

cfergeau:
> > > This is just provide user a better disk management when using CRC (since it is development use case).
> > 
> > 
> > For a development use case, you **want** logs.
> 
> We are not removing the **logs** just reducing size of the file. The question here is what would be right size one which provided by openshift 50MiB or one which we think should be good enough to get all the details.

I honestly have no idea what a good size would be, which is why I'd favour sticking with the defaults, and only deleting all logs before bundle generation.

PR #872, opened 2024-04-05
microshift script related changes to update it to 4.16

praveenkumar:

PR #873, opened 2024-04-08
Sync master with release-4.15 branch

cfergeau:
A few commits were not cherry-picked from 4.15 to master, I don't think this
was intentional.


cfergeau:
One of these commits is also part of https://github.com/crc-org/snc/pull/872#issuecomment-2040210293 , I only saw this comment now ^^

Issue #854, opened 2024-02-21
Spike: topolvm CSI driver (supporting resize, limits) as replacement of kubevirt-hostpath-provisioner


anjannath:
- https://github.com/openshift/microshift/tree/main/docs/contributor/storage
- installing as an operator: https://docs.openshift.com/container-platform/4.15/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html and https://github.com/openshift/lvm-operator

anjannath:
Was able to deploy the lvms operator following the instructions from https://docs.openshift.com/container-platform/4.15/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html

for testing this:
- first need to create a separate partition to be used by the lvms operator
- and partition should not be used to create physical volume, lvms operator will manage it

currently in the OCP preset, the root partition has 12G of free space:
```
/dev/vda4        31G   20G   12G  64% /sysroot
```
we can shrink the root partition by few gbs and create another partition out of the remaining free space, suppose `/dev/vda4` of `26G` and `/dev/vda5` of `6G` which we can set as the device/partition to be used by the lvms operator

with the partition created, we can apply the following manifests to deploy the lvms operator:

```yaml
apiVersion: lvm.topolvm.io/v1alpha1
kind: LVMCluster
metadata:
  name: my-lvmcluster
  namespace: openshift-storage
spec:
  storage:
    deviceClasses:
    - name: vg1
      fstype: xfs
      default: true
      deviceSelector:
        paths:
        - /dev/vda5
        forceWipeDevicesAndDestroyAllData: true
      thinPoolConfig:
        name: thin-pool-1
        sizePercent: 90
        overprovisionRatio: 10
---
apiVersion: v1
kind: Namespace
metadata:
  labels:
    openshift.io/cluster-monitoring: "true"
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/audit: privileged
    pod-security.kubernetes.io/warn: privileged
  name: openshift-storage
---
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: openshift-storage-operatorgroup
  namespace: openshift-storage
spec:
  targetNamespaces:
  - openshift-storage
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: lvms
  namespace: openshift-storage
spec:
  installPlanApproval: Automatic
  name: lvms-operator
  source: redhat-operators
[...]

PR #877, opened 2024-04-11
Add machine config resource to create topolvm partition

praveenkumar:
Topolvm require a unused partition, which can be used by this operator and create layout accordingly. In this PR, machineconfig resource is used to create such partition as per document.

- https://docs.openshift.com/container-platform/4.15/installing/installing_with_agent_based_installer/installing-with-agent-based-installer.html#installing-ocp-agent-disk-partition_installing-with-agent-based-installer
- https://docs.openshift.com/container-platform/4.15/storage/persistent_storage/persistent_storage_local/persistent-storage-using-lvms.html


98-topolvm-partition.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  config:
+    ignition:
+      version: 3.4.0
+    storage:
+      disks:
+        - device: /dev/vda
+          partitions:
+            - label: topolvm
+              sizeMiB: 3072
```
anjannath, line 16: can we increase the size of this by shrinking the root partition a bit and adding the freed up space to this partition?

anjannath:
Thanks so much for looking into this and making it work, so the changes has to be done through `MachineConfig` and not directly using the ignition during boot for it to persist?

anjannath:
bundles are at https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/crc-org_snc/877/pull-ci-crc-org-snc-release-4.15-e2e-snc/1793504299356000256/artifacts/e2e-snc/gather-snc/artifacts/bundles.txt

gbraad:
this had questions around the (re)size of the root-partition, right?

anjannath:
> this had questions around the (re)size of the root-partition, right?

yes, this was posted in the crc issue, https://github.com/crc-org/crc/issues/4117 i updated it with the information from the slack conversation

praveenkumar:
@anjannath Since we are going to move with 4.16 as default branch and same is what going  to be tested. Should we close this PR and open a new against release-4.16 ?

anjannath:
have changed the base branch, but need to rebase the branch

cfergeau:
> have changed the base branch, but need to rebase the branch

For what it's worth, rebasing `praveenkumar:disk_part` does not trigger any conflicts if you use
```
git rebase HEAD~4 --onto origin/master
```
or
```
git rebase HEAD~4 --onto origin/release-4.16
```

cfergeau:
> @anjannath Since we are going to move with 4.16 as default branch and same is what going to be tested. Should we close this PR and open a new against release-4.16 ?

Don't we want this in `master` as well?

anjannath:
> Another comment, while going over the commits in this PR, and reading this PR as well, I have no idea why we want it in crc. I understand it is enabling (or switch to?) topolvm, but there is no discussion/explanation why this is a good thing to do (shortcomings of the previous solution, advantages of topolvm, ...)

there was some discussion about it in the issue: https://github.com/crc-org/snc/issues/854 but not a lot about its usefulness or shortcoming of the `hostpath-provisioner`.

the `lvm-storage-operator` supports a few more knobs for the pvc, like `resize` and setting `size` for the pvc which is not supported by `hostpath-provisioner`

snc-library.sh
```diff
@@ -140,35 +140,15 @@ function create_json_description {
...
-    # create StorageClass crc-csi-hostpath-provisioner
-    retry ${OC} apply -f kubevirt-hostpath-provisioner-csi/csi-sc.yaml
+function create_registry_pvc() {
 
     # Apply registry pvc with crc-csi-hostpath-provisioner StorageClass
     retry ${OC} apply -f registry_pvc.yaml
 
     # Add registry storage to pvc
     retry ${OC} patch config.imageregistry.operator.openshift.io/cluster --patch='[{"op": "add", "path": "/spec/storage/pvc", "value": {"claim": "crc-image-registry-storage"}}]' --type=json
+    # Chnage rollout method to Recreate, the default RollingUpdate is not supported with pvc accessmode ReadWriteOnce
```
cfergeau, line 150: Why switch to readwriteonce in registry_pvc.yaml if it's causing issues?

cfergeau: There's a "Chnage" typo

snc.sh
```diff
@@ -191,12 +191,12 @@ ${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} sudo hostnamectl set-hostname
...
 create_json_description ${BUNDLE_TYPE}
 
-# Create persistent volumes
-create_pvs ${BUNDLE_TYPE}
-
 # Install LVMS operator
 install_lvms_operator ${BUNDLE_TYPE}
 
+# Create persistent volumes
+create_registry_pvc
```
cfergeau, line 198: It's not creating persistent volumes, it's creating the registry persistent volume claims (I'm mostly pointing out that the comments are paraphrasing the function names, and here the method was renamed, but the comment not updated. Imo these comments have very little added value since the function names are very explicit).

kubevirt-hostpath-provisioner-csi/csi-driver-hostpath-provisioner.yaml
cfergeau, line 1: `remve kubevirt-hostpath-provisioner manifests` could be squashed into the commit which stopped using them.

lvms-operator/lvmcluster.yaml
cfergeau, line 1:
```
 install LVMS operator in openshift bundle

this creates the following resources
- namespace named openshift-storage
- operator group for LVMS operator
- subscription resource for LVMS operator
- LVMCluster custom resource to create the storage class
```

What is missing from this commit log is why we want to use the LVMS operator. The steps you list are somehow easy to get from the commit diff, but the motivation for doing the change is not something which can be found in a git diff.
You can reuse what was discussed in https://github.com/crc-org/snc/issues/854#issuecomment-1980225644
> the hostpath-provisioner doesn't support resize and limits, this is the main reason for the switch i think it'd be good to give users the ability to experiment with these features.

> there were questions about the size of the PV and how to have smaller PVs on our slack channel, 

> no its a limitation of the hostpath-provisioner since its just creating directories in the host it has no mechanism to ensure the size, it just takes as much free space is available

98-topolvm-partition.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  config:
+    ignition:
+      version: 3.4.0
+    storage:
+      disks:
+        - device: /dev/vda
+          partitions:
+            - label: topolvm
+              startMiB: 28600
```
cfergeau, line 16: ideally this would be derived from `CRC_VM_DISK_SIZE`

registry_pvc.yaml
```diff
@@ -5,8 +5,7 @@ metadata:
   namespace: openshift-image-registry
 spec:
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
       storage: 20Gi
-  storageClassName: crc-csi-hostpath-provisioner
```
cfergeau, line 12: The commit log could detail that this also removes the configuration of the hostpath-provisioner and explain why by removing `storageClassName` the image registry starts using topolvm.

lvms-operator/lvmcluster.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: openshift-storage
+spec:
+  storage:
+    deviceClasses:
+    - name: rhel
+      fstype: xfs
+      default: true
+      deviceSelector:
+        paths:
+        - /dev/vda5
```
cfergeau, line 14: Did you try using `/dev/disk/by-label/topolvm` to match what is done in `98-topolvm-partition.yaml`? This would make us less dependent on the actual disk layout.

lvms-operator/lvmcluster.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: lvm.topolvm.io/v1alpha1
+kind: LVMCluster
+metadata:
+  name: crc-lvmcluster
+  namespace: openshift-storage
+spec:
+  storage:
+    deviceClasses:
+    - name: rhel
```
cfergeau, line 9: why is this called `rhel`? Maybe topolvm or such?
https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/4.12/html-single/deploying_and_managing_logical_volume_manager_storage_on_single_node_openshift_clusters/index#scaling-storage-of-single-node-openshift-cluster-using-rhacm_sno calls it `vg1`

snc-library.sh
anjannath: > Why switch to readwriteonce in registry_pvc.yaml if it's causing issues?

`topolvm` doesn't support the `ReadWriteMany` access mode, and fails to create the volume with an error saying:

```
topolvm.io_lvms-operator-756689956-j2nxw_0a69a3a7-94d0-412d-9dbf-9f4b410aaece  failed to provision volume with StorageClass "lvms-rhel": rpc error: code = InvalidArgument desc = unsupported access mode: MULTI_NODE_MULTI_WRITER
```

but also `ReadWriteOnce` access mode is appropriate here, as [k8s docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) says:

> ReadWriteMany
    the volume can be mounted as read-write by many nodes.

and we only have one node

lvms-operator/lvmcluster.yaml
anjannath: yes, tried and this works

anjannath:
Seems its not a good Return on investment, closing! sorry about the noise

Issue #868, opened 2024-04-03
Download/use the same podman as for the cluster on linux


cfergeau:
This code https://github.com/crc-org/snc/blob/2a4aee82cf2eba7d0d1d51c634f15b72c1de8b44/createdisk-library.sh#L391 which was added in https://github.com/crc-org/snc/commit/f4d7855ff6cb452b9d8d4d3ee3bce8d27eaf7a51 is probably no longer needed.

redbeam:
Merged

PR #879, opened 2024-05-02
remove hardcoded podman binary version

redbeam:
Per #868

This basically reverts the (quick fix) change.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -388,8 +388,8 @@ function download_podman() {
     local arch=$2
 
     mkdir -p podman-remote/linux
-    curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-static.tar.gz | tar -zx -C podman-remote/linux podman-remote-static
-    mv podman-remote/linux/podman-remote-static podman-remote/linux/podman-remote
+    curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-static-linux_${arch}.tar.gz | tar -zx -C podman-remote/linux
+    mv podman-remote/linux/bin/podman-remote-static-linux_${arch} podman-remote/linux/podman-remote
     chmod +x podman-remote/linux/podman-remote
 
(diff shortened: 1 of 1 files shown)


createdisk-library.sh
```diff
@@ -388,7 +388,7 @@ function download_podman() {
     local arch=$2
 
     mkdir -p podman-remote/linux
-    curl -L https://github.com/containers/podman/releases/download/v4.1.1/podman-remote-static.tar.gz | tar -zx -C podman-remote/linux podman-remote-static
+    curl -L https://github.com/containers/podman/releases/download/v${version}/podman-remote-static.tar.gz | tar -zx -C podman-remote/linux podman-remote-static
```
praveenkumar, line 391: CI is failing because https://github.com/containers/podman/releases/v4.6.1/ have linux bits as per the arch. https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/crc-org_snc/879/pull-ci-crc-org-snc-release-4.15-e2e-microshift/1786112647393972224 


Also better to revert f4d7855ff6cb452b9d8d4d3ee3bce8d27eaf7a51 one and then add commit to fix the arch info

redbeam: It appears that they changed the directory structure in the archive as well (binary is now inside `bin` directory). Should the command to rename the binary keep it in the same directory or move it to `podman-remote/linux`?

cfergeau: > It appears that they changed the directory structure in the archive as well (binary is now inside bin directory). 

If you look at the git history/git blame, there were several changes related to the podman archive structure in the past.

> Should the command to rename the binary keep it in the same directory or move it to podman-remote/linux?

I'd keep the destination file in this rename as other part of snc/crc might rely on it
```
createdisk-library.sh:    mv podman-remote/linux/podman-remote-static podman-remote/linux/podman-remote
```

redbeam: Thanks for the input Christophe.

Issue #875, opened 2024-04-09
[4.16] Use `oc adm prune machineconfigpools` to remove older machine config


praveenkumar:
With 4.16 `oc` command will have machineconfigpools prune functionality which will remove the older machine config. As of now we are kind of using hack, go over machineconfigs and check the timestamp and then delete all except latest one. 

- https://github.com/crc-org/snc/blob/release-4.15/snc.sh#L265-L272

redbeam:
I downloaded the 4.16 version of `oc` but there is no `oc adm prune machineconfigpools` command. Is it going to be added in future versions? I might be missing something though.

```
mskvarla@mskvarla.brq.csb:~$ ./oc version
Client Version: 4.16.0-rc.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Unable to connect to the server: dial tcp: lookup api.ocp4.dev.psi.redhat.com on 127.0.0.1:53: no such host
mskvarla@mskvarla.brq.csb:~$ ./oc adm prune machineconfigpools
error: unknown command "machineconfigpools"
See 'oc adm prune -h' for help and examples
mskvarla@mskvarla.brq.csb:~$
```

Is it the `renderedmachineconfigs` subcommand?

praveenkumar:
yes, looks like they changed the option from `machineconfigs` to `rendermachineconfigs` now since this is experiment command.

```
$ ./oc adm prune renderedmachineconfigs -h
Experimental: This command is under development and may change without notice.
 Prune rendered MachineConfigs for an OCP v4 cluster.
 oc adm prune renderedmachineconfigs

```

redbeam:
Alright, so should we wait with this until the version is finalized?

redbeam:
Merged

PR #909, opened 2024-06-13
Use `oc adm prune renderedmachineconfigs` to remove older machine configs

redbeam:
#875


praveenkumar:
Tested the bundle and it now have only single machine config for master and worker which doesn't have pull secret information.

PR #914, opened 2024-06-18
Only use qemu-guest-agent on macos

cfergeau:
On linux/windows, we don't setup the needed vsock devices, so
qemu-guest-agent fails to start.
We can use `ConditionVirtualization=apple` to only start it when running
on a mac.

### DIFF
--- a/qemu-guest-agent.service
+++ b/qemu-guest-agent.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=QEMU Guest Agent
 IgnoreOnIsolate=True
+ConditionVirtualization=apple
 
 [Service]
 UMask=0077


praveenkumar:
@cfergeau this we used for time sync on the mac right?

cfergeau:
> @cfergeau this we used for time sync on the mac right?

Yes. Maybe this can be solved in a simpler way with `makestep -1` in chrony configuration, I haven't re-tested this.

praveenkumar:
> `makestep -1`

Just looked the `man chrony.conf` and got following which sound promising and we don't need to that configuration of the guest-agent if that works as expected, how does podman machine handle time drift in mac?

```       makestep threshold limit
           Normally chronyd will cause the system to gradually correct any time offset, by slowing down or speeding up the
           clock as required. In certain situations, e.g. when chronyd is initially started, the system clock might be so
           far adrift that this slewing process would take a very long time to correct the system clock.

           This directive forces chronyd to step the system clock if the adjustment is larger than a threshold value, but
           only if there were no more clock updates since chronyd was started than the specified limit. A negative value
           disables the limit.

           On most systems it is desirable to step the system clock only on boot, before starting programs that rely on
           time advancing monotonically forwards.

           An example of the use of this directive is:

               makestep 0.1 3

           This would step the system clock if the adjustment is larger than 0.1 seconds, but only in the first three
           clock updates.
```

Should we create an issue and set it as `good first one` for some new team members to try it out?

cfergeau:
> Just looked the `man chrony.conf` and got following which sound promising and we don't need to that configuration of the guest-agent if that works as expected, how does podman machine handle time drift in mac?

Last I checked, they were using `makestep`

cfergeau:
> > Just looked the `man chrony.conf` and got following which sound promising and we don't need to that configuration of the guest-agent if that works as expected, how does podman machine handle time drift in mac?
> 
> Last I checked, they were using `makestep`

The corresponding podman issue was https://github.com/containers/podman/issues/11541

praveenkumar:
Created https://github.com/crc-org/snc/issues/916 one and we can have this PR in for time being.

qemu-guest-agent.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=QEMU Guest Agent
 IgnoreOnIsolate=True
+ConditionVirtualization=apple
```
praveenkumar, line 4: As per `man systemd.unit` looks like this is added in v244, and current used version in OCP is 252 which is good for this usecase. 

```
      ConditionVirtualization=
           Check whether the system is executed in a virtualized environment and optionally test whether it is a specific
           implementation. Takes either boolean value to check if being executed in any virtualized environment, or one of
           "vm" and "container" to test against a generic type of virtualization solution, or one of "qemu", "kvm",
           "amazon", "zvm", "vmware", "microsoft", "oracle", "powervm", "xen", "bochs", "uml", "bhyve", "qnx", "apple",
           "sre", "openvz", "lxc", "lxc-libvirt", "systemd-nspawn", "docker", "podman", "rkt", "wsl", "proot", "pouch",
           "acrn" to test against a specific implementation, or "private-users" to check whether we are running in a user
           namespace. See systemd-detect-virt(1) for a full list of known virtualization technologies and their
           identifiers. If multiple virtualization technologies are nested, only the innermost is considered. The test may
           be negated by prepending an exclamation mark.

           Added in version 244.
```

praveenkumar (approved):

PR #924, opened 2024-06-25
[release-4.16] Enable swap on the node

praveenkumar:
Testing https://issues.redhat.com/browse/OCPBUGS-31035 one

After switching to OVN-K we observed increase in memory resource and one way to tackle it to have swap enabled on the node.

This PR is try to do following.
- Enable only `NodeSwap` feature gate instead complete Tech preview features. - https://docs.openshift.com/container-platform/4.15/rest_api/config_apis/featuregate-config-openshift-io-v1.html
- Add `swapaccount=1` to kernel argument which is required for the swap - https://docs.openshift.com/container-platform/4.15/nodes/nodes/nodes-nodes-managing.html#nodes-nodes-swap-memory_nodes-nodes-managing
- Add swap feature to kubelet config
- Add systemd unit file to create swap and enable it
- Extend the disk size by 4GB because that much of space used for swap

### DIFF
--- a/99-kubelet-config.yaml
+++ b/99-kubelet-config.yaml
@@ -1,0 +1,12 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: swap-config
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      pools.operator.machineconfiguration.openshift.io/master: ""
+  kubeletConfig:
+    failSwapOn: false 
+    memorySwap:
+      swapBehavior: UnlimitedSwap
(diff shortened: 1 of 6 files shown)


adrianriobo:
do you want to create `rc9` from here?

praveenkumar:
> do you want to create `rc9` from here?

No, I just wanted to test it with 4.16 to comment on the Jira ticket.

praveenkumar:
This is still not working as expected.

praveenkumar:
I am going to close this issue because swap enablement for the node is removed from the OCP doc and it is only for kubevirt side.

Before: https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/nodes/working-with-nodes#nodes-nodes-swap-memory_nodes-nodes-managing

now as part of 4.22 this is not documented anymore.

PR #925, opened 2024-06-27
gen-bundle-image: Add arm64 bundle image creation for linux

praveenkumar:


gen-bundle-image.sh
```diff
@@ -67,6 +74,7 @@ function sign_bundle_files {
   rm -fr *.sig
   if [[ ${preset} != "okd" ]]; then
      gpg --batch --default-key crc@crc.dev --pinentry-mode=loopback --passphrase-file ${GPG_SECRET_KEY_PASSPHRASE_PATH} --armor --output ${vfkit_bundle_arm64}.sig --detach-sig ${vfkit_bundle_arm64}
+     gpg --batch --default-key crc@crc.dev --pinentry-mode=loopback --passphrase-file ${GPG_SECRET_KEY_PASSPHRASE_PATH} --armor --output ${libvirt_bundle_arm64}.sig --detach-sig ${libvirt_bundle_arm64}
```
anjannath, line 77: should this be outside of the `okd` check, since image will be generated also for okd?

gen-bundle-image.sh
```diff
@@ -19,6 +19,7 @@ function set_bundle_variables {
 
    vfkit_bundle=crc${bundlePreset}_vfkit_${version}_amd64.crcbundle
    libvirt_bundle=crc${bundlePreset}_libvirt_${version}_amd64.crcbundle
+   libvirt_bundle_arm64=crc${bundlePreset}_libvirt_${version}_arm64.crcbundle
```
anjannath, line 22: ahh, so we don't want to generate arm64 libvirt bundles only for `okd` by looking at the changes, down.. so maybe this should be inside the `okd` check above at L17

praveenkumar: @anjannath done.

gen-bundle-image.sh
praveenkumar: No, we are not having arm64 support for okd.

anjannath (approved):

PR #927, opened 2024-07-02
Mirror and env changes for release-4.17 branch

praveenkumar:


repos/mirror-microshift.repo
```diff
@@ -1,5 +1,5 @@
 [mirror-microshift]
 name=microshift repo for mirror
-baseurl=https://mirror.openshift.com/pub/openshift-v4/$basearch/microshift/ocp-dev-preview/latest-4.16/el9/os/
```
cfergeau, line 3: Side note (unrelated to this PR)- we'll need to change `ocp-dev-preview` to `ocp` in the 4.16 branch.

cfergeau (approved):

repos/mirror-microshift.repo
praveenkumar: for 4.16 branch we don't even need to use the this repo since it is released so the packages should be come from official released repo, they don't put minor release to mirror so if we look https://mirror.openshift.com/pub/openshift-v4/x86_64/microshift/ocp/latest-4.16/el9/os/Packages/ it will point the latest rc release not the current released version which is `4.16.0`.

PR #930, opened 2024-07-03
testing release-4.17 pipeline

redbeam:


redbeam:
SNC test failed - that is expected for now. What's important is that the microshift test is OK.

PR #934, opened 2024-07-09
Delete pods and lease using etcd container

praveenkumar:
As of now when user consume the bundle to provision the cluster, old data around pods and lease is consumed and timestamp for those pods are also around same time when we first provisioned this cluster. This PR try to use etcd image to run etcd server using same db which created by the cluster  and uses etcdctl command to delete the pods and lease details. It will help to make sure every pods comes up in clean state.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -132,6 +132,16 @@ fi
 
 cleanup_vm_image ${VM_NAME} ${VM_IP}
 
+# Delete all the pods and lease from the etcd db so that when this bundle is use for the cluster provision, everything comes up in clean state.
+if [ ${BUNDLE_TYPE} != "microshift" ]; then
+    etcd_image=$(${SSH} core@${VM_IP} -- "sudo jq -r '.spec.containers[] | select(.name == \"etcd\") | .image' /etc/kubernetes/manifests/etcd-pod.yaml")
+    ${SSH} core@${VM_IP} -- "sudo podman run --rm --network=host --privileged --replace --name crc-etcd --detach --entrypoint etcd -v /var/lib/etcd:/store \"${etcd_image}\" --data-dir /store"
+    sleep 5
+    ${SSH} core@${VM_IP} -- "sudo podman exec crc-etcd etcdctl del --prefix /kubernetes.io/pods"
+    ${SSH} core@${VM_IP} -- "sudo podman exec crc-etcd etcdctl del --prefix /kubernetes.io/leases"
+    ${SSH} core@${VM_IP} -- "sudo podman stop crc-etcd"
+fi
+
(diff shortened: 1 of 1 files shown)


createdisk.sh
```diff
@@ -132,6 +132,16 @@ fi
 
 cleanup_vm_image ${VM_NAME} ${VM_IP}
 
+# Delete all the pods and lease from the etcd db so that when this bundle is use for the cluster provision, everything comes up in clean state.
+if [ ${BUNDLE_TYPE} != "microshift" ]; then
+    etcd_image=$(${SSH} core@${VM_IP} -- "sudo jq -r '.spec.containers[] | select(.name == \"etcd\") | .image' /etc/kubernetes/manifests/etcd-pod.yaml")
+    ${SSH} core@${VM_IP} -- "sudo podman run --rm --network=host --privileged --replace --name crc-etcd --detach --entrypoint etcd -v /var/lib/etcd:/store \"${etcd_image}\" --data-dir /store"
+    sleep 5
+    ${SSH} core@${VM_IP} -- "sudo podman exec crc-etcd etcdctl del --prefix /kubernetes.io/pods"
+    ${SSH} core@${VM_IP} -- "sudo podman exec crc-etcd etcdctl del --prefix /kubernetes.io/leases"
```
cfergeau, line 141: Does this have an impact on startup time? Can this cause "pod leftovers" to be left somewhere in crio/podman/... if new pods are created because etcd thinks they no longer exists?

createdisk.sh
```diff
@@ -132,6 +132,16 @@ fi
 
 cleanup_vm_image ${VM_NAME} ${VM_IP}
 
+# Delete all the pods and lease from the etcd db so that when this bundle is use for the cluster provision, everything comes up in clean state.
+if [ ${BUNDLE_TYPE} != "microshift" ]; then
+    etcd_image=$(${SSH} core@${VM_IP} -- "sudo jq -r '.spec.containers[] | select(.name == \"etcd\") | .image' /etc/kubernetes/manifests/etcd-pod.yaml")
```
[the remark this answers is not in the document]
praveenkumar: I don't think it is any better than using the existing resource which comes from release artifact only.

createdisk.sh
praveenkumar: As per my testing it actually improved startup time because there is less pod restart because existing pod have old data. crio/podman doesn't have leftover containers from pod this I checked after starting the VM without enabling the kubelet service and `sudo podman ps -a` or `sudo crictl ps -a` and it doesn't list any container related to openshift. I am still going to test around startup time on generated bundle from this PR (till now tested using bit of hack around bundle).

cfergeau (approved):
Can you add details to the commit log about what "It will help to make sure every pods comes up in clean state" mean? Which command can be used to check this?

PR #936, opened 2024-07-10
[release-4.16] Delete pods and lease using etcd container

praveenkumar:
As of now when user consume the bundle to provision the cluster, old data around pods and lease is consumed and timestamp for those pods are also around same time when we first provisioned this cluster. This PR try to use etcd image to run etcd server using same db which created by the cluster  and uses etcdctl command to delete the pods and lease details. It will help to make sure every pods comes up in clean state.

Something similar done in recert project

- https://github.com/openshift-kni/lifecycle-agent/blob/fd1cf27502b9c3f136fa1d161bcbd7ecf7c19da4/lca-cli/ops/ops.go#L186-L221

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -133,6 +133,16 @@ fi
 
 cleanup_vm_image ${VM_NAME} ${VM_IP}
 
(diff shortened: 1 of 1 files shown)


praveenkumar:
/close

#959 surpass it.

Issue #944, opened 2024-07-27
disable monitor tests to allow pipeline to succeed


redbeam:
The pipeline is currently not passing because of failing monitor tests. This is because the monitoring stack is disabled by default.
The (temporary?) solution is to disable the monitoring tests with a flag in `openshift-tests` binary.

redbeam:
Commit f1fbc51865f4634c0466d1796547c14ae17520ee:

```
The pipeline has been failing for a long time because of failing monitor tests.
The failing tests were:
    [Jira:"Test Framework"] monitor test alert-summary-serializer collection
    [Jira:"Test Framework"] monitor test metrics-endpoints-down collection
    [Jira:"Monitoring"] monitor test metrics-api-availability setup
    [Jira:"Monitoring"] monitor test monitoring-statefulsets-recreation setup
    [sig-network] can collect host-to-host poller pod logs
    [sig-network] can collect pod-to-host poller pod logs

The tests were disabled using the `--disable-monitor` flag in the
`openshift-tests` binary.
```

PR #945, opened 2024-07-27
disable monitor tests

redbeam:
#944


praveenkumar:
CI is still not green, does that mean, it still have something missing?

praveenkumar:
we have  a green CI run 👍🏼

praveenkumar:
In the commit logs let's put more details around why we are disabling those monitor logs.

redbeam:
> In the commit logs let's put more details around why we are disabling those monitor logs.

Alright, so something like:
```
Disabling monitor tests that prevent the test from passing. These tests are not needed because...
```
?

gbraad:
Is this because we disable Monitoring during deployment?

PR #950, opened 2024-08-29
tools.sh: increased number of retries

redbeam:
Increasing the number of retries should help with various issues during setup (i.e. waiting for cert renewal, getting pods...).

### DIFF
--- a/tools.sh
+++ b/tools.sh
@@ -108,7 +108,9 @@ if ! which ${PATCH}; then
 fi
 
 function retry {
-    local retries=10
+    # total wait time = 2 ^ (retries - 1) - 1 seconds
+    local retries=14
+
     local count=0
     until "$@"; do
         exit=$?


redbeam:
This will go to release-4.16, but the plan is to have it merged to release-4.17 and master as well.

cfergeau:
are you testing on a not so powerful machine? or is this openshift which got slower because it's doing more things?

redbeam:
I'm doing the testing on GCP, I don't know how powerful it is compared to the pipeline runners.

cfergeau:
ok, no objection from my side if this helps, let's wait for feedback from Praveen though!

redbeam:
Well, it's mainly because of #946 and #947, which both failed due to cert renewal not finishing in time (and the script running out of retries). There are probably some differences between master and 4.17, and 4.16 branches, which make the former fail.

praveenkumar:
@redbeam should we create this PR against master or release-4.17 branch where the CI is failing instead this branch where CI already green. Better to cherry-pick this commit to either #946 or #947 and see if CI succeed.

PR #955, opened 2024-09-04
Update ocp-dev-preview to ocp to mirror url for release-4.17

praveenkumar:
rc builds for 4.17 is now available and it is on mirror/ocp location.


redbeam:
Merged in #946

PR #957, opened 2024-09-12
ci.sh: moved test artifacts folder to crc-tmp-install-data

redbeam:
Currently the artifacts generated by `openshift-tests` are lost when the pipeline finishes, with this change we move it inside the `crc-tmp-install-data` folder so that it gets uploaded to GCS.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -66,9 +66,9 @@ crc config set bundle crc_libvirt_*.crcbundle
 crc setup
 crc start --disk-size 80 -m 24000 -c 10 -p "${HOME}"/pull-secret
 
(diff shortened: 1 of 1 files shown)


praveenkumar:
It is now available https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/crc-org_snc/957/pull-ci-crc-org-snc-release-4.16-e2e-snc/1834158412548542464/artifacts/e2e-snc/gather-snc/artifacts/crc-tmp-install-data/ here

PR #980, opened 2024-10-21
add systemd services for configuration after start

anjannath:
this adds 4 small systemd services that:
- creates crc specific configurations for dnsmasq
- sets a new uuid as cluster id
- creates the pod for routes-controller
- tries to grow the disk and filesystem
- service that checks if all the cluster operators are ready


systemd/crc-dnsmasq.sh
```diff
@@ -0,0 +1,28 @@
...
+cat << EOF > /etc/dnsmasq.d/crc-dnsmasq.conf
+listen-address=$ip
+expand-hosts
+log-queries
+local=/crc.testing/
+domain=crc.testing
+address=/apps-crc.testing/$ip
+address=/api.crc.testing/$ip
+address=/api-int.crc.testing/$ip
+address=/$hostName.crc.testing/$iip
```
praveenkumar, line 22: why we need this one for node name resolve (in OCP context) Also I think `ip` should be node name dummy IP which is 192.168.126.11?

systemd/crc-routes-controller.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -x
+
+# exit early if system-mode networking
+ping -c1 gateway > /dev/null 2>&1
+[ $? -eq 2 ] && exit 0
+
+export KUBECONFIG=/opt/kubeconfig
+oc apply -f /opt/crc/routes-controller.yaml
```
praveenkumar, line 10: even the service start after kubelet what make sure it is serving to api call? so may be first we wait till it start serving?

systemd/crc-dnsmasq.sh
anjannath: yeah probably not needed now, i think this was needed before moving to SNO and in the [crc code-base](https://github.com/crc-org/crc/blob/f4ccc5891d66148cf5c9c2361b2784b7b3f10aa4/pkg/crc/services/dns/template.go#L19) we still have it so added here as well

systemd/crc-routes-controller.sh
anjannath: yes, need to add some kind of check to see if the API is up

systemd/crc-dnsmasq.sh
```diff
@@ -0,0 +1,28 @@
...
+EOF
+
+sleep 2
+
+systemctl enable dnsmasq.service
+systemctl start dnsmasq.service
```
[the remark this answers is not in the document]
anjannath: Added `Before=dnsmasq.service` in the `crc-dnsmasq.service` and the `dnsmasq.service` already has 

```
[install]
WantedBy=multi-user.target
```
so its already enabled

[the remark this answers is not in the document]
anjannath: this didn't work so used `ExecStartPost=/usr/bin/systemctl start dnsmasq.service` which is working

systemd/crc-dnsmasq.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=CRC Unit for configuring dnsmasq
+Requires=ovs-configuration.service
+After=ovs-configuration.service
+
+[Service]
+Type=oneshot
+ExecCondition=/usr/bin/bash -c "/usr/bin/ping -c1 gateway && echo 1 || echo 0"
```
praveenkumar, line 8: What is this condition for? user mode networking?

anjannath: yes, when in user-mode networking `gateway` is resolvable from inside the VM, so when ping returns success it means its user-mode networking

systemd/ocp-cluster-status.sh
```diff
@@ -0,0 +1,43 @@
...
+    sleep $CLUSTER_HEALTH_SLEEP
+    if [[ $COUNTER == $CLUSTER_HEALTH_RETRIES ]]
+    then
+        return 1
+    fi
+    ((COUNTER++))
+done
+
+# need to set a marker to let `crc` know the cluster is ready
+# touch /tmp/.crc-cluster-ready
```
cfergeau, line 42: Instead of creating a file in the guest, you can also write a 'ready' notification message to a socket. podman does something similar: https://github.com/cfergeau/snc/commit/9e805300d7b9449a196a17651bcc0d1cbe748fb0

systemd/ocp-clusterid.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=CRC Unit setting random cluster ID
+After=kubelet.service
+Requires=kubelet.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/local/bin/ocp-clusterid.sh
+StandardOutput=journal
```
cfergeau, line 9: I believe this is the default, so it is not needed to specify it.

systemd/ocp-userpasswords.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=CRC Unit for adding pull secret to cluster
```
praveenkumar, line 2: This is not the pull secret but user auth update.

anjannath: thanks, fixed it!

systemd/crc-dnsmasq.sh
praveenkumar: did we test it with dummy IP `192.168.126.11` ?

systemd/ocp-cluster-ca.sh
```diff
@@ -0,0 +1,49 @@
...
+# openssl req -x509 -new -nodes -key $NAME-ca.key -sha256 -days $VALIDITY -out $NAME-ca.crt -subj "$CA_SUBJ"
+# openssl req -nodes -newkey rsa:2048 -keyout $USER.key -subj "$USER_SUBJ" -out $USER.csr
+# openssl x509 -extfile <(printf "extendedKeyUsage = clientAuth") -req -in $USER.csr \
+#    -CA $NAME-ca.crt -CAkey $NAME-ca.key -CAcreateserial -out $USER.crt -days $VALIDITY -sha256
+
+set -x
+
+if [ -z $CRC_CLOUD ]; then
+    exit 1
+fi
```
praveenkumar, line 21: What is this condition for and from where `CRC_CLOUD` env will come from?

systemd/ocp-custom-domain.sh
```diff
@@ -0,0 +1,52 @@
...
+EOF
+oc patch ingresses.config.openshift.io cluster --type=merge --patch-file=ingress-patch.yaml
+
+# patch API server to use new CA secret
+oc patch apiserver cluster --type=merge -p '{"spec":{"servingCerts": {"namedCertificates":[{"names":["api.'$EIP'.nip.io"],"servingCertificate": {"name": "nip-secret"}}]}}}'
+
+# patch image registry route
+oc patch -p '{"spec": {"host": "default-route-openshift-image-registry.'$EIP'.nip.io"}}' route default-route -n openshift-image-registry --type=merge
+
+#wait_cluster_become_healthy "authentication|console|etcd|ingress|openshift-apiserver"
```
praveenkumar, line 52: This is mostly from `crc-cloud` side because crc side we don't support it as of now by any conifguration but user can set it up manually.

systemd/crc-dnsmasq.sh
anjannath: you mean using `address=/$hostName.crc.testing/192.168.126.11`? instead of use the result from `hostname -i`?

praveenkumar: yes right, because this is used by internal service resolution so it might work.

systemd/ocp-custom-domain.sh
anjannath: yes it is only for crc-cloud use-case, it can be enabled by setting the `CRC_CLOUD` env variable

systemd/ocp-cluster-ca.sh
anjannath: it is set by the systemd unit based on an env file, if you check the `ocp-cluster-ca.service` file there's a 
```
EnvironmentFile=/opt/crc/crc-cloud
```

so when we create the AMI or an disk image for a different provider, we'll create this environment file during that generation of the cloud images

systemd/ocp-custom-domain.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=CRC Unit setting nip.io domain for cluster
+After=kubelet.service
+Requires=kubelet.service
+
+[Service]
+Type=oneshot
+EnvironmentFile=/opt/crc/crc-cloud
```
praveenkumar, line 8: This service fails for crc because this environment file is not available so may be we should touch that file or may be check if file exist then run this service otherwise not.

systemd/ocp-custom-domain.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+set -x
+
+if [ -z $CRC_CLOUD ]; then
+    exit 1
```
praveenkumar, line 6: why `exit 1` may be we should do following to make sure this service is not failed as soon as start for crc side.
```
if [ -z $CRC_CLOUD ]; then
    echo "Not running in crc-cloud mode"
    exit 0
fi
```

systemd/ocp-cluster-ca.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=CRC Unit setting custom cluster ca
+After=kubelet.service
+Requires=kubelet.service
+
+[Service]
+Type=oneshot
+EnvironmentFile=/opt/crc/crc-cloud
```
praveenkumar, line 8: same https://github.com/crc-org/snc/pull/980/files#r1884065076

systemd/ocp-cluster-ca.sh
praveenkumar, line 21: same https://github.com/crc-org/snc/pull/980/files#r1884067998

systemd/ocp-custom-domain.service
anjannath: the cluster will still start and it doesn't have any adverse effect even if this service fails, so my initial thought was that this service failing to start is a good indication that the bundle is not running in crc-cloud

notes-self-sufficient.md
```diff
@@ -0,0 +1,24 @@
+# Self sufficient bundles
+
+Since release 4.19.0 of OpenShift Local, the bundles generated by `snc` contains additional systemd services to provision the cluster and removes the need of
```
rohanKanojia, line 3:
```suggestion
Since release 4.19.0 of OpenShift Local, the bundles generated by `snc` contain additional systemd services to provision the cluster and remove the need for
```

notes-self-sufficient.md
```diff
@@ -0,0 +1,24 @@
...
+an outside entity for provision the cluster, although an outside process needs to create some files on pre-defined locations inside the VM for the  systemd
```
rohanKanojia, line 4:
```suggestion
an outside entity to provision the cluster, although an outside process needs to create some files on pre-defined locations inside the VM for the  systemd
```

notes-self-sufficient.md
```diff
@@ -0,0 +1,24 @@
...
+services to do their work.
+
+## The following table lists the systemd services and the location of files they need to provision the cluster, users of SNC needs to create those files
```
rohanKanojia, line 7:
```suggestion
## The following table lists the systemd services and the location of files they need to provision the cluster, users of SNC need to create those files
```

notes-self-sufficient.md
```diff
@@ -0,0 +1,24 @@
...
+| crc-routes-controller |               both               |                 none                 |         none         |
+|    ocp-cluster-ca     |               ocp                |        /opt/crc/custom-ca.crt        |     CRC_CLOUD=1      |
+|     ocp-clusterid     |               ocp                |                 none                 |         none         |
+|   ocp-custom-domain   |               ocp                |                 none                 |     CRC_CLOUD=1      |
+|      ocp-growfs       |               ocp                |                 none                 |         none         |
+|   ocp-userpasswords   |               ocp                | /opt/crc/pass_{kubeadmin, developer} |         none         |
+
+> [!NOTE]
+> "Marker env variable" is set using an env file, if the required env variable is not set then unit is skipped
+> some units are ran only when CRC_CLOUD=1 is set, these are only needed when using the bundles with crc-cloud
```
rohanKanojia, line 23:
```suggestion
> some units are run only when CRC_CLOUD=1 is set, these are only needed when using the bundles with crc-cloud
```

systemd/crc-cluster-status.sh
```diff
@@ -0,0 +1,43 @@
...
+    sleep $CLUSTER_HEALTH_SLEEP
+    if [[ $COUNTER == $CLUSTER_HEALTH_RETRIES ]]
+    then
+        return 1
+    fi
+    ((COUNTER++))
+done
+
+# need to set a marker to let `crc` know the cluster is ready
+# touch /tmp/.crc-cluster-ready
```
rohanKanojia, line 42: nit, Are these commented-out lines important? If not, shall we remove them?

systemd/crc-cluster-status.sh
```diff
@@ -0,0 +1,43 @@
...
+    do
+        if [[ $i == "False" ]]
+        then
+            return 0
+        fi
+    done
+    return 1
+}
+
+# rm -rf /tmp/.crc-cluster-ready
```
rohanKanojia, line 25: https://github.com/crc-org/snc/pull/980/files#r1888535323

systemd/ocp-clusterid.sh
```diff
@@ -0,0 +1,17 @@
...
+uuid=$(uuidgen)
+
+retry=0
+max_retry=20
+until `oc get clusterversion > /dev/null 2>&1`
+do
+    [ $retry == $max_retry ] && exit 1
+    sleep 5
+    ((retry++))
+done
```
rohanKanojia, line 15: I see this block of code duplicated across various scripts where we are waiting for some resource to become available. I'm not 100% sure if it's possible to do it in SystemD setup but can we move this to a generic method that can handle various resource types? WDYT?

cfergeau: Maybe with https://fedoramagazine.org/systemd-template-unit-files/
Or just by using a shared script.

anjannath: sure, I'll extract this into a function that we can re-use across scripts

systemd/crc-pullsecret.sh
```diff
@@ -0,0 +1,20 @@
...
+wait_for_resource secret
+
+# check if existing pull-secret is valid if not add the one from /opt/crc/pull-secret
+existingPsB64=$(oc get secret pull-secret -n openshift-config -o jsonpath="{['data']['\.dockerconfigjson']}")
+existingPs=$(echo "${existingPsB64}" | base64 -d)
+
+echo "${existingPs}" | jq -e '.auths'
+
+if [[ $? != 0 ]]; then
+    pullSecretB64=$(cat /opt/crc/pull-secret)
```
praveenkumar, line 17: pullSecretB64=$(cat /opt/crc/pull-secret | base64 -w0)

systemd/crc-cluster-status.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+set -x
+
+export KUBECONFIG=/opt/kubeconfig
+
+function check_cluster_unhealthy() {
```
praveenkumar, line 7: name should be `check_cluster_healthy` and return value should be 0 in case successful and 1 in case failure.

systemd/crc-cluster-status.sh
```diff
@@ -0,0 +1,43 @@
...
+    return 1
+}
+
+# rm -rf /tmp/.crc-cluster-ready
+
+COUNTER=0
+CLUSTER_HEALTH_SLEEP=8
+CLUSTER_HEALTH_RETRIES=500
+
+while $(check_cluster_unhealthy)
```
praveenkumar, line 31: this should change to `while ! check_cluster_healthy`

systemd/crc-pullsecret.sh
```diff
@@ -0,0 +1,21 @@
...
+wait_for_resource secret
+
+# check if existing pull-secret is valid if not add the one from /opt/crc/pull-secret
+existingPsB64=$(oc get secret pull-secret -n openshift-config -o jsonpath="{['data']['\.dockerconfigjson']}")
+existingPs=$(echo "${existingPsB64}" | base64 -d)
+
+echo "${existingPs}" | jq -e '.auths'
+
+if [[ $? != 0 ]]; then
+    pullSecretB64=$(cat /opt/crc/pull-secret base64 -w0)
```
praveenkumar, line 17: you forgot about the `pipe` here `cat /opt/crc/pull-secret | base64 -w0`

anjannath: thanks, fixed!

systemd/ocp-mco-sshkey.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -x
+
+source /usr/local/bin/crc-systemd-common.sh
+export KUBECONFIG="/opt/kubeconfig"
+
+pub_key_path="/opt/crc/id_rsa.pub"
+
+if [ ! -f "${pub_key_path} "]; then
```
adrianriobo, line 10: Is there a typo here: need an extra space between `"]` -> `" ]`

anjannath: ahh right, thanks for catching it

systemd/ocp-cluster-ca.sh
```diff
@@ -0,0 +1,40 @@
...
+    echo "kubeadmin password file not found"
+    exit 1
+fi
+
+PASS_KUBEADMIN="$(cat /opt/crc/pass_kubeadmin)"
+
+oc create configmap client-ca-custom -n openshift-config --from-file=ca-bundle.crt=${custom_ca_path}
+oc patch apiserver cluster --type=merge -p '{"spec": {"clientCA": {"name": "client-ca-custom"}}}'
+oc create configmap admin-kubeconfig-client-ca -n openshift-config --from-file=ca-bundle.crt=${custom_ca_path} \
+--dry-run -o yaml | oc replace -f -
```
anjannath, line 27: these steps comes from `crc-cloud` and based on this doc; https://access.redhat.com/solutions/6054981

albfan:
Test results:

- windows https://reportportal-openshiftlocal.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#openshift-local/launches/all/3403

PR #982, opened 2024-10-24
Update DEST_DIR_SUFFIX in case of pr

praveenkumar:
Recently on crc side[0] we have enabled the regex to capture bundle info and as part of this regex we are capturing the custom bundle which is random generated number like
`crc_okd_vfkit_4.16.7_amd64_2342465234654.crcbundle` but for bundles generated by the pr have suffix `.pr<pr_number>` which means again adjust the regex. We can still identify the bundle with `_<pr_number>` and this way we don't need to update the regex.

- https://github.com/crc-org/crc/pull/4343


anjannath (approved):

cfergeau (approved):
Is there anything which depends on the format of the bundle name in CI/... ?

PR #990, opened 2024-12-13
create macadam-ready bundles

cfergeau:
This adds a few unit files to notify podman machine that the boot is over.
This PR purpose is to generate bundles which can be shared more easily for testing, it is not meant for merging yet.
/hold

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -241,6 +241,17 @@ function prepare_qemu_guest_agent() {
     ${SSH} core@${vm_ip} 'sudo systemctl enable qemu-guest-agent.service'
 }
 
+function copy_ready_systemd_units() {
+    local vm_ip=$1
+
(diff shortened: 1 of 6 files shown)


praveenkumar:
@cfergeau should we put `wip` or `hold` for this PR just to make sure we don't accidentally merge it until tested?

cfergeau:
> @cfergeau should we put `wip` or `hold` for this PR just to make sure we don't accidentally merge it until tested?

The `do-not-merge/hold` label is already set, were you suggesting something more?

cfergeau:
One issue is that microshift bundles don't contain socat nor netcat which the systemd units require.

praveenkumar:
@cfergeau do we still need those for macadam, if yes can you update this and push it for release-4.18 or release-4.19 branch?

Issue #992, opened 2024-12-13
Remove initramfs/vmlinuz references


cfergeau:
snc's code still references vmlinuz/initramfs/...
I don't think crc still need them after https://github.com/crc-org/crc/pull/4309 so they can be removed.

PR #996, opened 2024-12-30
Remove initramfs/vmlinuz references for mac bundles

praveenkumar:
After https://github.com/crc-org/crc/pull/4309 we don't need to use those references for running the bundle on macOS so better to remove this unused code/reference.

fixes: #992

### DIFF
--- a/crc-bundle-info.json.sample
+++ b/crc-bundle-info.json.sample
@@ -9,7 +9,7 @@
   # - 1.3: remove of 'clusterInfo.kubeadminPasswordFile'
   # - 1.4: addition of 'arch'
   # - 1.5: remove of 'node[0].kernelCmdLine', 'node[0].initramfs', 'node[0].kernel'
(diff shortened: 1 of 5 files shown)


praveenkumar:
/hold

This need to be tested first for mac.

crc-bundle-info.json.sample
```diff
@@ -8,6 +8,7 @@
   # - 1.2: addition of 'storage.fileList'
   # - 1.3: remove of 'clusterInfo.kubeadminPasswordFile'
   # - 1.4: addition of 'arch'
+  # - 1.5: remove of 'node[0].kernelCmdLine', 'node[0].initramfs', 'node[0].kernel'
   "version": "1.4",
```
cfergeau, line 12: This needs to be set to 1.5

cfergeau (commented):
One small comment, otherwise looks good to me.

PR #999, opened 2025-01-10
Create MicroShift iso using image mode and bootc image builder

praveenkumar:
With 4.18 microshift removed the steps of creating the iso using image builder and there is no more `build.sh` script which is consumed by mircoshift.sh script to create it. This PR use the image mode and bootc image builder (BIB) to create the iso which is now microshift team also pushing forward.


image-mode/microshift/config/Containerfile.bootc-rhel9
```diff
@@ -0,0 +1,34 @@
+FROM registry.redhat.io/rhel9/rhel-bootc:9.4
+
+ARG USHIFT_VER=4.17
```
cfergeau, line 3: 4.17? it is 4.19 in microshift.sh

image-mode/microshift/config/Containerfile.bootc-rhel9
```diff
@@ -0,0 +1,34 @@
...
+    fi
+RUN dnf install -y firewalld microshift microshift-release-info cloud-utils-growpart qemu-guest-agent dnsmasq && \
+    dnf clean all && rm -fr /etc/yum.repos.d/*
+
+# https://github.com/containers/bootc/discussions/1036
+# /Users is created to make sure share directory works on
+# mac because on linux it is /home and for windows it is /mnt
+# and both are symlink to `var` already
+RUN rm -fr /opt && ln -sf var/opt /opt && mkdir /var/opt
+RUN ln -sf var/Users /Users && mkdir /var/Users
```
cfergeau, line 23: is the missing `/` before `var` intentional?

image-mode/microshift/config/Containerfile.bootc-rhel9
```diff
@@ -0,0 +1,34 @@
...
+
+# Mandatory firewall configuration
```
cfergeau, line 25: Mandatory? Application-specific? it would be useful to have more details about this, ideally some links to documentation.
While reading this, I don't even know if these are microshift/k8s requirements, crc requirements, ...

image-mode/microshift/config/config.toml.template
```diff
@@ -0,0 +1,112 @@
...
+# Set static hostname
+echo "REPLACE_HOSTNAME" > /etc/hostname
+chmod 644 /etc/hostname
+
+# Configure the firewall (rules reload is not necessary here)
+firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16
+firewall-offline-cmd --zone=trusted --add-source=169.254.169.1
+
+
+# Configure systemd journal service to persist logs between boots and limit their size to 1G
```
cfergeau, line 73: 1G is quite big.

image-mode/microshift/config/config.toml.template
```diff
@@ -0,0 +1,112 @@
...
+cat > /etc/systemd/journald.conf.d/microshift.conf <<EOF
+[Journal]
+Storage=persistent
+SystemMaxUse=1G
+RuntimeMaxUse=1G
+EOF
+
+# Update certificate trust storage in case new certificates were
+# installed at /etc/pki/ca-trust/source/anchors directory
+update-ca-trust
```
cfergeau, line 84: do we need this?

image-mode/microshift/config/config.toml.template
```diff
@@ -0,0 +1,112 @@
...
+ln -s /usr/lib/systemd/user/podman.socket /home/core/.config/systemd/user/default.target.wants/podman.socket
+chown -R core:core /home/core/.config
+
+mkdir -p /home/core/.config/containers
+tee /home/core/.config/containers/containers.conf <<EOF
+[containers]
+netns="bridge"
+rootless_networking="cni"
+EOF
+chown -R core:core /home/core/.config/containers
```
cfergeau, line 100: There's already such a `chown` higher in the script, better to group all of them at the end

```
# Make sure core user directory contents ownership is correct
chown -R core:core /home/core/
```

image-mode/microshift/build.sh
```diff
@@ -0,0 +1,144 @@
...
+    echo "          Size of the system root LVM partition. The remaining"
+    echo "          disk space will be allocated for data (default: ${LVM_SYSROOT_SIZE})"
+    echo "  -authorized_keys_file path_to_file"
+    echo "          Path to an SSH authorized_keys file to allow SSH access"
+    echo "          into the default 'redhat' account"
+    echo "  -use-mirror-repo <mirror_repo>"
+    echo "          Use mirror repo to get release candidate and engineering preview rpms"
+    echo "          like (https://mirror.openshift.com/pub/openshift-v4/x86_64/microshift/ocp-dev-preview/latest-4.18/el9/os/)"
+    echo "  -ushift-version <microshift-version>"
+    echo "          Version of microshift for image generation (default: ${USHIFT_VERSION}"
```
cfergeau, line 46: Why ushift and not microshift?

image-mode/microshift/build.sh
```diff
@@ -0,0 +1,144 @@
...
+title "Creating ISO image"
+sudo podman run --authfile ${OCP_PULL_SECRET_FILE} --rm -it \
+    --privileged \
+    --security-opt label=type:unconfined_t \
+    -v /var/lib/containers/storage:/var/lib/containers/storage \
+    -v "${BUILDDIR}"/config.toml:/config.toml \
+    -v "${BUILDDIR}":/output \
+    registry.redhat.io/rhel9/bootc-image-builder:latest \
+    --local \
+    --type iso \
```
cfergeau, line 142: Why do we generate an iso image which we then install? can't we directly generate a qcow2 image that we'll use for our VMs?

image-mode/microshift/config/Containerfile.bootc-rhel9
```diff
@@ -0,0 +1,34 @@
...
+
+ARG USHIFT_VER=4.17
+RUN if [ -z "${MIRROR_REPO}" ]; then \
+      dnf config-manager --set-enabled "rhocp-${USHIFT_VER}-for-rhel-9-$(uname -m)-rpms" \
+          --set-enabled "fast-datapath-for-rhel-9-$(uname -m)-rpms"; \
+    else \
+      # This is required to update the gpgcheck for repoID
+      repoID=$(echo "${MIRROR_REPO#*://}" | tr '/:' '_'); \
+      dnf config-manager --add-repo "${MIRROR_REPO}" \
+          --add-repo "https://mirror.openshift.com/pub/openshift-v4/$(uname -m)/dependencies/rpms/${USHIFT_VER}-el9-beta" \
```
cfergeau, line 11: Does this `https://mirror.openshift.com/pub/openshift-v4/...` repo need to be hardcoded in addition to the user-specified `"${MIRROR_REPO}"`?

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: it is microshift requirement, will add a comment.

image-mode/microshift/config/config.toml.template
praveenkumar: in case of custom cert but we can remove it.

image-mode/microshift/build.sh
praveenkumar: this is copied from existing script of microshift, can be updated.

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: yes this is dependency repo and it is different from mirror repo.

image-mode/microshift/build.sh
praveenkumar: we can generate qcow2 but I wanted to use same behavior like we have earlier (using the iso) which cause less changes on snc side.

image-mode/microshift/config/config.toml.template
praveenkumar: what do you think would be best 200M ?

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: yes it is just an arg and updated by build.sh script.

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: yes it was intentional since all files appear like following so just to have consistency. 

```
lrwxrwxrwx.   2 root root   8 Jan  1  1970 home -> var/home
lrwxrwxrwx.   2 root root   7 Jan  1  1970 lib -> usr/lib
lrwxrwxrwx.   2 root root   9 Jan  1  1970 lib64 -> usr/lib64
lrwxrwxrwx.   2 root root   9 Jan  1  1970 media -> run/media
lrwxrwxrwx.   2 root root   7 Jan  1  1970 mnt -> var/mnt
drwxr-xr-x.   1 root root   6 Jan  2 07:04 opt
lrwxrwxrwx.   2 root root  14 Jan  1  1970 ostree -> sysroot/ostree
dr-xr-xr-x. 544 root root   0 Feb  5 11:41 proc
lrwxrwxrwx.   2 root root  12 Jan  1  1970 root -> var/roothome
drwxr-xr-x.   1 root root  40 Feb  5 11:41 run
lrwxrwxrwx.   2 root root   8 Jan  1  1970 sbin -> usr/sbin
lrwxrwxrwx.   2 root root   7 Jan  1  1970 srv -> var/srv
```

image-mode/microshift/config/Containerfile.bootc-rhel9
cfergeau: we'll only start using this from 4.18 so 4.17 is quite unexpected.

image-mode/microshift/build.sh
cfergeau: Would be good to document it, as generating an installation iso is not how I expect a bootc workflow to be.

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: yes we can update it to 4.18/4.19

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: I don't find a good name around but adding the comment why this dependencies repo is required.

microshift.sh
```diff
@@ -42,96 +37,29 @@ create_libvirt_resources
...
-    # The home dir and files must have read permissions to group
-    # and others because osbuilder is running from another non-priviledged user account
-    # and allow it to read the files on current user home (like reading yum repo which is created as part of build script), it is required.
-    # https://github.com/openshift/microshift/blob/main/scripts/image-builder/configure.sh#L29-L32
-    chmod 0755 $HOME
-
-    scripts/image-builder/build.sh -microshift_rpms ${pkgDir} -pull_secret_file ${OPENSHIFT_PULL_SECRET_PATH} -lvm_sysroot_size 15360 -authorized_keys_file $(realpath ../id_ecdsa_crc.pub)
-    popd
+  local buildDir=$1
+  local extra_arg=""
```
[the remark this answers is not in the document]
praveenkumar: Done

image-mode/microshift/config/config.toml.template
praveenkumar: Done

image-mode/microshift/config/config.toml.template
praveenkumar: Removed it.

image-mode/microshift/config/config.toml.template
praveenkumar: done

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: Added comment

image-mode/microshift/build.sh
praveenkumar: Added comment

image-mode/microshift/build.sh
praveenkumar: updated to `microshift` instead ushift

createdisk.sh
```diff
@@ -72,12 +72,11 @@ EOF
     # also in case of microshift the ports like 2222, 443, 80 ..etc need to be manually added
     # and OCP/OKD/podman bundles have it disabled by default.
     ${SSH} core@${VM_IP} -- sudo systemctl disable firewalld
-    ${SSH} core@${VM_IP} -- cat /etc/microshift/config.yaml.default > config.yaml
-    ${YQ} eval --inplace ".dns.baseDomain = \"${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\""  config.yaml
-    ${SCP} config.yaml core@${VM_IP}:/home/core
-    ${SSH} core@${VM_IP} -- 'sudo mv /home/core/config.yaml /etc/microshift/config.yaml'
+    ${YQ} eval --inplace ".dns.baseDomain = \"${SNC_PRODUCT_NAME}.${BASE_DOMAIN}\""  00-microshift-dns.yaml
```
cfergeau, line 75: I think this will cause a file tracked in git to be modified each time `createdisk.sh` is called? This could be a template file instead, from which we generate a yaml file without modifying it in place.

isn't it late to do this at `createdisk.sh` time, ie once the install is done? Maybe it should be created by the kickstart file?

image-mode/microshift/config/Containerfile.bootc-rhel9
```diff
@@ -0,0 +1,36 @@
...
+# Mandatory firewall configuration for microshift to work correctly
+RUN firewall-offline-cmd --zone=public --add-port=22/tcp && \
+    firewall-offline-cmd --zone=trusted --add-source=10.42.0.0/16 && \
+    firewall-offline-cmd --zone=trusted --add-source=169.254.169.1 && \
+    firewall-offline-cmd --zone=trusted --add-source=fd01::/48
+# Application-specific firewall configuration to access route correctly
+RUN firewall-offline-cmd --zone=public --add-port=80/tcp && \
+    firewall-offline-cmd --zone=public --add-port=443/tcp && \
+    firewall-offline-cmd --zone=public --add-port=30000-32767/tcp && \
+    firewall-offline-cmd --zone=public --add-port=30000-32767/udp
```
cfergeau, line 36: we disable firewall in `createdisk.sh` for microshift bundles, this seems redundant?

image-mode/microshift/config/config.toml.template
cfergeau: the comment is now outdated though.

image-mode/microshift/config/Containerfile.bootc-rhel9
cfergeau: maybe `unreleased-mirror` or such? I'm worried the use of `-beta` is unexpected in some cases, and is really non obvious from the option name.

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: Yes, removing this from container file.

image-mode/microshift/config/config.toml.template
praveenkumar: removed this complete section because we don't need to persist logs between the boots since, this is something required for prod kind of setup.

image-mode/microshift/config/Containerfile.bootc-rhel9
praveenkumar: Used `unreleased-mirror` as environment variable to avoid confusion.

createdisk.sh
praveenkumar: Migrated this to kickstart file and remove the dropin file template.

cfergeau (approved):

ci_microshift.sh
```diff
@@ -5,6 +5,10 @@ set -exuo pipefail
 sudo yum install -y make golang
 
 ./shellcheck.sh
+ARCH=$(uname -m)
+if [[ "$ARCH" == "aarch64" ]]; then
+    export SNC_NON_FATAL_PREFLIGHT_CHECKS=1
```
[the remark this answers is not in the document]
praveenkumar: It is not directly related to this PR but I wanted to make sure both arm and amd build succeed before merge it. 

>  fix the capabilities check on aarch64. 

Yes, I think GCP doesn't allow native nested virtualization for arm64 and current job is using emulation only and it is slow but at least we are getting bundle build for arm64, will check if some different cloud provider can be used.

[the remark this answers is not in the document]
cfergeau: If we want to allow emulation on a specific platform, it should be the script starting snc on this platform which sets this variable. Ideally we'd have a way to only disable the "kvm" check there.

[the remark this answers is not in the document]
cfergeau: yeah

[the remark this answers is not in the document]
praveenkumar: Dropping this commit as of now and going with merge, will create follow up PR for it.

PR #1002, opened 2025-01-17
Add logic to make filesystem immutable for directory sharing from host

praveenkumar:
As part of `enable-shared-dirs`, in crc codebase we first make it mutable and then create top level directory for mount and again make it immutable. This is we are doing only for Mac because for linux it is $HOME and for windows we are mounting it under `/mnt`. In RHCOS `/home` and `/mnt` is already symlink to `/var/home` and `/var/mnt` which is by default mutable. So for mac we only need `/Users` to be symlink to `/var/Users`. We can remove that code on crc side during shared directories.

Another motivation is as part of `image-mode` base system `chattr` is not going to make system mutable for rootfs and there also we are going to use same workaround (make /Users to symlink to `/var`). This way both the bundle should behave same way.

### DIFF
--- a/99_master-create-users-symlink.yaml
+++ b/99_master-create-users-symlink.yaml
@@ -1,0 +1,60 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-openshift-machineconfig-master-users-symlink
+spec:
+  config:
+    ignition:
+      version: 3.4.0
+    systemd:
+      units:
+      - contents: |
+          [Unit]
+          Description=Allow systemd to create mount points on /
+          DefaultDependencies=no
+          Before=remote-fs.target
+
+          [Service]
+          Type=oneshot
+          ExecStart=chattr -i /
+
+          [Install]
+          WantedBy=remote-fs.target
+        enabled: true
+        name: immutable-root-off.service
+      - contents: |
+          [Unit]
+          Description=Set / back to immutable after mounts are done
+          DefaultDependencies=no
+          After=remote-fs.target
+
+          [Service]
+          Type=oneshot
+          ExecStart=chattr +i /
+
+          [Install]
+          WantedBy=remote-fs.target
+        enabled: true
+        name: immutable-root-on.service
+      - contents: |
+          [Unit]
+          Description=Create Users directory and mount to var
+          After=immutable-root-off.service
+          Before=immutable-root-on.service
+
+          [Service]
+          Type=oneshot
+          ExecStartPre=rm -f /Users
+          ExecStartPre=mkdir -p /var/Users
+          ExecStart=ln -sf /var/Users /Users 
+
+          [Install]
+          WantedBy=multi-user.target
+        enabled: true
+        name: create-symlink-users.service
+    networkd: {}
+    passwd: {}
+    storage: {}
+    osImageURL: ""
--- a/snc.sh
+++ b/snc.sh
@@ -147,6 +147,10 @@ cp cluster-network-03-config.yaml ${INSTALL_DIR}/manifests/
 cp 99_master-chronyd-mask.yaml $INSTALL_DIR/openshift/
 # Add dummy network unit file
 cp 99-openshift-machineconfig-master-dummy-networks.yaml $INSTALL_DIR/openshift/
+# Add unit file for creating /Users and symlink it to /var/Users
+# It is used for mounting shared directory on Mac
+# https://github.com/coreos/rpm-ostree/issues/337
+cp 99_master-create-users-symlink.yaml $INSTALL_DIR/openshift/
 # Add kubelet config resource to make change in kubelet
 DYNAMIC_DATA=$(base64 -w0 node-sizing-enabled.env) envsubst < 99_master-node-sizing-enabled-env.yaml.in > $INSTALL_DIR/openshift/99_master-node-sizing-enabled-env.yaml
 # Add codeReadyContainer as invoker to identify it with telemeter


99_master-create-users-symlink.yaml
```diff
@@ -0,0 +1,60 @@
...
+          [Unit]
+          Description=Allow systemd to create mount points on /
+          DefaultDependencies=no
+          
+          [Service]
+          Type=oneshot
+          ExecStart=chattr -i /
+          
+          [Install]
+          WantedBy=remote-fs-pre.target
```
anjannath, line 23: there's also a `local-fs-pre.target` and `local-fs.target` since we are not mounting any remote partition like nfs, maybe we can use these target instead.. found them on: https://www.freedesktop.org/software/systemd/man/latest/systemd.special.html

cfergeau: We need the file system to be immutable to create `/Users` before doing a virtio-fs mount. I don't know if systemd considers this local or remote, but an argument can be made that this is a remote filesystem since it lives on the host which is a different machine than the guest.

`local-fs-pre.target` feels like it would be too early since at this point we have no guarantee that the filesystem where we want to create the directory is already mounted.

snc.sh
```diff
@@ -147,6 +147,10 @@ cp cluster-network-03-config.yaml ${INSTALL_DIR}/manifests/
 cp 99_master-chronyd-mask.yaml $INSTALL_DIR/openshift/
 # Add dummy network unit file
 cp 99-openshift-machineconfig-master-dummy-networks.yaml $INSTALL_DIR/openshift/
+# Add unit file for creating /Users and symlink it to /var/Users
+# It is used for mounting shared directory on Mac
+# https://github.com/coreos/rpm-ostree/issues/337
+cp 99_master-create-users-symlink.yaml $INSTALL_DIR/openshift/
```
cfergeau, line 153: This will be run during the cluster installation, but will this be part of the bundle? Ideally it will not be part of what is shipped as once the install is done this should no longer be needed.

cfergeau (commented):
Just one question, otherwise makes sense to me.
Note that while `crc config` only allows the sharing of the home directory, the crc driver code should allow to share arbitrary directories. Ideally we would not break that too much (I understand newer microshift will need unbreaking if we ever want to share something other than `$HOME`, we'll solve that if that's ever needed)

cfergeau:
> Another motivation is as part of `image-mode` base system `chattr` is not going to make system mutable for rootfs and there also we are going to use same workaround (make /Users to symlink to `/var`). This way both the bundle should behave same way.

Do you know how they prevent image-mode systems from being mutables?

snc.sh
praveenkumar: > This will be run during the cluster installation, but will this be part of the bundle?

@cfergeau This is machineconfig resource which create the require services and it is going to part of bundle same way we have dummy network unit file. 

>  Ideally it will not be part of what is shipped as once the install is done this should no longer be needed.

You mean once `/Users` is part of bundle so we don't need those services?

praveenkumar:
> Do you know how they prevent image-mode systems from being mutables?

@cfergeau I don't know about tech details around how it is prevented but https://github.com/containers/bootc/discussions/1036 only talks about workaround.

snc.sh
cfergeau: Yes, these services are no longer needed once the directory is in the image. This could most likely even be done over ssh in createdisk.sh.
The part which worries me if it's enabled in the bundle is `ExecStartPre=rm -fr /Users`. If for any reasons the user homedir was mounted over virtiofs before this runs, we'll be deleting their homedir on the macos system.

praveenkumar: > Yes, these services are no longer needed once the directory is in the image. This could most likely even be done over ssh in createdisk.sh.

This is what I tried first time but it didn't work :( after reboot there is no `/Users` on the instance and that's why I created machine config with unit file so it can run for every boot.

> The part which worries me if it's enabled in the bundle is ExecStartPre=rm -fr /Users. If for any reasons the user homedir was mounted over virtiofs before this runs, we'll be deleting their homedir on the macos system.

@cfergeau the mount happen as part of crc codebase so I am not sure how it can be deleted the user's homedir because till that time VM is up and all the services are running.

cfergeau: > @cfergeau the mount happen as part of crc codebase so I am not sure how it can be deleted the user's homedir because till that time VM is up and all the services are running.

I agree right now this should not happen, however  who knows which changes we or users could do to the bundles 6 months from now.  Given the impact of this `rm -rf` if the directory is already mounted, we really should not be running it for every boot.

praveenkumar: > > @cfergeau the mount happen as part of crc codebase so I am not sure how it can be deleted the user's homedir because till that time VM is up and all the services are running.
> 
> I agree right now this should not happen, however who knows which changes we or users could do to the bundles 6 months from now. Given the impact of this `rm -rf` if the directory is already mounted, we really should not be running it for every boot.

But we are deleting the symlink not the actual content does this also affect because following what I tried and looks like it should be good or we should use the `unlink` command just to be sure we are not deleting?
```
prkumar@fedora:/tmp$ mkdir foobar
prkumar@fedora:/tmp$ touch foobar/hello

prkumar@fedora:/tmp$ ln -s /tmp/foobar /tmp/test

prkumar@fedora:/tmp$ ls test
hello
prkumar@fedora:/tmp$ rm -fr /tmp/test
prkumar@fedora:/tmp$ ls /tmp/foobar
hello
```

cfergeau: > But we are deleting the symlink not the actual content 

If you are only interested in deleting symlinks, you don't need the `-r` argument to `rm` (in `rm -fr /tmp/test`), and it's 100% clear the `rm` will not recursively delete anything. 
There's indeed a difference (that I was not really aware of) between `rm -rf /tmp/test/` and `rm -rf /tmp/test` when `test` is a symlink, but given the consequences if this `-r` triggers inadvertantly, I'd prefer that we don't add it.

praveenkumar: > > But we are deleting the symlink not the actual content
> 
> If you are only interested in deleting symlinks, you don't need the `-r` argument to `rm` (in `rm -fr /tmp/test`), and it's 100% clear the `rm` will not recursively delete anything. There's indeed a difference (that I was not really aware of) between `rm -rf /tmp/test/` and `rm -rf /tmp/test` when `test` is a symlink, but given the consequences if this `-r` triggers inadvertantly, I'd prefer that we don't add it.

Updated with `rm -f` without `-r` in the unit file.

praveenkumar: > > This is what I tried first time but it didn't work :( after reboot there is no /Users on the instance and that's why I created machine config with unit file so it can run for every boot.
> 
> Might be useful to record somewhere (commit log?) as this is non-obvious and unexpected :(

Added to commit log message.

PR #1003, opened 2025-01-21
gvforwarder as a systemd service

vyasgun:
- Create a tap device using nmcli with a hardcoded mac address
- Start gvforwarder systemd service which will use this device

Based on the following code:
https://github.com/crc-org/snc/pull/673
https://github.com/cfergeau/snc/commit/03a4054451adac2f5d53d2bb6cab0059722578f5

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -97,12 +97,38 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
     image_tag=${OPENSHIFT_VERSION}
 fi
 
+# create the tap device interface with specified mac address
+# this mac address is used to allocate a specific IP to the VM
+# when tap device is in use.
+${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+  nmcli connection add type tun ifname tap0 con-name tap0 mode tap autoconnect yes 802-3-ethernet.cloned-mac-address 5A:94:EF:E4:0C:EE
+EOF
+
+
 # Add gvisor-tap-vsock service
 ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
-  podman create --name=gvisor-tap-vsock --privileged --net=host -v /etc/resolv.conf:/etc/resolv.conf -it quay.io/crcont/gvisor-tap-vsock:latest
-  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  podman create --name=gvisor-tap-vsock quay.io/crcont/gvisor-tap-vsock:latest
+  podman cp gvisor-tap-vsock:/vm /usr/local/bin/gvforwarder
+  podman rm gvisor-tap-vsock
+  tee /etc/systemd/system/gv-user-network@.service <<TEE
+[Unit]
+Description=gvisor-tap-vsock Network Traffic Forwarder
+After=NetworkManager.service
+BindsTo=sys-devices-virtual-net-%i.device
+After=sys-devices-virtual-net-%i.device
+
+[Service]
+Environment=GV_VSOCK_PORT="1024"
+EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecStart=/usr/local/bin/gvforwarder -preexisting -iface %i -url vsock://2:\\\${GV_VSOCK_PORT}/connect
+
+[Install]
+WantedBy=multi-user.target
+
+TEE
   systemctl daemon-reload
-  systemctl enable gvisor-tap-vsock.service
+  systemctl enable gv-user-network@tap0.service
+
 EOF
 
 # Add dummy crio-wipe service to instance


cfergeau:
I'd recommend also picking up the changes from https://github.com/cfergeau/snc/commits/gvisor_service/ which update the unit files used in the PR to use a unit file close to https://github.com/containers/gvisor-tap-vsock/tree/main/contrib/systemd

With the current code, I still have this question/concern https://github.com/crc-org/snc/pull/673#discussion_r1150187087

createdisk.sh
```diff
@@ -97,12 +97,39 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
...
+ExecStart=/usr/local/bin/gvforwarder -preexisting -iface %i -url vsock://2:\\\${GV_VSOCK_PORT}/connect
+
+[Install]
+WantedBy=multi-user.target
+
+TEE
   systemctl daemon-reload
-  systemctl enable gvisor-tap-vsock.service
+  systemctl enable gv-user-network@tap0.service
+  systemctl start gv-user-network@tap0.service
```
cfergeau, line 131: Is the `start` required?

gv-user-network@.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=gvisor-tap-vsock Network Traffic Forwarder
```
cfergeau, line 2: This file is in the VM created in `createdisk.sh` with `tee /etc/systemd/system/gv-user-network@.service <<TEE ...`, is this file used?

cfergeau (commented):
Looks good to me, thanks a lot for putting this into shape/testing it.

gv-user-network@.service
vyasgun: yes, it's not needed. Removed

createdisk.sh
vyasgun: Removed

createdisk.sh
```diff
@@ -97,12 +97,38 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
     image_tag=${OPENSHIFT_VERSION}
 fi
 
+# create the tap device interface with specified mac address
+# this mac address is used to allocate a specific IP to the VM
+# when tap device is in use.
+${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+  nmcli connection add type tun ifname tap0 con-name tap0 mode tap autoconnect yes 802-3-ethernet.cloned-mac-address 5A:94:EF:E4:0C:EE
+EOF
```
cfergeau, line 105: Isn't this equivalent to `${SSH} core@${VM_IP} 'sudo nmcli connection add type tun ifname tap0 con-name tap0 mode tap autoconnect yes 802-3-ethernet.cloned-mac-address 5A:94:EF:E4:0C:EE'`?

cfergeau:
@praveenkumar can you take a look at this PR? you also looked into this in the past.

createdisk.sh
```diff
@@ -97,12 +97,38 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
...
+EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecStart=/usr/local/bin/gvforwarder -preexisting -iface %i -url vsock://2:\\\${GV_VSOCK_PORT}/connect
+
+[Install]
+WantedBy=multi-user.target
+
+TEE
   systemctl daemon-reload
-  systemctl enable gvisor-tap-vsock.service
+  systemctl enable gv-user-network@tap0.service
```
gbraad, line 130: :+1:

createdisk.sh
gbraad: In short; you are questioning why this needs to be wrapped in a `sudo bash -x -s`.
Would otherwise an error occur? I do not see characters that would be wrongly interpreted by the host shell (like zsh could do).

gbraad (commented):
I would approve, but I question the use of the command... first why invoked like this... though, even ... why is the creation of the tap not part of a systemd unit by itself? As in that case you can depend on it...

createdisk.sh
gbraad: ... I like how this can be targeted with `%i`, but this 'depends' on actions performed previously by creating this device.

For this increment, this would work. But most likely would change with the self-sufficient bundle.

cfergeau: The `nmcli` command creates a Network Manager configuration file in the bundle, the self-sufficient bundle should be similar from that perspective?

createdisk.sh
anjannath: this is not being changed for the self-sufficient bundle, its being with the existing situation which is that there is a container image which runs the `gvforwarder` and that container also has a dhcp client script which configures the interface using the dhcp service from gvisor-tap-vsock

anjannath: i think we can also scp the `NetworkManger` config file to `/etc/NetworkManager/system-connections` instead of running `nmcli` commands, there's a config file in: https://github.com/containers/gvisor-tap-vsock/blob/main/contrib/networkmanager/vsock0.nmconnection 

```
[connection]
id=tap0
type=tun
autoconnect=true
interface-name=tap0

[tun]
mode=2

[802-3-ethernet]
cloned-mac-address=5A:94:EF:E4:0C:EE

[ipv4]
method=auto

[proxy]
```

createdisk.sh
```diff
@@ -97,12 +97,38 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
...
 ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
-  podman create --name=gvisor-tap-vsock --privileged --net=host -v /etc/resolv.conf:/etc/resolv.conf -it quay.io/crcont/gvisor-tap-vsock:latest
-  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  podman create --name=gvisor-tap-vsock quay.io/crcont/gvisor-tap-vsock:latest
+  podman cp gvisor-tap-vsock:/vm /usr/local/bin/gvforwarder
+  podman rm gvisor-tap-vsock
+  tee /etc/systemd/system/gv-user-network@.service <<TEE
+[Unit]
+Description=gvisor-tap-vsock Network Traffic Forwarder
+After=NetworkManager.service
```
praveenkumar, line 116: Should we run this after `NetworkManager-wait-online.service`  beacuse as per `man NetworkManager-wait-online.service` this make sure that delays reaching the network-online target until NetworkManager reports that the startup is completed on the D-Bus

praveenkumar: The way I see this `gv-user-network` service is running the forwarder command for an existing interface and NetworkManager-wait-online service make sure that interface exist because it waits until all the network profile is enabled. Or my understanding is wrong here?

praveenkumar: >it is not clear when a tun interface reaches the activate state.

Yes and looking at the CI failure I think it hit the network failure so let's stick to `NetworkManager.service` only instead `NetworkManager-wait-online.service`

vyasgun: I tried this and we should keep the `After` value as it is

praveenkumar:
/close 

this is taken care by #1052

PR #1021, opened 2025-02-18
use the same apps domain for both ocp and microshift

anjannath:
since we don't have any requirement to keep the apps domain for ocp to be `apps-crc.testing` while making it same for both 'ocp' and 'microshift' means we can use the  dnsmasq config for both


anjannath:
@lilyLuLiu could you please run the downstream `e2e` pipeline for this

lilyLuLiu:
Downstream e2e test result on this PR :
`proxy test` failed on all platform  

Details are available https://issues.redhat.com/browse/CRC-137

anjannath:
we've decided to not change the existing `appsDomain` for OCP bundle as user's might have different scripts with the existing domain which would break once we change this, and for the purposes of the self-sufficient bundle work, we can handle it in the `createdisk-library.sh` script

PR #1031, opened 2025-03-18
microshift: Use no-cache option for microshift image build

praveenkumar:
By default podman uses the cache to build image and since we are using same node that means cached layer is already exist and old microshift package installed instead of latest one.


gbraad:
:+1 this can prevent a lot of issues when builds are created by others.

PR #1037, opened 2025-03-18
remove tip in login page about crc credentials 

anjannath:
the current patch is failing to be applied
with version 4.19.0-ec.1

### DIFF
--- a/oauth_cr.yaml
+++ b/oauth_cr.yaml
@@ -13,6 +13,3 @@ spec:
     htpasswd:
       fileData:
         name: htpass-secret 
-  templates:
-    login:
-      name: login-template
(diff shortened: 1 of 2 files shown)


anjannath:
this does not work as the `login.html` content on `4.19` is compressed/minified, which means `patch` cannot determine which lines to modify

we need to use a tool like `tidy` to first unminify/normalize the content and then apply the patch -> and then patch the openshift secret

login.html.patch
```diff
@@ -1,26 +1,29 @@
...
-               <div class="pf-c-form__group">
-                 <label class="pf-c-form__label" for="inputPassword">
+                 <div class="pf-v6-c-helper-text__item pf-m-error" id="-helper">
+                   <span class="pf-v6-c-helper-text__item-icon">
+@@ -1189,7 +1200,7 @@
+                   class="pf-v6-c-form__label-required" aria-hidden="true">&#42;</span></label>
+               <span class="pf-v6-c-form-control pf-m-required">
+                 <input required type="text" id="inputUsername" name="{{ .Names.Username }}"
+-                  value="{{ .Values.Username }}" autofocus />
++                  value="{{ if .Values.Username }}{{ .Values.Username }}{{ else }}developer{{ end }}" autofocus />
```
gbraad, line 26: Confused by this. What case does this handle?

anjannath: in the login form it automatically fills the username field with `developer` when its empty. This is was there in the older patch as well

PR #1038, opened 2025-03-19
Include cloud init package

praveenkumar:


createdisk-library.sh
```diff
@@ -189,11 +189,24 @@ function install_additional_packages() {
...
+            podman run --rm -v /home/core/packages:/packages:Z quay.io/centos/centos:stream9 sh -c "dnf install -y createrepo && createrepo /packages"
+            podman rmi quay.io/centos/centos:stream9
+EOF
+        ${SSH} core@${vm_ip} "sudo bash -c 'cat > /etc/yum.repos.d/local.repo << EOF
+[local]
+name=Local repo
+baseurl=file:///home/core/packages/
+enabled=1
+gpgcheck=0
+EOF'"
```
cfergeau, line 203: Have you considered serving this repo from the host? this might be a bit simpler, no need to copy files between host and guest, possibly less cleanup to do, ...

createdisk-library.sh
```diff
@@ -203,7 +203,7 @@ function install_additional_packages() {
 function prepare_hyperV() {
     local vm_ip=$1
 
-    install_additional_packages ${vm_ip} hyperv-daemons
+    ADDITIONAL_PACKAGES+=" hyperv-daemons"
```
cfergeau, line 219: This new global variable looks more error-prone than explicit calls to `install_additional_packages` with the list of packages to install. If you modify `ADDITIONAL_PACKAGES` after `install_additional_packages` has been called, the package you wanted won’t be installed for no obvious reason.

createdisk-library.sh
praveenkumar: No, but except scp packages it would be same steps even on host and then another step to run a local server on host which can serve it to guest.

createdisk-library.sh
praveenkumar: I didn't find any better way, may be have it local and then add packages to it and as part of install_additional_packages pass that list of packages. I wanted to run this install_additional_packages only once because it add a overlay layer each time so better to have all the changes in single layer.

cfergeau: I suspected this kind of reasons but couldn’t see it in the commit log.
Maybe add a comment right after the call to `install_additional_packages` which says `Beyond this point, packages added to the ADDITIONAL_PACKAGES variable won’t be installed in the guest` or so?

praveenkumar: Added

adrianriobo:
sure just starting the bundle creation to test it

adrianriobo:
Got an error while building in regards the cloud init units enablement: 

```
+ ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc core@192.168.126.11 -- 'sudo systemctl enable cloud-init cloud-config cloud-final'
Warning: Permanently added '192.168.126.11' (ED25519) to the list of known hosts.
Failed to enable unit: Unit file cloud-init.service does not exist.
++ handle_error
```

anjannath:
in the logs after package installation using `rpm-ostree` it says:

> Changes queued for next boot. Run "systemctl reboot" to start a reboot

maybe we need to run `rpm-ostree apply-live` or similar, only then the units files will become available

praveenkumar:
> Got an error while building in regards the cloud init units enablement:
> 
> ```
> + ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i id_ecdsa_crc core@192.168.126.11 -- 'sudo systemctl enable cloud-init cloud-config cloud-final'
> Warning: Permanently added '192.168.126.11' (ED25519) to the list of known hosts.
> Failed to enable unit: Unit file cloud-init.service does not exist.
> ++ handle_error
> ```

Ah, I thought I added those service after cleanup_vm (which restart it), updated the PR, let's wait for CI.

adrianriobo:
now it has finished on my end properly..I will leave it uploading the AMI tomorrow I will give a try on aws

Issue #1041, opened 2025-03-26
investigate `immutable-root-{on,off}.service` failure on `4.19.0-rc.3`


anjannath:
```
[core@crc ~]$ sudo journalctl -u immutable-root-on.service
Mar 24 14:50:33 crc systemd[1]: Starting Set / back to immutable after mounts are done...
Mar 24 14:50:33 crc chattr[1036]: chattr: Operation not supported while reading flags on /
Mar 24 14:50:33 crc systemd[1]: immutable-root-on.service: Main process exited, code=exited, status=1/FAILURE
Mar 24 14:50:33 crc systemd[1]: immutable-root-on.service: Failed with result 'exit-code'.
Mar 24 14:50:33 crc systemd[1]: Failed to start Set / back to immutable after mounts are done.
-- Boot f0dfcad5e18b4f4b8d5041d5b8bf4f85 --
Mar 26 05:39:22 crc systemd[1]: Starting Set / back to immutable after mounts are done...
Mar 26 05:39:22 crc chattr[1004]: chattr: Operation not supported while reading flags on /
Mar 26 05:39:22 crc systemd[1]: immutable-root-on.service: Main process exited, code=exited, status=1/FAILURE
Mar 26 05:39:22 crc systemd[1]: immutable-root-on.service: Failed with result 'exit-code'.
Mar 26 05:39:22 crc systemd[1]: Failed to start Set / back to immutable after mounts are done.

[core@crc ~]$ sudo journalctl -u immutable-root-off.service
Mar 24 14:50:32 crc chattr[861]: chattr: Operation not supported while reading flags on /
-- Boot f0dfcad5e18b4f4b8d5041d5b8bf4f85 --
Mar 26 05:39:21 crc chattr[847]: chattr: Operation not supported while reading flags on /

[core@crc ~]$ sudo journalctl -u create-symlink-users.service
Mar 24 14:50:33 crc systemd[1]: Starting Create Users directory and mount to var...
Mar 24 14:50:33 crc ln[1028]: ln: failed to create symbolic link '/Users': Read-only file system
Mar 24 14:50:33 crc systemd[1]: create-symlink-users.service: Main process exited, code=exited, status=1/FAILURE
Mar 24 14:50:33 crc systemd[1]: create-symlink-users.service: Failed with result 'exit-code'.
Mar 24 14:50:33 crc systemd[1]: Failed to start Create Users directory and mount to var.
[...]

PR #1047, opened 2025-04-08
snc: Add logic to create /User top level directory for OCP

praveenkumar:
Since OCP now moved to use bootc where top level directory is immutable
and `chattr -i /` doesn't work as it worked before so only option is to
create a custom-os image and deploy that as part of day-2 operation.

More details : https://github.com/crc-org/snc/issues/1041#issuecomment-2785928976

## Summary by Sourcery

Modify Single Node Cluster (SNC) setup to create a custom OS image with `/Users` directory for OCP using bootc

New Features:
- Create a custom RHCOS image with `/Users` directory to address immutable top-level directory limitations in OCP

Enhancements:
- Implement a Containerfile-based approach to create a custom OS image with `/Users` symlink

Deployment:
- Add logic to build, push, and apply a custom machine config with a new OS image

Chores:
- Remove existing 99_master-create-users-symlink.yaml configuration


snc.sh
```diff
@@ -245,6 +241,39 @@ retry ${OC} delete mc chronyd-mask
...
+podman push --authfile ${INSTALL_DIR}/reg.json --tls-verify=false default-route-openshift-image-registry.apps-crc.testing/openshift-machine-config-operator/rhcos:latest
+cat << EOF > ${INSTALL_DIR}/custom-os-mc.yaml
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: custom-image
+spec:
+  osImageURL: image-registry.openshift-image-registry.svc:5000/openshift-machine-config-operator/rhcos:latest
```
anjannath, line 267: how much does the it increase the bundle size? and can we later delete the image after machine config pool is updated?

anjannath (approved):

snc.sh
praveenkumar: Yes that's what I am exploring.

praveenkumar: I checked the images if it is stored in bundle after creating the bundle
```
rhel-coreos                                    sha256:66b5f9d8f1af5acee783db84a25c08338252ff9c276d33e391680168a22f42ee
```

but it is not available in the bundle.

PR #1052, opened 2025-04-29
gvforwarder as a systemd service

praveenkumar:
- Create a tap device using nmcli with a hardcoded mac address
- Start gvforwarder systemd service which will use this device

## Summary by Sourcery

Configure `gvforwarder` (gvisor-tap-vsock) to run as a native systemd service instead of a container.

Build:
- Create a `tap0` network interface using `nmcli` with a fixed MAC address during VM setup.
- Install the `gvforwarder` binary directly onto the host system.
- Configure and enable a native systemd service (`gv-user-network@.service`) to manage the `gvforwarder` process.
- Remove the previous container-based systemd service for `gvisor-tap-vsock`.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -16,7 +16,7 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
 OPENSHIFT_VERSION=$(${JQ} -r .clusterInfo.openshiftVersion $INSTALL_DIR/crc-bundle-info.json)
 BASE_DOMAIN=$(${JQ} -r .clusterInfo.baseDomain $INSTALL_DIR/crc-bundle-info.json)
 BUNDLE_TYPE=$(${JQ} -r .type $INSTALL_DIR/crc-bundle-info.json)
-ADDITIONAL_PACKAGES="cloud-init"
+ADDITIONAL_PACKAGES="cloud-init gvisor-tap-vsock-gvforwarder"
 
 case ${BUNDLE_TYPE} in
     microshift)
@@ -104,9 +104,6 @@ EOF
 
 # Add gvisor-tap-vsock service
 ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
-  podman create --name=gvisor-tap-vsock quay.io/crcont/gvisor-tap-vsock:latest
-  podman cp gvisor-tap-vsock:/vm /usr/local/bin/gvforwarder
-  podman rm gvisor-tap-vsock
   tee /etc/systemd/system/gv-user-network@.service <<TEE
 [Unit]
 Description=gvisor-tap-vsock Network Traffic Forwarder
@@ -118,7 +115,7 @@ After=sys-devices-virtual-net-%i.device
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
-ExecStart=/usr/local/bin/gvforwarder -preexisting -iface %i -url vsock://2:"\\\${GV_VSOCK_PORT}"/connect
+ExecStart=/usr/libexec/podman/gvforwarder -preexisting -iface %i -url vsock://2:"\\\${GV_VSOCK_PORT}"/connect
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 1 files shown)


cfergeau:
Is it the same as #1003 but against a different branch? or are there some differences?

praveenkumar:
> Is it the same as #1003 but against a different branch? or are there some differences?

It is same but against `release-4.18` branch.

createdisk.sh
```diff
@@ -94,12 +94,38 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
...
-  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  podman create --name=gvisor-tap-vsock quay.io/crcont/gvisor-tap-vsock:latest
+  podman cp gvisor-tap-vsock:/vm /usr/local/bin/gvforwarder
+  podman rm gvisor-tap-vsock
+  tee /etc/systemd/system/gv-user-network@.service <<TEE
+[Unit]
+Description=gvisor-tap-vsock Network Traffic Forwarder
+After=NetworkManager.service
+BindsTo=sys-devices-virtual-net-%i.device
+After=sys-devices-virtual-net-%i.device
```
cfergeau, line 115: Do we need to add `Before=nodeip-configuration.service` as in https://github.com/crc-org/snc/pull/1054 ?

praveenkumar: I don't know but I can try to do that also.

createdisk.sh
```diff
@@ -94,12 +94,38 @@ if podman manifest inspect quay.io/crcont/routes-controller:${OPENSHIFT_VERSION}
...
 ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
-  podman create --name=gvisor-tap-vsock --privileged --net=host -v /etc/resolv.conf:/etc/resolv.conf -it quay.io/crcont/gvisor-tap-vsock:latest
-  podman generate systemd --restart-policy=no gvisor-tap-vsock > /etc/systemd/system/gvisor-tap-vsock.service
+  podman create --name=gvisor-tap-vsock quay.io/crcont/gvisor-tap-vsock:latest
+  podman cp gvisor-tap-vsock:/vm /usr/local/bin/gvforwarder
+  podman rm gvisor-tap-vsock
+  tee /etc/systemd/system/gv-user-network@.service <<TEE
+[Unit]
+Description=gvisor-tap-vsock Network Traffic Forwarder
+After=NetworkManager.service
```
anjannath, line 113: Should this be `Before=NetworkManager.service` ? otherwise it starts very late after boot

praveenkumar: @anjannath networkManager is responsible to activate tun/tap device so it that is not even active how would this work.

cfergeau: > Should this be `Before=NetworkManager.service` ? otherwise it starts very late after boot

Is it `NetworkManager.service` which would start late? or `network-online`?
Since we have `After=sys-devices-virtual-net-%i.device` , do we even need the reference to `NetworkManager.service`?

gbraad: I spoke with Anjan, and the TAP creation is independent of the `network` which configures. But the current, `After` is too undetermined, as it can start at any point after; anything with priority or dependency goes first.

gbraad: In short, rather use `Before` than `After` statements, as that ensures the order.

createdisk.sh
```diff
@@ -16,7 +16,7 @@ INSTALL_DIR=${1:-crc-tmp-install-data}
 OPENSHIFT_VERSION=$(${JQ} -r .clusterInfo.openshiftVersion $INSTALL_DIR/crc-bundle-info.json)
 BASE_DOMAIN=$(${JQ} -r .clusterInfo.baseDomain $INSTALL_DIR/crc-bundle-info.json)
 BUNDLE_TYPE=$(${JQ} -r .type $INSTALL_DIR/crc-bundle-info.json)
-ADDITIONAL_PACKAGES="cloud-init"
+ADDITIONAL_PACKAGES="cloud-init gvisor-tap-vsock-gvforwarder"
```
gbraad, line 19: 👍

gbraad: This does mean we need to keep this package up to date along with Podman's timeline/sechedule.

praveenkumar: @gbraad This is subpackage and I think as of now it is not used by podman machine but build everytime there is release of gvisor-tap-vsock so we can just use it without maintaining (like building ourself).

PR #1065, opened 2025-05-19
build-patch: Check if brew build exist is not failed one

praveenkumar:
If brew build fails then there is no image artifacts but this doesn't make script to rerun the brew build and failed. This PR adds a check about the status of the build and make sure if the status is complete then only don't run the brew build step.

## Summary by Sourcery

Bug Fixes:
- Modify brew build existence check to also verify "State: COMPLETE" before skipping the build step.


praveenkumar:
This is tested in internal CI and works as expected.

anjannath (approved):

Issue #1068, opened 2025-05-19
Network manager restart hangs the VM


praveenkumar:
In one of the CI result we are seeing following which looks like after https://github.com/crc-org/snc/commit/d33fb16451023135968918b4edce3f38576450be it happens, I tried same bundle locally but this is not reproducible :( 
 
```
DEBU Running '/home/packer/.crc/bin/crc-admin-helper-linux-amd64 rm api.crc.testing host.crc.testing oauth-openshift.apps-crc.testing console-openshift-console.apps-crc.testing downloads-openshift-console.apps-crc.testing canary-openshift-ingress-canary.apps-crc.testing default-route-openshift-image-registry.apps-crc.testing' 
DEBU Running '/home/packer/.crc/bin/crc-admin-helper-linux-amd64 add 127.0.0.1 api.crc.testing host.crc.testing oauth-openshift.apps-crc.testing console-openshift-console.apps-crc.testing downloads-openshift-console.apps-crc.testing canary-openshift-ingress-canary.apps-crc.testing default-route-openshift-image-registry.apps-crc.testing' 
DEBU Running SSH command: systemctl status ovs-configuration.service 
DEBU SSH command results: err: Process exited with status 3, output: ○ ovs-configuration.service - Configures OVS with proper host networking configuration
     Loaded: loaded (/etc/systemd/system/ovs-configuration.service; enabled; preset: disabled)
     Active: inactive (dead) 
DEBU Using root access: Executing systemctl daemon-reload command 
DEBU Running SSH command: sudo systemctl daemon-reload 
DEBU SSH command results: err: <nil>, output:     
DEBU Using root access: Executing systemctl start ovs-configuration.service 
DEBU Running SSH command: sudo systemctl start ovs-configuration.service 
DEBU SSH command results: err: <nil>, output:     
DEBU Using root access: Update resolv.conf file   
DEBU Running SSH command: sudo nmcli con modify --temporary ovs-if-br-ex ipv4.dns 192.168.127.1 ipv4.dns-search crc.testing 
DEBU SSH command results: err: <nil>, output:     
DEBU Using root access: Executing systemctl daemon-reload command 
DEBU Running SSH command: sudo systemctl daemon-reload 
DEBU SSH command results: err: <nil>, output:     
DEBU Using root access: Executing systemctl restart NetworkManager.service 
DEBU Running SSH command: sudo systemctl restart NetworkManager.service
```

cfergeau:
[...]

PR #1069, opened 2025-05-19
CI: Use debug as log level for crc start

praveenkumar:
Looks like without debug message it is hard to find which stage actually failed or hanged, took some time to figure out start hangs because of NetworkManager instead Configure shared directory.

- https://github.com/crc-org/snc/issues/1068

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -64,7 +64,7 @@ popd
 
 crc config set bundle crc_libvirt_*.crcbundle
 crc setup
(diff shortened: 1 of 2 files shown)


cfergeau:
>  took some time to figure out start hangs because of NetworkManager instead Configure shared directory.

We could also add more "info" message during crc startup to help pinpoint where issues happen

praveenkumar:
> We could also add more "info" message during crc startup to help pinpoint where issues happen

sure but that would be PR from crc side.

PR #1071, opened 2025-05-20
Disable the ipv6 for tap0 connection

praveenkumar:
During Network restart looks like it also try to assign ipv6 for tap0 so disabling it.

## Summary by Sourcery

Bug Fixes:
- Disable IPv6 on the tap0 connection to avoid unexpected IPv6 addressing during network restarts


cfergeau (approved):
I  wonder if there are some improvements gvisor-tap-vsock could do to cope better with network manager trying to use ipv6, but looks good to me if this fixes the restart issues.

PR #1064, opened 2025-05-16
add start condition for post start systemd units

anjannath:

PR #1073, opened 2025-05-20
Don't add device checks for tap0 service

praveenkumar:
Instead use `ExecStartPre` to check if tap0 exist with a retry loop for 10 sec. During failure `timeout waiting for device /sys/class/net/tap0` error is observed which means that device was not available and check failed. so instead of device check let's see if this make it work.

Let's see if that fix the CI.

## Summary by Sourcery

Replace systemd templated gv-user-network unit with a dedicated gvisor-tap-vsock.service that waits for the tap0 interface via an ExecStartPre retry loop instead of relying on device unit bindings and update service installation accordingly.

Enhancements:
- Rename templated gv-user-network@.service to a single gvisor-tap-vsock.service
- Remove BindsTo and After device unit dependencies for tap0
- Add an ExecStartPre loop to retry checking for the tap0 interface up to 10 seconds
- Update ExecStart to use a fixed tap0 interface and adjust systemctl enable invocation

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -107,14 +107,13 @@ ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
   tee /etc/systemd/system/gv-user-network@.service <<TEE
 [Unit]
 Description=gvisor-tap-vsock Network Traffic Forwarder
-After=NetworkManager.service
-BindsTo=sys-devices-virtual-net-%i.device
 After=sys-devices-virtual-net-%i.device
 
 [Service]
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecStartPre=/bin/sh -c 'for i in {1..10}; do ip link show "\\\$1" && exit 0; sleep 1; done; exit 1' _ %i
 ExecStart=/usr/libexec/podman/gvforwarder -preexisting -iface %i -url vsock://2:"\\\${GV_VSOCK_PORT}"/connect
 
 [Install]


createdisk.sh
```diff
@@ -104,25 +104,24 @@ EOF
...
-BindsTo=sys-devices-virtual-net-%i.device
-After=sys-devices-virtual-net-%i.device
 
 [Service]
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
-ExecStart=/usr/libexec/podman/gvforwarder -preexisting -iface %i -url vsock://2:"\\\${GV_VSOCK_PORT}"/connect
+ExecStartPre=/bin/sh -c 'for i in {1..10}; do ip link show tap0 && exit 0; sleep 1; done; exit 1'
+ExecStart=/usr/libexec/podman/gvforwarder -preexisting -iface tap0 -url vsock://2:"\\\${GV_VSOCK_PORT}"/connect
```
cfergeau, line 117: I thought podman was using a unit file similar to ours, but what they have is a NM configuration file for `tap0` (they call it `vsock0`) which is not autostarted, and then in `ExecStartPost` (so after starting `gvforwarder`), they explicitly call `nmcli conn up vsock0` https://github.com/containers/podman/blob/main/pkg/machine/hyperv/stubber.go#L523-L550

praveenkumar: In our case we do need to make sure it is autostarted because during our execution we make change to resolve.conf after running `ovs-configuration` and restart NM so if it is not autostarted then in our case tap0 fails to start.

anjannath:
rebased the #1064 PR on this one, and that also passed the CI finally, https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/crc-org_snc/1064/pull-ci-crc-org-snc-release-4.18-e2e-snc/1925493930065924096

createdisk.sh
```diff
@@ -108,13 +108,12 @@ ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
...
 Description=gvisor-tap-vsock Network Traffic Forwarder
 After=NetworkManager.service
-BindsTo=sys-devices-virtual-net-%i.device
-After=sys-devices-virtual-net-%i.device
 
 [Service]
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecStartPre=/bin/sh -c 'for i in {1..10}; do ip link show "\\\$1" && exit 0; sleep 1; done; exit 1' _ %i
```
cfergeau, line 116: Imo, `After` should achieve this in a better way. You can keep a `ExecStartPre` check which makes sure `%i` is present if you want better error reporting.

The commit log could be updated, with this change, `gv-user-network@.service` will not be restarted when NetworkManager restarts, which explains why the issues we were getting when `gv-user-network@.service` restarts will not be seen.

praveenkumar: Done, so as per man page of `systemd.netdev` looks like `systemd-network-generator.service is responsible for creating those devices and this make sure the device is available.

praveenkumar: > > Done, so as per man page of `systemd.netdev` looks like `systemd-network-generator.service is responsible for creating those devices and this make sure the device is available.
> 
> I see nothing related to this in the comments or in the commit:

This is just an info, I wanted to check how all it works :)

cfergeau:
Err, this was a NACK from me as it currently was. This  was still tagged WIP when I looked yesterday.

PR #1087, opened 2025-05-29
Fixes okd bits which make sure okd build happen successfully 

praveenkumar:
## Summary by Sourcery

Update OKD-specific automation to ensure build and installation succeed by standardizing client retrieval, VM configuration, mirror URL, and image selection.

Bug Fixes:
- Unify Windows oc client download via curl instead of conditional oc adm extract for OKD bundles.

Enhancements:
- Make UEFI boot optional and disable it for OKD VMs.
- Update default OKD mirror URL to github.com/okd/releases.
- Select ‘stream-coreos’ image for OKD clusters instead of ‘rhel-coreos’.

### DIFF
--- a/tools.sh
+++ b/tools.sh
@@ -195,6 +195,11 @@ function create_libvirt_resources {
 function create_vm {
     local iso=$1
 
+    bootOption=""
+    if ${BUNDLE_TYPE} != "okd"; then
+        bootOption="--boot uefi"
+    fi
+
     sudo virt-install \
         --name ${SNC_PRODUCT_NAME} \
         --vcpus ${SNC_CLUSTER_CPUS} \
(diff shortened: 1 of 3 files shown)


tools.sh
```diff
@@ -207,7 +212,7 @@ function create_vm {
         --cdrom /var/lib/libvirt/${SNC_PRODUCT_NAME}/${iso} \
         --events on_reboot=restart \
         --autoconsole none \
-        --boot uefi \
+        ${bootOption} \
```
anjannath, line 215: i think in crc also we create the VM and uefi boot loader, at least for macOS, will the bundle work in that case?

praveenkumar: @anjannath OKD doesn't have arm64 support and iirc it was required for that.

PR #1088, opened 2025-05-30
Add missing services for self sufficient bundle

anjannath:
this adds systemd services to perform the following tasks:
- service that waits for api server to become available
  (this will help to schedule other services after this)
- service to delete MCO leader leases
- service to check if pull-secret was written to disk

## Summary by Sourcery

Add missing systemd services and enhance the common helper script to support a fully self-sufficient bundle by waiting for the API server, cleaning up MCO leases, and ensuring the pull-secret is written to disk.

New Features:
- Add systemd service to wait for the API server to become available
- Add systemd service to delete machine-config-operator leader leases
- Add systemd service to verify the pull-secret file is present on disk

Enhancements:
- Parameterize the wait interval in the common wait_for_resource helper function

### DIFF
--- a/systemd/crc-cluster-status.service
+++ b/systemd/crc-cluster-status.service
@@ -1,7 +1,9 @@
 [Unit]
 Description=CRC Unit checking if cluster is ready
-After=kubelet.service ocp-clusterid.service ocp-cluster-ca.service ocp-custom-domain.service
-After=crc-pullsecret.service
+After=crc-wait-apiserver-up.service crc-pullsecret.service
+After=ocp-mco-sshkey.service ocp-cluster-ca.service
+After=ocp-custom-domain.service ocp-userpasswords.service
+After=ocp-clusterid.service
 
 [Service]
 Type=oneshot
--- a/systemd/crc-cluster-status.service
+++ b/systemd/crc-cluster-status.service
@@ -4,10 +4,13 @@ After=crc-wait-apiserver-up.service crc-pullsecret.service
 After=ocp-mco-sshkey.service ocp-cluster-ca.service
 After=ocp-custom-domain.service ocp-userpasswords.service
 After=ocp-clusterid.service
+StartLimitIntervalSec=450
+StartLimitBurst=10
 
 [Service]
 Type=oneshot
 Restart=on-failure
+RestartSec=40
 ExecCondition=/usr/local/bin/crc-check-cloud-env.sh
 ExecStart=/usr/local/bin/crc-cluster-status.sh
 RemainAfterExit=true
--- a/systemd/crc-pullsecret.sh
+++ b/systemd/crc-pullsecret.sh
@@ -14,7 +14,7 @@ existingPs=$(echo "${existingPsB64}" | base64 -d)
 echo "${existingPs}" | jq -e '.auths'
 
 if [[ $? != 0 ]]; then
-    pullSecretB64=$(cat /opt/crc/pull-secret | base64 -w0)
+    pullSecretB64=$(base64 -w0 < /opt/crc/pull-secret)
     oc patch secret pull-secret -n openshift-config --type merge -p "{\"data\":{\".dockerconfigjson\":\"${pullSecretB64}\"}}"
 fi
 
--- a/systemd/crc-pullsecret.sh
+++ b/systemd/crc-pullsecret.sh
@@ -16,6 +16,5 @@ echo "${existingPs}" | jq -e '.auths'
 if [[ $? != 0 ]]; then
     pullSecretB64=$(cat /opt/crc/pull-secret | base64 -w0)
     oc patch secret pull-secret -n openshift-config --type merge -p "{\"data\":{\".dockerconfigjson\":\"${pullSecretB64}\"}}"
-    rm -f /opt/crc/pull-secret
 fi
 
--- a/systemd/ocp-cluster-ca.sh
+++ b/systemd/ocp-cluster-ca.sh
@@ -11,73 +11,76 @@ export KUBECONFIG="/opt/kubeconfig"
 
 wait_for_resource configmap
 
-custom_ca_path=/opt/crc/custom-ca.crt
 external_ip_path=/opt/crc/eip
-
-if [ ! -f ${custom_ca_path} ]; then
-    echo "Cert bundle /opt/crc/custom-ca.crt not found, generating one..."
-    # generate a ca bundle and use it, overwrite custom_ca_path
-    CA_SUBJ="/OU=openshift/CN=admin-kubeconfig-signer-custom"
-    openssl genrsa -out /tmp/custom-ca.key 4096
-    openssl req -x509 -new -nodes -key /tmp/custom-ca.key -sha256 -days 365 -out "${custom_ca_path}" -subj "${CA_SUBJ}"
-fi
 
 if [ ! -f /opt/crc/pass_kubeadmin ]; then
     echo "kubeadmin password file not found"
     exit 1
 fi
 
+if oc get configmap client-ca-custom -n openshift-config; then
+    echo "API Server Client CA already rotated..."
+    exit 0
+fi
+
+# generate CA
+CA_FILE_PATH="/tmp/custom-ca.crt"
+CA_KEY_FILE_PATH="/tmp/custom-ca.key"
+CLIENT_CA_FILE_PATH="/tmp/client-ca.crt"
+CLIENT_CA_KEY_FILE_PATH="/tmp/client-ca.key"
+CLIENT_CSR_FILE_PATH="/tmp/client-csr.csr"
+CA_SUBJ="/OU=openshift/CN=admin-kubeconfig-signer-custom"
+CLIENT_SUBJ="/O=system:masters/CN=system:admin"
+VALIDITY=365
 PASS_KUBEADMIN="$(cat /opt/crc/pass_kubeadmin)"
-oc create configmap client-ca-custom -n openshift-config --from-file=ca-bundle.crt=${custom_ca_path}
-oc patch apiserver cluster --type=merge -p '{"spec": {"clientCA": {"name": "client-ca-custom"}}}'
-oc create configmap admin-kubeconfig-client-ca -n openshift-config --from-file=ca-bundle.crt=${custom_ca_path} \
-    --dry-run=client -o yaml | oc replace -f -
 
+# generate the CA private key
+openssl genrsa -out ${CA_KEY_FILE_PATH} 4096
+# Create the CA certificate
+openssl req -x509 -new -nodes -key ${CA_KEY_FILE_PATH} -sha256 -days $VALIDITY -out ${CA_FILE_PATH} -subj "${CA_SUBJ}"
 # create CSR
-openssl req -new -newkey rsa:4096 -nodes -keyout /tmp/newauth-access.key -out /tmp/newauth-access.csr -subj "/CN=system:admin"
+openssl req -new -newkey rsa:4096 -nodes -keyout ${CLIENT_CA_KEY_FILE_PATH} -out ${CLIENT_CSR_FILE_PATH} -subj "${CLIENT_SUBJ}"
+# sign the CSR with above CA
+openssl x509 -extfile <(printf "extendedKeyUsage = clientAuth") -req -in ${CLIENT_CSR_FILE_PATH} -CA ${CA_FILE_PATH} \
+    -CAkey ${CA_KEY_FILE_PATH} -CAcreateserial -out ${CLIENT_CA_FILE_PATH} -days $VALIDITY -sha256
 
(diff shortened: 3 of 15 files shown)


systemd/ocp-delete-mco-leases.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
praveenkumar, line 1: we don't need that because we are doing it as part of bundle creation https://github.com/crc-org/snc/blob/release-4.18/createdisk.sh#L174-L185

systemd/ocp-growfs.service
```diff
@@ -1,5 +1,6 @@
 [Unit]
 Description=CRC Unit to grow the root filesystem
+Before=crc-dnsmasq.service
```
praveenkumar, line 3: does it really need any service like before/after since it is mostly on filesystem side change?

anjannath: updated!

systemd/crc-cluster-status.service
```diff
@@ -4,10 +4,13 @@ After=crc-wait-apiserver-up.service crc-pullsecret.service
 After=ocp-mco-sshkey.service ocp-cluster-ca.service
 After=ocp-custom-domain.service ocp-userpasswords.service
 After=ocp-clusterid.service
+StartLimitIntervalSec=450
+StartLimitBurst=10
```
praveenkumar, line 8: > values such that the service  never hits the re-start limit and is always re-started

@anjannath Is there any source/doc/link which states that with using above limits service never hits re-start limit?

systemd/ocp-cluster-ca.sh
```diff
@@ -15,11 +15,8 @@ custom_ca_path=/opt/crc/custom-ca.crt
 external_ip_path=/opt/crc/eip
 
 if [ ! -f ${custom_ca_path} ]; then
-    echo "Cert bundle /opt/crc/custom-ca.crt not found, generating one..."
-    # generate a ca bundle and use it, overwrite custom_ca_path
-    CA_SUBJ="/OU=openshift/CN=admin-kubeconfig-signer-custom"
-    openssl genrsa -out /tmp/custom-ca.key 4096
-    openssl req -x509 -new -nodes -key /tmp/custom-ca.key -sha256 -days 365 -out "${custom_ca_path}" -subj "${CA_SUBJ}"
```
praveenkumar, line 22: does that mean we don't need to generated the custom ca bundle key and it is user's who should take care of that? I am not sure how  this is part of this commit, may be a separate one?

systemd/crc-cluster-status.service
anjannath: yes these two are covered in the `systemd.unit` man page, https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#StartLimitIntervalSec=interval and the `RestartSec` is covered in the `systemd.service` man page.. https://www.freedesktop.org/software/systemd/man/latest/systemd.service#RestartSec=

`StartLimitIntervalSec` sets the amount of time (in seconds) within which a service is allowed to restart, the `StartLimitBurst` controls how many re-starts are allowed within the time set by `StartLimitInterValSec` and the `RestartSec` sets the amount of time to wait before a restart, so if we set the `StartLimitIntervalSec` greater than `StartLimitBurst * RestartSec` then we'll never hit the restart limit which prevents further restarts

systemd/ocp-cluster-ca.sh
anjannath: Yeah, when using the bundle with `mapt` it is created by `mapt` with crc or directly using with vfkit,libvirt users have to create this.. we can automate it when using with `crc`

systemd/crc-cluster-status.service
praveenkumar: In this case if service fails then it will again try to restart after `StartLimitIntervalSec+RestartSec` which is around 490sec ~ 8min ?

Edit: so looks like below explanation as per per AI so in our case it sets `450sec` for interval window and during that window it will try it 10 times only and finally fails if even after 10 retry service is failing.
```
StartLimitIntervalSec=

    Sets the time window (in seconds) for counting starts.

    Default: 10s (if not set).

StartLimitBurst=

    Sets the maximum number of service starts allowed within the StartLimitIntervalSec time window.

    If the service is restarted more than this number within the interval, systemd will stop trying and mark the service as failed.
```

systemd/ocp-cluster-ca.sh
anjannath: if there's actually no use of having the cert and the ca key being supplied from outside by the user/tool, i.e we only need to access the cluster for which the kubeconfig is enough, then i think we can remove this option and always generate the cert in the VM and not allow users to supply it, wdyt?

systemd/crc-cluster-status.service
anjannath: no, it'll retry after sleeping for `RestartSec` which is 40 seconds.. and it'll retry 10 times.. so in total 400 seconds.. but the limit is 450 seconds, so it can again restart since it didn't cross the start limit interval

systemd/crc-pullsecret.sh
```diff
@@ -16,6 +16,5 @@ echo "${existingPs}" | jq -e '.auths'
 if [[ $? != 0 ]]; then
     pullSecretB64=$(cat /opt/crc/pull-secret | base64 -w0)
     oc patch secret pull-secret -n openshift-config --type merge -p "{\"data\":{\".dockerconfigjson\":\"${pullSecretB64}\"}}"
-    rm -f /opt/crc/pull-secret
```
cfergeau, line 19: Maybe these can be removed in an `ExecStartPost`  or in a unit/service which would be run after everything is successful?

systemd/ocp-cluster-ca.sh
anjannath: i think it makes sense to not let the users provide a CA to be used and always generate it inside the bundle

systemd/ocp-cluster-ca.sh
```diff
@@ -11,72 +11,51 @@ export KUBECONFIG="/opt/kubeconfig"
...
-until `openssl x509 -in /tmp/newauth-access.crt -noout`
-do
-    oc get csr newauth-access -o jsonpath='{.status.certificate}' | base64 -d > /tmp/newauth-access.crt
-    sleep 2
-done
 oc config set-credentials system:admin --client-certificate=/tmp/newauth-access.crt --client-key=/tmp/newauth-access.key --embed-certs --kubeconfig="${updated_kubeconfig_path}"
 oc config set-context system:admin --cluster="${cluster_name}" --namespace=default --user=system:admin --kubeconfig="${updated_kubeconfig_path}"
-oc get secret localhost-recovery-client-token -n openshift-kube-controller-manager -ojsonpath='{.data.ca\.crt}'| base64 -d > /tmp/bundle-ca.crt
-oc config set-cluster "${cluster_name}" --server="${apiserver_url}" --certificate-authority=/tmp/bundle-ca.crt \
+#oc get secret localhost-recovery-client-token -n openshift-kube-controller-manager -ojsonpath='{.data.ca\.crt}'| base64 -d > /tmp/bundle-ca.crt
```
praveenkumar: if it is commented should we just remove it?

systemd/ocp-cluster-ca.sh
```diff
@@ -11,72 +11,51 @@ export KUBECONFIG="/opt/kubeconfig"
...
-
 if [ ! -f /opt/crc/pass_kubeadmin ]; then
     echo "kubeadmin password file not found"
     exit 1
 fi
 
-# base64 decode the cert
-base64 -d < "${custom_ca_path}" > /tmp/custom-kubeadmin-ca.crt
-custom_ca_path=/tmp/custom-kubeadmin-ca.crt
+# generate CA
```
praveenkumar, line 26: We should mention as comment that those steps are part of https://access.redhat.com/solutions/6054981 so that in future, it can be referred

anjannath: there's a comment listing the links at the top of the file https://github.com/crc-org/snc/pull/1088/files#diff-57ea196af424972bc0595d747b0f0a60be2f3ac197b41b1b357a6a173cde7abaL4-L5

systemd/crc-pullsecret.sh
anjannath: the pull secret file is also present in `/var/lib/kubelet/pullsecret.json` and then the other two files that i thought of removing were `/opt/crc/pass_kubeadmin` and `/opt/crc/pass_developer` but these are a bit tricky to remove as two services needs the kubeadmin password, but since the order of them starting is not known before, sometimes the file is removed before one of the dependent services has got the chance to start

we can remove them in the `ExecStartPost` of `crc-cluster-status.service` or we can add another `crc-cleanup.service` that runs after all of the other services

systemd/ocp-cluster-ca.sh
```diff
@@ -53,10 +53,10 @@ updated_kubeconfig_path=/opt/crc/kubeconfig
 
 oc config set-credentials system:admin --client-certificate=${CLIENT_CA_FILE_PATH} --client-key=${CLIENT_CA_KEY_FILE_PATH} --embed-certs --kubeconfig="${updated_kubeconfig_path}"
 oc config set-context system:admin --cluster="${cluster_name}" --namespace=default --user=system:admin --kubeconfig="${updated_kubeconfig_path}"
-oc config set-cluster "${cluster_name}" --server="${apiserver_url}" --certificate-authority=$CA_FILE_PATH \
+oc config set-cluster "${cluster_name}" --server="${apiserver_url}" --insecure-skip-tls-verify=true \
     --kubeconfig="${updated_kubeconfig_path}" --embed-certs
 
-if oc get co --kubeconfig="${updated_kubeconfig_path}"; then
+if `oc get co --kubeconfig="${updated_kubeconfig_path}"`; then
```
praveenkumar: is this really required because `if oc get co --kubeconfig="${updated_kubeconfig_path}"` suppose to do same?

systemd/crc-cluster-status.service
```diff
@@ -13,6 +13,9 @@ Restart=on-failure
 RestartSec=40
 ExecCondition=/usr/local/bin/crc-check-cloud-env.sh
 ExecStart=/usr/local/bin/crc-cluster-status.sh
+ExecStartPost=/usr/bin/rm -rf /opt/crc/pull-secret
+ExecStartPost=/usr/bin/rm -rf /opt/crc/pass_developer
+ExecStartPost=/usr/bin/rm -rf /opt/crc/pass_kubeadmin
```
praveenkumar: why are removing those so that stop start it doesn't start those service?

anjannath: no it is just for cleanup, but as a side effect the two services will also be marked as *failed* after stop->start because of the missing files 

earlier these were removed in the service's script but that had race condition as the `pass_kubeadmin` file is used by two services, so we are now removing these at the very end in this service as it is supposed to run after all the other services

praveenkumar: if we remove those and then user start the cloud instance first time it pass the cloud-init file which will put all those files like pull-secret, pass_developer ..etc to right place. But then user done using this instance and stop it using cloud cli or from web. If it is again start this instance and this time there is no cloud-init run doesn't it make the services fails ?

anjannath: yes the service will be marked as failed but that is fine and cluster will also work as expected, since the required setup was already done in the first boot

and as we are planning to add some mechanism to prevent running the services again after stop -> start this failure will also not happen after this is added

cfergeau: Maybe make the service conditional on the existence of these files so that systemd does not happen to start the service if these are missing?

anjannath: while testing i found that using `ConditionPathExists` doesn't respect `Restart=on-failure`, and in our case the units always has to restart several times before those files are created by `cloud-init` , so when the service starts before cloud-init and the files are not yet present it'll not be restarted again 

(maybe we can schedule the self sufficient services that needs the files to start after `cloud-init-finale.service` and get around this issue)

anjannath:
@praveenkumar tested and working as expected now, can you also give it a try with `mapt` : `mapt aws openshift-snc create --spot --version 4.18.16 --project-name anath-mapt-snc --backed-url file:///<workspace> --conn-details-output /tmp/snc --pull-secret-file "<pull-secret-file-path>" --ca-cert-file /tmp/custom-ca.crt --debug`

systemd/crc-cluster-status.service
praveenkumar: If those files are around in the VM does it create some issue if not then why we want to remove those, I know it is sensitive info but anyway those are already part of the VM?

anjannath: yes they are already in the VM in some form, the pull secret is as it is, the user passwords maybe base64 encoded, but other than removing them because they are sensitive, there is no other reason,  i can drop this commit!

PR #1091, opened 2025-06-03
Check for openshift-install version

eb4x:
The `oc adm release extract` command is also quite time consuming when running multiple times. So introduce a check to see if we even need to get the openshift-install binary.

I think checking the version of the binary is probably a good idea to do in any case, so this could be further improved upon by dropping the -z test, and do something like

```shell
if [[ ! -f "${OPENSHIFT_INSTALL:=./openshift-install}" ... ]]; then
```
(If it's provided verify the file exists AND the version is what we're expecting.) I could make the code move the extracted binary to the path provided by OPENSHIFT_INSTALL in that case.

## Summary by Sourcery

Enhancements:
- Add existence and version check for the openshift-install binary and only extract when missing or outdated

### DIFF
--- a/snc.sh
+++ b/snc.sh
@@ -68,9 +68,11 @@ echo "Setting OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE to ${OPENSHIFT_INSTALL_RE
 
 # Extract openshift-install binary if not present in current directory
 if test -z ${OPENSHIFT_INSTALL-}; then
-    echo "Extracting OpenShift installer binary"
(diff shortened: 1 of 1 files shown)


eb4x:
@praveenkumar What do you think of always verifying the version reported by `OPENSHIFT_INSTALL` against `OPENSHIFT_VERSION` ?

eb4x:
> > @praveenkumar What do you think of always verifying the version reported by `OPENSHIFT_INSTALL` against `OPENSHIFT_VERSION` ?
> 
> @eb4x yes, it will make sure we are using same version of installer as provided in release payload.

No, I mean if `OPENSHIFT_INSTALL` is already set externally, would we still want to verify the version then?
if someone does `export OPENSHIFT_INSTALL=$HOME/.local/bin/openshift-install` before running `snc.sh` then this bit of shell code wouldn't kick in. I'm wondering if we'd want to always verify the version, and if it doesn't match overwrite the "external" file, or just stop the script?

PR #1118, opened 2025-06-30
Add DISABLE_VIRT_CAPABILITIES_CHECKS variable

praveenkumar:
Useful for openshift CI environments or systems where KVM is not available,
allowing the scripts to proceed using the QEMU emulator instead.

## Summary by Sourcery

Enable conditional disabling of hardware-accelerated hypervisor checks via a new environment variable and configure it for arm64 CI environments

New Features:
- Introduce DISABLE_VIRT_CAPABILITIES_CHECKS environment variable to allow skipping KVM hypervisor capability checks

Enhancements:
- Apply the new flag in snc-library.sh to conditionally skip preflight virtualization checks
- Add default for DISABLE_VIRT_CAPABILITIES_CHECKS in tools.sh

CI:
- Automatically set SNC_DISABLE_VIRT_CAPABILITIES_CHECK=true for aarch64 in ci_microshift.sh


cfergeau:
/hold
(I'm on PTO today and I think this deserves a wider discussion, so don't merge this right away :)

cfergeau:
This is close to the discussion we had in https://github.com/containers/gvisor-tap-vsock/pull/508 , in general we don't want to rely on emulation for "important" workloads. Not sure we can fully trust the results from this when it fails (could be a qemu emulation bug), and this could be slow?
Can we guarantee somehow that we'll be running on machines with virt support?

praveenkumar:
> Can we guarantee somehow that we'll be running on machines with virt support?

@cfergeau As per internal discussion also, this variable is disable by default and only using as part of CI run becasue GCP instances on arm64 doesn't have nested virtualization support and it makes it to have bundles for arm64 also. For actual deliverables (arm64 bundles for linux/mac) we are are running it without disabling virt support check.

> Not sure we can fully trust the results from this when it fails (could be a qemu emulation bug), and this could be slow?

Right, if it there is failure because of emulation/slow we do need to ignore. I am hoping this would work most of time but if this test become red more than green then we might remove it all together until we have system with virt supports.

cfergeau:
> > Can we guarantee somehow that we'll be running on machines with virt support?
> 
> @cfergeau As per internal discussion also, this variable is disable by default and only using as part of CI run becasue GCP instances on arm64 doesn't have nested virtualization support and it makes it to have bundles for arm64 also. For actual deliverables (arm64 bundles for linux/mac) we are are running it without disabling virt support check.
> 
> > Not sure we can fully trust the results from this when it fails (could be a qemu emulation bug), and this could be slow?
> 
> Right, if it there is failure because of emulation/slow we do need to ignore. I am hoping this would work most of time but if this test become red more than green then we might remove it all together until we have system with virt supports.

CI is flakey enough, I’d rather not add another potential failure point for now :-/

praveenkumar:
/close

I am going to remove the arm64 job from openshift/release side.

PR #1122, opened 2025-07-04
createdisk: Add cloud-utils-growpart package

praveenkumar:
This package provides the growpart script for growing a partition. It is primarily used in cloud images in conjunction with the dracut-modules-growroot package to grow the root partition on first boot. It suppose to solve the issue for cloud deployment of crc using mapt when specify disk size bigger than default size(30G) by auto resize the disk.

- https://github.com/canonical/cloud-utils/

mapt issue: https://github.com/redhat-developer/mapt/issues/524

## Summary by Sourcery

New Features:
- Add cloud-utils-growpart to the list of additional packages for automatic root partition expansion on boot.

## Summary by CodeRabbit

* **Chores**
  * Updated the disk creation process to include an additional package for enhanced guest VM image capabilities.
  * Adjusted container image package installation to optimize included components.


praveenkumar:
/hold

we first need to test created bundle to see if that expand the disk size automatic.

cfergeau (approved):

praveenkumar:
for microshift we have this package already https://github.com/crc-org/snc/blob/release-4.19/image-mode/microshift/config/Containerfile.bootc-rhel9#L17 so may be we should remove it from there.

PR #1123, opened 2025-07-04
add 9p fuse client

redbeam:
9P support for filesharing on Windows: https://github.com/crc-org/crc/issues/4168


createdisk.sh
```diff
@@ -163,7 +163,17 @@ EOF
...
+# set up 9p file sharing on Windows
+if [ "${SNC_GENERATE_WINDOWS_BUNDLE}" != "0" ]; then
+    ${SSH} core@${VM_IP} -- "sudo dnf -y copr enable mskvarla/9pfs"
+    ${SSH} core@${VM_IP} -- "mkdir -p ~/packages && dnf download --downloadonly --downloaddir ~/packages 9pfs --resolve"
+    ${SSH} core@${VM_IP} -- "sudo dnf -y copr disable mskvarla/9pfs"
+    PRE_DOWNLOADED_ADDITIONAL_PACKAGES+=" 9pfs"
+fi
+
+# Beyond this point, packages added to the ADDITIONAL_PACKAGES and PRE_DOWNLOADED_ADDITIONAL_PACKAGES
+# variables won’t be installed in the guest
```
praveenkumar, line 176: What is our plan around it, always going to use the copr or it is just for initial POC side? As per https://download.copr.fedorainfracloud.org/results/mskvarla/9pfs/srpm-builds/09226159/9pfs.spec file looks like we are just building the binary, can we migrate to container image for this which can be built on brew internally?

redbeam: Initially yes, I think the COPR repo should be enough to have this merged, after we see it works, we can move to brew.

praveenkumar:
@redbeam both commit should be squashed because you are replacing container build binary to rpm based one.

createdisk.sh
```diff
@@ -163,7 +163,17 @@ EOF
    PRE_DOWNLOADED_ADDITIONAL_PACKAGES+=" qemu-user-static-x86"
 fi
 
-# Beyond this point, packages added to the ADDITIONAL_PACKAGES variable won’t be installed in the guest
+# install 9pfs binary from COPR repo so that it can be used to
+# set up 9p file sharing on Windows
+if [ "${SNC_GENERATE_WINDOWS_BUNDLE}" != "0" ]; then
```
anjannath, line 168: when we generate bundles in the downstream CI job for release, it always generates bundles for all the platforms, maybe we can even remove this check later to allow use of 9pfs on the other platforms as well

praveenkumar: currently we want this only for windows right? on linux and mac we have virtio-fs working as expected.

redbeam: yes, this is currently needed only for windows

anjannath: ahh, so this is only installed for the hyper-v bundles, then its fine.. i thought this is downloaded and part of all the bundles, since we generate the libvirt bundle first then convert that to the different formats

redbeam: It could be added to other bundles, but then it has no use. It doesn't break anything and is very small, but still unused, so better to have it only in HyperV bundles.

anjannath: right now this package is included in the all the bundles, so the check is a bit misleading, the `SNC_GENERATE_{WINDOWS|MACOS}_BUNDLE` env variables is only useful if you want to skip generating the bundle for a platform

redbeam: Ahh you're right, but we can still skip installing it when Windows bundles are not generated, no? Or is there a way to install it only in hyperv bundles?

anjannath: yeah, its fine to keep, and i don't think there's a way to only install it for hyper-v bundles

redbeam:
Closing this in favor of #1193.

PR #1125, opened 2025-07-10
Introduce PRE_DOWNLOADED_ADDITIONAL_PACKAGES variable

praveenkumar:
Having only ADDITIONAL_PACKAGES variable and then adding the packages to it which are already downloaded because those comes from different repo makes the whole installation of additional packages logic break because in that logic we first download those additional packages using yum download then then create local repo and finally push it to VM for installation. Since some packages like qemu-user-static is comes from fedora repo and we just can't enable this because it might affect other packages and we hit situation like https://github.com/crc-org/snc/commit/718a9c4f9c3e65128a2d0c4f707e3c109a3182db one.

So this PR separate addtional packages which can be downloaded from official repo to other packages which comes from different repo but at the installation time it should be creating single layer on ostree side.

## Summary by Sourcery

Introduce PRE_DOWNLOAD_ADDITIONAL_PACKAGES to separate packages that require pre-download from external repos and include them during the rpm-ostree installation in a single layer.

Enhancements:
- Add PRE_DOWNLOAD_ADDITIONAL_PACKAGES variable to track pre-downloaded external packages
- Include PRE_DOWNLOAD_ADDITIONAL_PACKAGES in the rpm-ostree install command alongside ADDITIONAL_PACKAGES

## Summary by CodeRabbit

* **Chores**
  * Improved handling of additional package installation by introducing a separate variable for pre-downloaded packages.
  * Adjusted package assignment for specific architectures to ensure correct packages are installed or pre-downloaded as intended.

### DIFF
--- a/createdisk-library.sh
+++ b/createdisk-library.sh
@@ -202,7 +202,7 @@ enabled=1
 gpgcheck=0
 EOF'"
         # Install these rpms to VM
-        ${SSH} core@${vm_ip} -- "sudo rpm-ostree install $ADDITIONAL_PACKAGES"
+        ${SSH} core@${vm_ip} -- "sudo rpm-ostree install $ADDITIONAL_PACKAGES $PRE_DOWNLOADED_ADDITIONAL_PACKAGES"
 
         # Remove the packages and repo from VM
         ${SSH} core@${vm_ip} -- sudo rm -fr /home/core/packages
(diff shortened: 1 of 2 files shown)


createdisk-library.sh
```diff
@@ -202,7 +202,7 @@ enabled=1
 gpgcheck=0
 EOF'"
         # Install these rpms to VM
-        ${SSH} core@${vm_ip} -- "sudo rpm-ostree install $ADDITIONAL_PACKAGES"
+        ${SSH} core@${vm_ip} -- "sudo rpm-ostree install $ADDITIONAL_PACKAGES $PRE_DOWNLOAD_ADDITIONAL_PACKAGES"
```
cfergeau: Maybe name this `ADDITIONAL_LOCAL_PACKAGES`? If you go with this naming, `ADDITIONAL_PACKAGES` could be renamed to `ADDITIONAL_DNF_PACKAGES`
If you want to keep the current name, this should be `PRE_DOWNLOADED_ADDITIONAL_PACKAGES`

praveenkumar:
@coderabbitai review

createdisk-library.sh
praveenkumar: >If you want to keep the current name, this should be PRE_DOWNLOADED_ADDITIONAL_PACKAGES

I will go with that and in my mind I had `PRE_DOWNLOADED_..` only :)

PR #1131, opened 2025-07-29
Fix unit target for crc/ocp service

praveenkumar:
This makes sure services are not depend on multi-user target and
cloud-final service able to run in time because as of now all our
services are making multi-user target to wait and indirectly cloud-final
service was also waiting. with this we get following dep graph
```
$ systemctl list-dependencies crc-custom.target
crc-custom.target
● ├─crc-cluster-status.service
○ ├─crc-pullsecret.service
○ ├─crc-routes-controller.service
○ ├─crc-wait-apiserver-up.service
○ ├─ocp-cluster-ca.service
○ ├─ocp-clusterid.service
○ ├─ocp-custom-domain.service
● ├─ocp-mco-sshkey.service
○ ├─ocp-userpasswords.service
● └─kubelet-dependencies.target
●   ├─chrony-wait.service
[---]
```

## Summary by CodeRabbit

* **New Features**
  * Introduced a new custom systemd target for managing service startup order.

* **Chores**
  * Updated multiple services to use the new custom systemd target instead of the default target, improving control over service activation.
  * Added a dependency to ensure certain services are activated alongside the new custom target.
  * Enhanced deployment scripts to copy custom systemd target files to the VM for consistent service management.
  * Added a startup condition to a network service to verify environment suitability before activation.
  * Increased retry attempts for cluster API server readiness checks to improve robustness.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -114,6 +114,7 @@ After=sys-devices-virtual-net-%i.device
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecCondition=/bin/sh -c '! /usr/local/bin/crc-check-cloud-env.sh'
 ExecStartPre=/bin/sh -c 'for i in {1..10}; do ip link show "\\\$1" && exit 0; sleep 1; done; exit 1' _ %i
 ExecStart=/usr/libexec/podman/gvforwarder -preexisting -iface %i -url vsock://2:"\\\${GV_VSOCK_PORT}"/connect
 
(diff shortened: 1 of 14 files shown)


createdisk.sh
```diff
@@ -114,6 +114,7 @@ After=sys-devices-virtual-net-%i.device
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
+ExecCondition=/usr/local/bin/crc-check-cloud-env.sh
```
anjannath: we need to do the reverse, the `crc-check-cloud-env.sh` script returns success (exit code 0) when its running in the cloud, so this makes it run in the cloud and not run when running on local virtualization

this should be `ExecCondition=/bin/bash -c '! /usr/local/bin/crc-check-cloud-env.sh'`

PR #1134, opened 2025-08-04
Preload aws-cli container image

anjannath:
this is needed when running the bundle on aws in
self sufficient mode, the awscli is used to  get
the pull-secret from the AWS SSM service

having this image available removes the need  to
download and extract the binary from the relased
zip whiches increases the bundle start time

## Summary by CodeRabbit

* **Chores**
  * Preloaded the AWS CLI container image on the VM to streamline setup steps.

### DIFF
--- a/createdisk.sh
+++ b/createdisk.sh
@@ -18,6 +18,7 @@ BASE_DOMAIN=$(${JQ} -r .clusterInfo.baseDomain $INSTALL_DIR/crc-bundle-info.json
 BUNDLE_TYPE=$(${JQ} -r .type $INSTALL_DIR/crc-bundle-info.json)
 ADDITIONAL_PACKAGES="cloud-init gvisor-tap-vsock-gvforwarder"
 PRE_DOWNLOADED_ADDITIONAL_PACKAGES=""
+AWSCLI_VERSION=2.28.1
 
 case ${BUNDLE_TYPE} in
     microshift)
@@ -134,6 +135,9 @@ ${SSH} core@${VM_IP} -- "sudo podman pull quay.io/crcont/routes-controller:${ima
 TAG=${image_tag} envsubst < routes-controller.yaml.in > $INSTALL_DIR/routes-controller.yaml
 ${SCP} $INSTALL_DIR/routes-controller.yaml core@${VM_IP}:/home/core/
 ${SSH} core@${VM_IP} -- 'sudo mkdir -p /opt/crc && sudo mv /home/core/routes-controller.yaml /opt/crc/'
+
+# Preload aws-cli image
+${SSH} core@${VM_IP} -- "sudo podman pull docker.io/amazon/aws-cli:${AWSCLI_VERSION}"
 
 if [ ${BUNDLE_TYPE} != "microshift" ]; then
     # Add internalIP as node IP for kubelet systemd unit file


createdisk.sh
```diff
@@ -135,6 +136,9 @@ TAG=${image_tag} envsubst < routes-controller.yaml.in > $INSTALL_DIR/routes-cont
 ${SCP} $INSTALL_DIR/routes-controller.yaml core@${VM_IP}:/home/core/
 ${SSH} core@${VM_IP} -- 'sudo mkdir -p /opt/crc && sudo mv /home/core/routes-controller.yaml /opt/crc/'
 
+# Preload aws-cli image
+${SSH} core@${VM_IP} -- "sudo podman pull docker.io/amazon/aws-cli:${AWSCLI_VERSION}"
+
```
praveenkumar, line 141: @anjannath I tried following directly in mapt codebase and it take around 20-30sec max. Also I think we shouldn't cache cloud specific images otherwise it would be lot storage usage like for gcp, azure, openstack ..etc. Better to have cloud specific logic and images part of mapt only.

```patch
diff --git a/pkg/provider/aws/action/openshift-snc/cloud-config b/pkg/provider/aws/action/openshift-snc/cloud-config
index 7feec974..ff7a7dbf 100644
--- a/pkg/provider/aws/action/openshift-snc/cloud-config
+++ b/pkg/provider/aws/action/openshift-snc/cloud-config
@@ -6,17 +6,13 @@ bootcmd:
   - xfs_growfs /sysroot
 runcmd:
   - systemctl enable --now kubelet
-  - pushd /tmp
-  - curl "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" -o "awscliv2.zip"
-  - podman run -it --rm --workdir /data --privileged -v ${PWD}:/data registry.access.redhat.com/ubi9/toolbox unzip awscliv2.zip
-  - ./aws/install
-  - export PS=$(aws ssm get-parameter --name "{{ .SSMPullSecretName }}" --with-decryption --query "Parameter.Value" --output text)
+  - export PS=$(podman run --rm docker.io/amazon/aws-cli ssm get-parameter --name "{{ .SSMPullSecretName }}" --with-decryption --query "Parameter.Value" --output text)
   - echo ${PS} > /opt/crc/pull-secret
   - chmod 0644 /opt/crc/pull-secret
-  - export KP=$(aws ssm get-parameter --name "{{ .SSMKubeAdminPasswordName }}" --with-decryption --query "Parameter.Value" --output text)
+  - export KP=$(podman run --rm docker.io/amazon/aws-cli ssm get-parameter --name "{{ .SSMKubeAdminPasswordName }}" --with-decryption --query "Parameter.Value" --output text)
   - echo ${KP} > /opt/crc/pass_kubeadmin
   - chmod 0644 /opt/crc/pass_kubeadmin
-  - export DV=$(aws ssm get-parameter --name "{{ .SSMDeveloperPasswordName }}" --with-decryption --query "Parameter.Value" --output text)
+  - export DV=$(podman run --rm docker.io/amazon/aws-cli ssm get-parameter --name "{{ .SSMDeveloperPasswordName }}" --with-decryption --query "Parameter.Value" --output text)

```

adrianriobo: Yeah I was gonna say the same...I think it should not be snc responsible for tihs,otherwise you are opening the door to manage every provider here

anjannath: Right, thanks for taking a look, closing.. better to update to cloud-config in mapt

PR #1142, opened 2025-09-01
pki: Use 2022 root CA instead of outdated 2015 one

praveenkumar:
Since 2015 CA now not working and get following error when skopeo used with it. This PR update to 2022 which should work as expected.
```
skopeo copy --dest-authfile updated_pullsecret.json --all --src-cert-dir=pki/ docker://registry-proxy.engineering.redhat.com/rh-osbs/openshift-crc-cluster-kube-apiserver-rhel9-operator:v4.19.0-202508060837.p0.ge2ad6c1.assembly.stream.el9 docker://quay.io/crcont/openshift-crc-cluster-kube-apiserver-operator:4.19.8
time="2025-09-01T06:16:21-04:00" level=fatal msg="initializing source docker://registry-proxy.engineering.redhat.com/rh-osbs/openshift-crc-cluster-kube-apiserver-rhel9-operator:v4.19.0-202508060837.p0.ge2ad6c1.assembly.stream.el9: pinging container registry registry-proxy.engineering.redhat.com: Get \"https://registry-proxy.engineering.redhat.com/v2/\": tls: failed to verify certificate: x509: certificate signed by unknown authority"
```


cfergeau (commented):
It’s more about registry-proxy.engineering.redhat.com which apparently has switched to a cert signed with the new CA rather than about skopeo not working with the previous one.
Fwiw, we could also keep both certs in `pki/` since skopeo uses `--src-cert-dir=pki/` but I agree it’s not useful to keep an obsolete CA in this dir.

PR #1153, opened 2025-09-09
Changes to have reduce provision time and remove cloud dependency

praveenkumar:


docs/self-sufficient-bundle.md
```diff
@@ -23,7 +23,7 @@ related to their `*.service` counterparts and starts the service when the paths
 
 > [!NOTE]
 > "Marker env variable" is set using an env file, if the required env variable is not set then unit is skipped
-> some units are run only when CRC_CLOUD=1 is set, these are only needed when using the bundles with crc-cloud
+> some units are run only when CRC_NG=1 is set, these are only needed when using the bundles with crc-cloud
```
cfergeau: The name `CRC_CLOUD` feels better to me, it’s more descriptive than `CRC_NG`, and is relatively accurate.
`CRC_STANDALONE_BUNDLE`, `CRC_STANDALONE_EXECUTION`, `CRC_NO_CRC` (not really serious about this one) may work too.

docs/self-sufficient-bundle.md
```diff
@@ -12,9 +12,9 @@ services to do their work.
 |    `crc-pullsecret.service`    |               both               |         /opt/crc/pull-secret         |         none         |
 |      `crc-dnsmasq.service`     |               both               |                 none                 |         none         |
 | `crc-routes-controller.service`|               both               |                 none                 |         none         |
-|    `ocp-cluster-ca.service`    |               ocp                |        /opt/crc/custom-ca.crt        |     CRC_NG=1         |
+|    `ocp-cluster-ca.service`    |               ocp                |        /opt/crc/custom-ca.crt        |         none         |
```
cfergeau, line 15: Not sure why this is changed in this commit? Does this depend on ocp-custom-domain.service?

systemd/crc-cluster-status.sh
```diff
@@ -4,12 +4,33 @@ set -x
...
 
 if ! oc adm wait-for-stable-cluster --minimum-stable-period=1m --timeout=10m; then
     exit 1
 fi
 
+set +x
+echo "Logging into OpenShift with kubeadmin user to update $KUBECONFIG"
+COUNTER=1
+MAXIMUM_LOGIN_RETRY=10
+until oc login --insecure-skip-tls-verify=true -u kubeadmin -p "$PASS_KUBEADMIN" https://api.crc.testing:6443 > /dev/null 2>&1; do
```
cfergeau, line 24: Not directly related to this change, but since this runs on the cluster, this most likely could use 
```
    --certificate-authority='':
	Path to a cert file for the certificate authority
```
instead of skipping TLS verification altogether.

createdisk.sh
```diff
@@ -114,7 +114,7 @@ After=sys-devices-virtual-net-%i.device
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
-ExecCondition=/bin/sh -c '! /usr/local/bin/crc-check-cloud-env.sh'
+ExecCondition=/bin/sh -c '! /usr/local/bin/crc-ng-env.sh'
```
cfergeau: This should be removed/moved to `crc-no-tap-on-apple.service`

systemd/crc-cluster-status.sh
cfergeau, line 24: This is not only a move, this fixes a few things suggested by coderabbit. I’d favour having 2 commits, one with only a move, and one with the changes so that they are easier to review

docs/self-sufficient-bundle.md
praveenkumar: `CRC_CLOUD` feels like this variable depend on running the crc on cloud env which is not the case so to avoid confusion I thought having `NG` => next gen would be much better. @crc-org/crc-team wdyt?

createdisk.sh
praveenkumar: yes, I think it will need bit more work and with this change now running bundle with crc doesn't make gvforwarder to stop in cloud env.

docs/self-sufficient-bundle.md
praveenkumar: looks like it was by mistake presented and `ocp-custom-domain` already had this check about cloud-ng env.

systemd/crc-cluster-status.sh
praveenkumar: @cfergeau may be some follow up PR but I think in case of mapt this file is copied to user host and there no valid cert auth file present so having `insecure-skip-tls-verify=true` make is easier to test and consume the cluster even in CI use case.

systemd/crc-ng-env.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+if [ -n "$CRC_NG" ] || [ -n "$CRC_CLOUD" ]; then
+    exit 1
```
anjannath: shouldn't this be `exit 0`? in case there's a value set for `CRC_NG` or `CRC_CLOUD` we want to `exit 0` (so that self sufficient services run) and in case this env variables are not set `exit 1`

systemd/crc-no-tap.service
```diff
@@ -7,7 +7,8 @@ RequiresMountsFor=/etc/NetworkManager/system-connections
 
 [Service]
 Type=oneshot
-ExecStart=/usr/bin/rm -f /etc/NetworkManager/system-connections/tap0.nmconnection
+EnvironmentFile=-/etc/sysconfig/crc-env
+ExecStart=/usr/local/bin/crc-no-tap.sh
```
anjannath, line 10: maybe we should have two versions of this service, one with 

`ConditionVirtualization=apple` and the other with `ExecCondition=/usr/local/bin/check-crc-ng-env.sh`

the first removes this file in case we are running on macOS, and the other service will remove it when `CRC_CLOUD` or `CRC_NG` is set

in any condition one of them will be skipped, unless we run the self sufficient bundle on a mac machine in aws, where both of these services will start, ~so we might also ignore the error from `Exec` in the services~ `rm -f` is used!

cfergeau, line 10: `crc-no-tap.sh` is empty?

cfergeau: Yup, sounds good to me, this way we keep as much logic as possible in the unit file instead of having to go look at another script.

Unrelated thought, but I wonder if instead of running `crc-ng-env.sh` again and again in an `ExecCondition`, we could have a single unit file doing it, and then have the other units "depend" on that unit to only be started if the `crc-ng-env.sh` succeeded (or did not succeed). Just food for thoughts, no need to address it in this PR.

praveenkumar: > maybe we should have two versions of this service, one with

> ConditionVirtualization=apple and the other with ExecCondition=/usr/local/bin/check-crc-ng-env.sh

> the first removes this file in case we are running on macOS, and the other service will remove it when CRC_CLOUD or CRC_NG is set

if we have crc-no-tap logic where it remove the tap device any of the condition run then why we want to have 2 versions of it? Does it fails for some condition?

praveenkumar: Ah looks like it was not saved and just added, I need to configure the editor for autosave.

praveenkumar: > Unrelated thought, but I wonder if instead of running crc-ng-env.sh again and again in an ExecCondition, we could have a single unit file doing it, and then have the other units "depend" on that unit to only be started if the crc-ng-env.sh succeeded (or did not succeed). Just food for thoughts, no need to address it in this PR.

yeah, something can be explored.

@coderabbitai create follow up issue

systemd/crc-no-tap.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Return true if running under Apple Virtualization or CRC_NG is set, otherwise false
+
+if systemd-detect-virt | grep -q '^apple$' || [ -n "$CRC_NG" ]; then
+    rm -f /etc/NetworkManager/system-connections/tap0.nmconnection
+fi
```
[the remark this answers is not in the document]
praveenkumar: @coderabbitai review

systemd/crc-no-tap.service
cfergeau: > if we have crc-no-tap logic where it remove the tap device any of the condition run then why we want to have 2 versions of it? Does it fails for some condition?

I prefer to have as much as possible in systemd units, and as little as possible in external shell scripts, this way everything is in one place.

docs/self-sufficient-bundle.md
cfergeau: The doc is still saying "some units are run only when CRC_NG=1 is set, these are only needed when using the bundles with crc-cloud", so `CRC_CLOUD` makes some sense.

What does `CRC_NG` mean in this context? We can’t rewrite the sentence above with "These are only needed when using the next generation" or something like this, can we? Ie I’d describe with words what it means to set the variable, and from this description, maybe we can come up with a good name.

cfergeau: > `CRC_CLOUD` feels like this variable depend on running the crc on cloud env which is not the case

It’s a good approximation though

createdisk.sh
```diff
@@ -109,6 +109,7 @@ ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
 [Unit]
 Description=gvisor-tap-vsock Network Traffic Forwarder
 After=sys-devices-virtual-net-%i.device
+ConditionVirtualization=!apple
```
cfergeau: Is it possible to ensure the `tap0` device does not exist at all on apple?
I’d prefer to have `crc-no-tap` run `systemctl disable gv-user-network@tap0.service` rather than adding this condition here.

createdisk.sh
```diff
@@ -109,12 +109,13 @@ ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
...
 
 [Service]
 Restart=on-failure
 Environment="GV_VSOCK_PORT=1024"
 EnvironmentFile=-/etc/sysconfig/gv-user-network
-ExecCondition=/bin/sh -c '! /usr/local/bin/crc-check-cloud-env.sh'
+ExecCondition=/bin/sh -c '! /usr/local/bin/crc-ng-env.sh'
```
anjannath: Irrespective of where the bundle runs, we want this service to run when the VM is configured to use network-mode `user`, so instead of coupling this check with the cloud-env checking script, i think it'd be simpler to just create another check script that checks for the `CRC_NETWORK_MODE_USER` value

if `CRC_NETWORK_MODE_USER` is unset or set to `1` we want this service to run

```bash
#!/bin/bash

if [ -z "${CRC_NETWORK_MODE_USER}" ]; then
    exit 0
fi

if [ "${CRC_NETWORK_MODE_USER}" == "1" ]; then
    exit 0
fi
exit 1
```

similarly the existing `crc-ng-env.sh` script does:

```bash
#!/bin/bash

if [ -z  "${CRC_CLOUD}" ] || [ -z "${CRC_NG}" ]; then
    exit 1
fi

if [ "${CRC_CLOUD}" == "1" ] || [ "${CRC_NG}" == "1" ]; then
    exit 0
fi
exit 1
```

then we'll use this scripts in the units for `ExecCondition` without any negation and keep the checks simple

### with this when running the bundle locally with crc:

1. the systemd services for self-sufficient will not run (since `CRC_CLOUD` is not set)
2. the existing behavior of `crc` will be the same (i.e using user network mode without any cloud-init config)
3. when running on cloud, using cloud-init we'll be able to enable/disable the user network-mode using `/etc/sysconfig/crc-env` file

createdisk.sh
anjannath: > I’d prefer to have crc-no-tap run systemctl disable gv-user-network@tap0.service rather than adding this condition here.

this way we have the `ConditionVirtualization` check only on the `crc-no-tap` service and would be easier to understand the condition checks for this `gv-user-network` services

praveenkumar: Done.

systemd/crc-no-tap.service
praveenkumar: > I prefer to have as much as possible in systemd units, and as little as possible in external shell scripts, this way everything is in one place.

Can you check again, I think now that external script not only delete tap configuration but also disable the forwarder service, do you still prefer to have 2 separate service doing same thing depending on the environment and virtualization?

systemd/crc-cluster-status.sh
praveenkumar: Done

docs/self-sufficient-bundle.md
praveenkumar: Updated the readme about self sufficient bundle and add context for CRC_NG reference, hope that will make it bit easier to consume.

createdisk.sh
cfergeau: > Irrespective of where the bundle runs, we want this service to run when the VM is configured to use network-mode user

We want the service if and only if the corresponding `tap0` device exists, I’d prefer we don’t add more conditions to this service, and focus on getting the `tap0` device existence right.

There are 2 variants of usermode networking:
- over vsock, where you need `gvforwarder` and a `tap` device
- over virtio-net, which needs hypervisor support (to redirect the traffic from the virtio-net device to a unix socket), but does not need anything in the guest

Both would qualify as "usermode networking", and both requires the routes-controller in the cluster. I think this will complicate things if we add more conditions in this gvforwarder unit.

docs/self-sufficient-bundle.md
```diff
@@ -12,18 +12,18 @@ services to do their work.
...
 |   `ocp-userpasswords.service`  |               ocp                | /opt/crc/pass_{kubeadmin, developer} |         none         |
 
 In addition to the above services we have `ocp-cluster-ca.path`, `crc-pullsecret.path` and `ocp-userpasswords.path` that monitors the filesystem paths
 related to their `*.service` counterparts and starts the service when the paths become available.
 
 > [!NOTE]
-> "Marker env variable" is set using an env file, if the required env variable is not set then unit is skipped
-> some units are run only when CRC_CLOUD=1 is set, these are only needed when using the bundles with crc-cloud
+> "Marker env variable" is set using an env file (/etc/sysconfig/crc.env), if the required env variable is not set then unit is skipped
+> some units are run only when CRC_NG=1 is set, these are only needed when using self-sufficient functionality which is part of CRC next generation (CRC_NG) initative.
```
cfergeau: But why use a cryptic name for this? `CRC_SELF_SUFFICIENT` would be a lot better than `NG`.

praveenkumar: Done, changed with `CRC_SELF_SUFFICIENT`

anjannath:
latest run using the bundles from this PR (4.18.12) before the latest push, shows substantial improvements to cluster start time (up in ~12mins), with `mapt` on aws spot instance:

```
INFO @ updating.......
INFO  +  command:remote:Command main-ocp-ca-rotated-aos-cmd creating (4s) NAME   STATUS   ROLES                         AGE   VERSION
INFO  +  command:remote:Command main-ocp-ca-rotated-aos-cmd creating (4s) crc    Ready    control-plane,master,worker   46h   v1.32.8
INFO  +  command:remote:Command main-ocp-ca-rotated-aos-cmd created (4s) crc    Ready    control-plane,master,worker   46h   v1.32.8
INFO  +  command:remote:Command main-kubeconfig-aos-cmd creating (0s)
INFO @ updating.......
INFO  +  command:remote:Command main-kubeconfig-aos-cmd created (3s)
INFO  +  pulumi:pulumi:Stack mapt-snc-stackOpenshiftSNC-mapt-snc created (661s)
INFO Outputs:
INFO     VPCID                 : "vpc-0c34995a62dace03f"
INFO     main-aosDeveloperPass : [secret]
INFO     main-aosHost          : "main-aos-lb-82e5c9b-9af31ae5c8692cdc.elb.us-west-1.amazonaws.com"
INFO     main-aosKubeAdminPasss: [secret]
INFO     main-aosKubeconfig    : [secret]
INFO     main-aosPrivatekey    : [secret]
INFO     main-aosUsername      : "core"
INFO
INFO Resources:
INFO     + 31 created
INFO
INFO Duration: 11m2s
INFO
```

systemd/ocp-custom-domain.sh
```diff
@@ -2,8 +2,8 @@
 
 set -x
 
-if [ -z $CRC_CLOUD ]; then
-    echo "Not running in crc-cloud mode"
+if [ -z $CRC_SELF_SUFFICIENT ]; then
```
anjannath: i think this came back after rebasing again, we can remove this check its already part of the service `ExecCondition`?

praveenkumar: I thought it is removed in next commit but looks like during rebase it is not, will do that.

praveenkumar: Done.

PR #1155, opened 2025-09-17
qemu-guest-agent.service: fix deprecated --blacklist flag

kpouget:
After this patch, `qemu-ga` will correctly use RHOCS the default value from RHCOS `/etc/sysconfig/qemu-ga`:

```
# Guest agent command with comma-separated blocked RPCs to disable,
# or empty list to enable all.
#
# You can get the list of RPC commands using "qemu-ga --block-rpcs='?'".
# There should be no spaces between commas and commands in the block list.
# FILTER_RPC_ARGS="--block-rpcs=guest-file-open,guest-file-close,guest-file-read,guest-file-write,guest-file-seek,guest-file-flush,guest-exec,guest-exec-status"

# Guest agent command with comma-separated allowed RPCs to enable,
# or empty list to disable all.
#
# You can get the list of RPC commands using "qemu-ga --allow-rpcs='?'".
# There should be no spaces between commas and commands in the allow list.
FILTER_RPC_ARGS="--allow-rpcs=guest-sync-delimited,guest-sync,guest-ping,guest-get-time,guest-set-time,guest-info,guest-shutdown,guest-fsfreeze-status,guest-fsfreeze-freeze,guest-fsfreeze-freeze-list,guest-fsfreeze-thaw,guest-fstrim,guest-suspend-disk,guest-suspend-ram,guest-suspend-hybrid,guest-network-get-interfaces,guest-get-vcpus,guest-set-vcpus,guest-get-disks,guest-get-fsinfo,guest-set-user-password,guest-get-memory-blocks,guest-set-memory-blocks,guest-get-memory-block-info,guest-get-host-name,guest-get-users,guest-get-timezone,guest-get-osinfo,guest-get-devices,guest-ssh-get-authorized-keys,guest-ssh-add-authorized-keys,guest-ssh-remove-authorized-keys,guest-get-diskstats,guest-get-cpustats"
```


kpouget:
@cfergeau WDYT?

qemu-guest-agent.service
```diff
@@ -9,7 +9,7 @@ EnvironmentFile=/etc/sysconfig/qemu-ga
 ExecStart=/usr/bin/qemu-ga \
   --method=vsock-listen \
   --path=3:1234 \
-  --blacklist=${BLACKLIST_RPC} \
+  $FILTER_RPC_ARGS \
```
cfergeau, line 12: The RHEL systemd unit uses `${FILTER_RPC_ARGS}`, otherwise fine with me.

qemu-guest-agent.service
```diff
@@ -9,7 +9,7 @@ EnvironmentFile=/etc/sysconfig/qemu-ga
 ExecStart=/usr/bin/qemu-ga \
   --method=vsock-listen \
   --path=3:1234 \
-  --blacklist=${BLACKLIST_RPC} \
+  ${FILTER_RPC_ARGS} \
```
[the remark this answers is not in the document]
kpouget: @cfergeau, I followed this suggestion which looks [legit](https://www.man7.org/linux/man-pages/man5/systemd.service.5.html)
> Use
>       "$FOO" as a separate word on the command line, in which case it
>       will be replaced by the value of the environment variable split at
>       whitespace, resulting in zero or more arguments.

qemu-guest-agent.service
kpouget: see the outdated comment where the bot  suggested that `$FILTER_RPC_ARGS` is better than `${...}`

cfergeau: Would be worth filing an issue against the RHEL package instead of silently diverging from it here.

kpouget: https://issues.redhat.com/browse/RHEL-116179

PR #1189, opened 2025-11-04
ssf: add script to start bundle with macadam

anjannath:
with the self sufficient changes for the bundle
have to make sure it runs well with CRC as well
as via the cloud-init path with an user-data

for crc we have a test in openshift-ci that run
the bundle with CRC once it is generated

this adds a shell script, which we'll run on CI
for every PR to catch any regressions

Assisted-By: Gemini

Needs: https://github.com/cfergeau/podman/pull/37

## Summary by CodeRabbit

* **Chores**
  * Replaced OpenShift CRC-based CI test setup with macadam-based smoke test workflow for improved testing infrastructure
  * Updated microshift testing configuration
  * Updated default release candidate mirror URL and version retrieval endpoint

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -48,6 +48,10 @@ make containerized
 sudo mv out/linux-amd64/crc /usr/local/bin/
 popd
 
+export CRC_BUNDLE_PATH=$(pwd)/$(ls crc_libvirt_*.crcbundle)
+export PULL_SECRET_PATH="${HOME}/pull-secret"
+./run-with-macadam.py
+
 crc config set bundle crc_libvirt_*.crcbundle
 crc setup
 crc start --disk-size 80 -m 24000 -c 10 -p "${HOME}"/pull-secret --log-level debug
--- a/run-with-macadam.py
+++ b/run-with-macadam.py
@@ -1,0 +1,624 @@
+#!/usr/bin/env python3
+"""CRC Bundle Testing with Macadam Hypervisor - Port of run-with-macadam.sh"""
+
+import http.client
+import json
+import os
+import re
+import shutil
+import socket
+import subprocess
+import sys
+import time
+import urllib.request
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+@dataclass
+class Config:
+    pass_developer: str
+    pass_kubeadmin: str
+    crc_bundle_path: Path
+    pull_secret_path: Path
+    vm_name: str = "crc-ng"
+
+state: Dict[str, Any] = {
+    'priv_key_path': None,
+    'pub_key_path': None,
+    'user_data_path': None,
+    'disk_image_path': None,
+    'gvproxy_socket_path': None,
+    'oc_binary_path': None,
+    'kubeconfig_path': None,
+    'pull_secret': '',
+    'pub_key': '',
+}
+
+class MacadamError(Exception):
+    pass
+
+class DependencyError(MacadamError):
+    pass
+
+class VMError(MacadamError):
+    pass
+
+def run_command(
+    cmd: List[str],
+    *,
+    capture_output: bool = False,
+    check: bool = True,
+    stdout: Any = None,
+    stderr: Any = None,
+    env: Optional[Dict[str, str]] = None,
+) -> subprocess.CompletedProcess:
+    try:
+        return subprocess.run(
+            cmd,
+            capture_output=capture_output,
+            check=check,
+            stdout=stdout,
+            stderr=stderr,
+            env=env,
+            text=True,
+        )
+    except subprocess.CalledProcessError as e:
+        raise MacadamError(f"Command failed: {' '.join(cmd)}") from e
+    except FileNotFoundError as e:
+        raise MacadamError(f"Command not found: {cmd[0]}") from e
+
+def load_config_from_env() -> Config:
+    pull_secret_path_str = os.environ.get("PULL_SECRET_PATH", "")
+    if not pull_secret_path_str:
+        raise MacadamError("PULL_SECRET_PATH environment variable must be set.")
+
+    pull_secret_path = Path(pull_secret_path_str)
+    if not pull_secret_path.exists():
+        raise MacadamError(f"Pull secret file not found at {pull_secret_path}.")
+
+    crc_bundle_path_str = os.environ.get(
+        "CRC_BUNDLE_PATH",
+        str(Path.home() / "Downloads" / "crc_vfkit_4.19.3_arm64.crcbundle")
+    )
+    crc_bundle_path = Path(crc_bundle_path_str)
+
+    return Config(
+        pass_developer=os.environ.get("PASS_DEVELOPER", "P@ssd3v3loper"),
+        pass_kubeadmin=os.environ.get("PASS_KUBEADMIN", "P@sskub3admin"),
+        crc_bundle_path=crc_bundle_path,
+        pull_secret_path=pull_secret_path,
+    )
+
+def ensure_dependencies() -> None:
+    print("--- Checking dependencies ---")
+
+    # Basic utilities
+    basic_deps = ["zstd", "tar", "curl"]
+
+    for dep in basic_deps:
+        if shutil.which(dep) is None:
+            print(f"'{dep}' not found, attempting to install...")
+            try:
+                run_command(["sudo", "yum", "-y", "install", dep])
+            except MacadamError as e:
+                raise DependencyError(
+                    f"Failed to install '{dep}'. "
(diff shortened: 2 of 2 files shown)


run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+    echo "Temporary SSH keypair generated."
+}
+
+# Ensures all required dependencies are available in the PATH
+function ensure_deps() {
+    echo "--- Checking dependencies ---"
+    local deps=("zstd" "tar" "curl" "ssh" "scp" "oc")
+    for dep in "${deps[@]}"; do
+        if ! command -v "${dep}" &>/dev/null; then
+            die "${dep} is not installed. Please install it and try again."
```
praveenkumar: instead of die shouldn't we install it because in case of CI run this will die?

run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+- path: /etc/sysconfig/crc-env
+  content: |
+    CRC_SELF_SUFFICIENT=1
+    CRC_NETWORK_MODE_USER=1
+  owner: root:root
+  permissions: '0644'
+- path: /usr/local/bin/crc-check-cloud-env.sh
+  content: |
+    #!/bin/bash
+    exit 0
```
praveenkumar: Do we still need this?

run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+            "Name": "api",
+            "IP": "192.168.127.2"
+        },
+        {
+            "Name": "api-int",
+            "IP": "192.168.127.2"
+        },
+        {
+            "Regexp": "crc-(.*?)-master-0",
+            "IP": "192.168.126.11"
```
praveenkumar: This is also not required, we now have fix hostname `crc` so need to updated to it.

run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+    ]
+}'
+    local APPS_CRC_TESTING_ZONE='{"Name": "apps-crc.testing.","DefaultIP": "192.168.127.2"}'
+    local CONTAINERS_INTERNAL_ZONE='{
+    "Name": "containers.internal.",
+    "Records": [
+        {
+            "Name": "gateway",
+            "IP": "192.168.127.254"
+        }
```
praveenkumar: I think this was for podman to work but since now podman-machine is de-facto for podman related stuff so we might need to remove it?

run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+    # for this flag contains the socket path
+    local gvproxy_cmd
+    if ! gvproxy_cmd=$(pgrep -af gvproxy); then
+        die "The gvproxy process is not running."
+    fi
+
+    GVPROXY_SOCKET_PATH=$(echo "${gvproxy_cmd}" | sed -n 's/.*-services unix:\/\/\([^ ]*\).*/\1/p')
+    if [[ -z "${GVPROXY_SOCKET_PATH}" ]]; then
+        die "Could not find the gvproxy socket path."
+    fi
```
cfergeau: Maybe this is / should be exposed in `macadam inspect`

run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+    if [[ ! -f "${CRC_BUNDLE_PATH}" ]]; then
+        die "CRC bundle not found at ${CRC_BUNDLE_PATH}."
+    fi
+
+    zstd -d --format=zstd -o bundle.tar "${CRC_BUNDLE_PATH}"
+    local bundle_name
+    bundle_name=$(basename "${CRC_BUNDLE_PATH}")
+    local disk_image_name
+    disk_image_name="crc.qcow2"
+    tar -O -xvf bundle.tar "${bundle_name%.*}"/${disk_image_name} >${disk_image_name}
```
cfergeau: maybe you can do `zstd` + `tar` in one step with `tar -xvf --zstd …`

run-with-macadam.sh
```diff
@@ -0,0 +1,348 @@
...
+    $PULL_SECRET
+  permissions: '0644'
+- path: /opt/crc/pass_kubeadmin
+  content: '$PASS_KUBEADMIN'
+  permissions: '0644'
+- path: /opt/crc/pass_developer
+  content: '$PASS_DEVELOPER'
+  permissions: '0644'
+- path: /opt/crc/ocp-custom-domain.service.done
+  permissions: '0644'
```
cfergeau: This is not related to this PR, but hopefully handling of custom domains can be improved so that this is not needed.

cfergeau: Actually `ocp-custom-domain.service` has `ExecCondition=/usr/local/bin/crc-self-sufficient-env.sh`, will it be run with macadam?

run-with-macadam.sh
anjannath: no, its not needed now, removed, thanks

run-with-macadam.sh
anjannath: done, added `Name: crc` record instead of the `Regexp` record

run-with-macadam.sh
anjannath: removed

run-with-macadam.sh
anjannath: While running with `macadam` with a cloud-init `user-data` it is running in self-sufficient mode (i.e `CRC_SEFL_SUFFICIENT=1` and systemd services doing the setup) 

so this service for domain name change will also run, I think we could add another config variable in the `/etc/sysconfig/crc-env` to enable/disable it and provide a proper way to control it, for now this is a hack to skip the service

run-with-macadam.sh
```diff
@@ -0,0 +1,327 @@
...
+    if [[ ! -f "${PUB_KEY_PATH}" ]]; then
+        die "Public key file not found at ${PUB_KEY_PATH}. This should have been generated."
+    fi
+
+    PULL_SECRET=$(cat "${PULL_SECRET_PATH}")
+    PUB_KEY=$(cat "${PUB_KEY_PATH}")
+}
+
+# Generates cloud-init user-data file for VM configuration
+function gen_cloud_init() {
```
anjannath: run `cloud-init schema -c user-data --annotate` to validate the generated cloud-init user data file

run-with-macadam.sh
```diff
@@ -0,0 +1,327 @@
...
+        --cloud-init "${USER_DATA_PATH}" \
+        --log-level debug
+
+    if ! macadam start crc-ng --log-level debug; then
+        echo "Machine didn't come up in time"
+    fi
+}
+
+# Exposes a port from the VM to the host
+function forward_port_gvproxy() {
```
anjannath: - need to ensure gvproxy is available on the machine
- run `sudo setcap 'cap_net_bind_service=+ep' gvproxy`
- also forward ports 80 and 443

praveenkumar:
```bash
# Ensures gvproxy binary has the required capability
function ensure_gvproxy_capability() {
    echo "--- Checking gvproxy capability ---"
    local gvproxy_path=""

    # Check for gvproxy in known locations
    if [ -f "/usr/bin/gvproxy" ]; then
        gvproxy_path="/usr/bin/gvproxy"
    elif [ -f "/usr/libexec/podman/gvproxy" ]; then
        gvproxy_path="/usr/libexec/podman/gvproxy"
    else
        die "gvproxy binary not found in /usr/bin or /usr/libexec/podman"
    fi

    echo "Found gvproxy at: ${gvproxy_path}"

    # Check if the capability is already set
    local current_caps
    current_caps=$(getcap "${gvproxy_path}" 2>/dev/null || true)

    if echo "${current_caps}" | grep -q "cap_net_bind_service=.*ep"; then
        echo "gvproxy already has cap_net_bind_service=+ep capability set."
        return
    fi

    echo "Setting cap_net_bind_service=+ep capability on gvproxy..."
    if ! sudo setcap cap_net_bind_service=+ep "${gvproxy_path}"; then
        die "Failed to set capability on gvproxy. Please ensure you have sudo privileges."
    fi

    # Verify the capability was set
    current_caps=$(getcap "${gvproxy_path}" 2>/dev/null || true)
    if echo "${current_caps}" | grep -q "cap_net_bind_service=.*ep"; then
        echo "Capability successfully set on gvproxy."
    else
        die "Failed to verify capability was set on gvproxy."
    fi
}
```

This should check if gvproxy present and required capability set to it.

cfergeau:
If this is running on RHEL, this is expected, we may have forgotten to handle this in macadam.
Create a symlink from `/usr/libexec/qemu-kvm` to `qemu-system-x86_64`

PR #1216, opened 2026-04-02
Install qemu-user-static from newer fedora release

cfergeau:
This PR gets qemu-user-static from fedora 44 instead of using the out of support f41.
Maybe we should consider getting qemu-user-static from centos if it’s built there.

This PR also attempts to enable gpgcheck on yum repos, but I expect it will take a few tries to get it to work (if at all).

- **Install qemu-user-static from fedora 44**
- **Enable gpgcheck on yum repos**

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced package signature verification by enabling GPG checks across repository configurations

* **Chores**
  * Updated Fedora repository targeting from version 41 to version 44

### DIFF
--- a/repos/mirror-microshift.repo
+++ b/repos/mirror-microshift.repo
@@ -2,4 +2,4 @@
 name=microshift repo for mirror
 baseurl=https://mirror.openshift.com/pub/openshift-v4/$basearch/microshift/ocp-dev-preview/latest-4.20/el9/os/
 enabled=1
-gpgcheck=0
+gpgcheck=1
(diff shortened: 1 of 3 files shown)


repos/mirror-microshift.repo
```diff
@@ -2,4 +2,4 @@
 name=microshift repo for mirror
 baseurl=https://mirror.openshift.com/pub/openshift-v4/$basearch/microshift/ocp-dev-preview/latest-4.20/el9/os/
 enabled=1
-gpgcheck=0
+gpgcheck=1
```
praveenkumar, line 5: This might break for mirror repos and it is not tested becasue this repo is only used for non release microshift payload so might be for 4.22 release side.

cfergeau:
This has been superseded by https://github.com/crc-org/snc/pull/1223

praveenkumar:
since #1223 merged, closing this one.

PR #1226, opened 2026-04-29
Add libvirt service check helper

praveenkumar:
We are planning to switch the GCP image in CI from our custom one to publicly available on gcp marketplace. This PR is a result to missing pieces to achieve it.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -2,7 +2,7 @@
 
 set -exuo pipefail
 
-sudo yum install -y podman make golang rsync
+sudo yum install -y git-core podman make golang rsync
 
 ./shellcheck.sh
 ./snc.sh
--- a/ci.sh
+++ b/ci.sh
@@ -2,7 +2,7 @@
 
 set -exuo pipefail
 
-sudo yum install -y podman make golang rsync
+sudo yum install -y git-core virtiofsd podman make golang rsync
 
 ./shellcheck.sh
 ./snc.sh
--- a/tools.sh
+++ b/tools.sh
@@ -107,6 +107,16 @@ if ! which ${PATCH}; then
     sudo yum -y install /usr/bin/patch
 fi
 
+function check_libvirt_service {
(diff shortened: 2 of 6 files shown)


ci.sh
```diff
@@ -2,7 +2,7 @@
 
 set -exuo pipefail
 
-sudo yum install -y podman make golang rsync
+sudo yum install -y git podman make golang rsync
```
cfergeau: `git-core` should be enough

tools.sh
```diff
@@ -107,6 +107,22 @@ if ! which ${PATCH}; then
...
+        return
+    fi
+
+    for service in virtqemud virtstoraged virtnetworkd; do
+        if ! sudo systemctl is-active --quiet ${service}; then
+            echo "${service} is not running, starting it..."
+            sudo systemctl start ${service}
+        else
+            echo "${service} is already running"
+        fi
```
cfergeau: There is nothing redhat/fedora/… specific in there, I’m not sure the check you added is really needed.
Might be worth using the command from RHEL documentation instead of doing our own thing:
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/configuring_and_managing_virtualization/assembly_enabling-virtualization-in-rhel-9_configuring-and-managing-virtualization#proc_enabling-virtualization-in-rhel-9_assembly_enabling-virtualization-in-rhel-9
```
for drv in qemu network nodedev nwfilter secret storage interface; do systemctl start virt${drv}d{,-ro,-admin}.socket; done
```

Should this also try to install the packages if they are missing?

praveenkumar: > Should this also try to install the packages if they are missing?

Packages are installed as dependency of virt-filesystem  if missing when installing which is part of tools.sh script.

praveenkumar:
created https://github.com/openshift/release/pull/78606 one for ci side but either this should go first or CI one to make it green.

anjannath (approved):

PR #1228, opened 2026-05-14
snc: Replace custom OS image with transient-ro for /Users directory 

praveenkumar:
Replace the custom RHCOS container image approach for creating the
/Users top-level directory with ostree's transient-ro feature.

The previous approach built a custom container image with a /Users
symlink, pushed it to the in-cluster registry, and applied a
MachineConfig with osImageURL to deploy it. This was ~40 lines of
complex logic involving Containerfile creation, podman build/push,
kubeadmin login for registry access, and imagestream cleanup.

The new approach uses a MachineConfig with two systemd services:

1. enable-transient-ro.service (oneshot, runs once):
   - Uses rpm-ostree usroverlay to temporarily make /usr writable
   - Sets composefs.enabled from "maybe" to "yes" in prepare-root.conf
   - Appends [root] transient-ro = true to prepare-root.conf
   - Regenerates the bootloader initramfs with the updated config
   - Creates /var/Users as the symlink target

2. create-users-mount.service (runs every boot):
   - Creates /var/Users directory
   - Uses unshare -m to create /Users -> var/Users symlink on the
     transient-ro overlayfs root

The MachineConfig is included in the install manifests directory so
it is applied during initial cluster bootstrap, avoiding an extra
MCP reboot cycle.

Doc: https://bootc.dev/bootc/filesystem.html
Ref: https://github.com/crc-org/snc/issues/1041

## Summary by CodeRabbit

* **New Features**
  * Enable transient read-only workflow for /Users on master nodes, allowing /Users to be managed as a transient overlay and become writable when configured.

* **Chores**
  * Simplified provisioning to apply the transient-ro configuration directly during setup instead of building/pushing a custom image.
  * Updated provisioning verification to test writable /Users behavior; removed custom-image build/cleanup steps.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crc-org/snc/pull/1228?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/99-enable-transient-ro.yaml
+++ b/99-enable-transient-ro.yaml
@@ -1,0 +1,47 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+  name: 99-enable-transient-ro
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    systemd:
+      units:
+      - name: enable-transient-ro.service
+        enabled: true
+        contents: |
+          [Unit]
+          Description=Enable transient-ro in ostree prepare-root config
+          ConditionPathExists=!/var/lib/transient-ro-configured
+
+          [Service]
+          Type=oneshot
+          RemainAfterExit=yes
+          ExecStart=/bin/rpm-ostree usroverlay
+          ExecStart=/bin/bash -c 'sed -i "s/enabled = maybe/enabled = yes/" /usr/lib/ostree/prepare-root.conf'
+          ExecStart=/bin/bash -c 'echo -e "\\n[root]\\ntransient-ro = true" >> /usr/lib/ostree/prepare-root.conf'
+          ExecStart=/bin/mount -o remount,rw /boot
+          ExecStart=/bin/bash -c 'dracut -f $(grep -h initrd /boot/loader/entries/*.conf | awk "{print \\$2}") $(uname -r)'
+          ExecStart=/bin/mkdir -p /var/Users
+          ExecStartPost=/bin/touch /var/lib/transient-ro-configured
+
+          [Install]
+          WantedBy=multi-user.target
+      - name: create-users-mount.service
+        enabled: true
+        contents: |
+          [Unit]
+          Description=Create /Users mountpoint using transient-ro
+          After=local-fs-pre.target
+
+          [Service]
+          Type=oneshot
+          RemainAfterExit=yes
+          ExecStart=/bin/mkdir -p /var/Users
+          ExecStart=-/bin/bash -c 'unshare -m -- /bin/sh -c "mount -o remount,rw / && ln -sf var/Users /Users"'
+
+          [Install]
+          WantedBy=local-fs.target
--- a/snc.sh
+++ b/snc.sh
@@ -342,4 +342,4 @@ ${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- 'sudo mkdir /Users/foo && 
 
 # Remove the image stream of custom image
 retry ${OC} delete imagestream rhcos -n openshift-machine-config-operator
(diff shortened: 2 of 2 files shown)


99-enable-transient-ro.yaml
```diff
@@ -0,0 +1,47 @@
...
+        contents: |
+          [Unit]
+          Description=Create /Users mountpoint using transient-ro
+          After=local-fs-pre.target
+
+          [Service]
+          Type=oneshot
+          RemainAfterExit=yes
+          ExecStart=/bin/mkdir -p /var/Users
+          ExecStart=-/bin/bash -c 'unshare -m -- /bin/sh -c "mount -o remount,rw / && ln -sf var/Users /Users"'
```
anjannath, line 44: we don't have to remount with `ro` again after creating the symlink?

snc.sh
```diff
@@ -280,14 +280,7 @@ FROM scratch
...
 podman build --from ${RHCOS_IMAGE} --authfile ${OPENSHIFT_PULL_SECRET_PATH} -t default-route-openshift-image-registry.apps-crc.testing/openshift-machine-config-operator/rhcos:latest --file ${INSTALL_DIR}/Containerfile .
-(
-    set +x # disable the logging in the subshell to prevent the password leakage
-    kubeadmin_pass=$(cat ${INSTALL_DIR}/auth/kubeadmin-password)
-    retry ${OC} login -u kubeadmin -p "$kubeadmin_pass" --insecure-skip-tls-verify=true api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}:6443
-    rm -f ${INSTALL_DIR}/auth/kubeadmin-password
-    esc_pw="$(printf '%s' "$kubeadmin_pass" | sed -e 's/[\/&|\\]/\\&/g')"
-    sed -i "s|$esc_pw|REDACTED|g" "${INSTALL_DIR}/.openshift_install.log"
-)
+retry ${OC} login -u kubeadmin -p $(cat ${INSTALL_DIR}/auth/kubeadmin-password) --insecure-skip-tls-verify=true api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}:6443
```
anjannath: is this revert intentional.. ?

99-enable-transient-ro.yaml
praveenkumar: No this is required every single time https://bootc.dev/bootc/filesystem.html?highlight=prepare-root.conf#dynamic-mountpoints-with-transient-ro check https://bootc.dev/bootc/filesystem.html?highlight=prepare-root.conf#how-it-works section.

snc.sh
```diff
@@ -340,13 +299,5 @@ ${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- 'sudo crictl rmi --prune'
 # Remove the baremetal_runtimecfg container which is temp created
 ${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- "sudo podman rm baremetal_runtimecfg"
 
-# Create the /var/Users directory so it can become writeable
-# todo: remove it once custom image able to perform it
-${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- 'sudo mkdir /var/Users'
-
-# Check /Users directory is writeable
+# Verify /Users is writable via transient-ro
 ${SSH} core@api.${SNC_PRODUCT_NAME}.${BASE_DOMAIN} -- 'sudo mkdir /Users/foo && sudo rm -fr /Users/foo'
```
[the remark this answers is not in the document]
praveenkumar: @coderabbitai Without transient-ro active, /Users is not able to symlink to /var/Users so if transient-ro is not active then it will fail.

snc.sh
praveenkumar: In next revert we removed this login logic itself. So each revert happen stage by stage to have right history of commits.

anjannath (approved):

PR #1237, opened 2026-06-04
Add optional cluster-wide proxy support for snc.sh

praveenkumar:
Allow provisioning behind an HTTP/HTTPS proxy via SNC_USE_PROXY, SNC_HTTP_PROXY, and SNC_HTTPS_PROXY. When enabled, patch install-config with proxy URLs and a noProxy list that includes the internal API (api-int.<cluster>.<baseDomain>) plus standard cluster-local entries.

## Summary by CodeRabbit

* **New Features**
  * Cluster-wide proxy configuration for bundle/cluster installation via environment variables (SNC_USE_PROXY, SNC_HTTP_PROXY, SNC_HTTPS_PROXY)
  * Proxying disabled by default; HTTP/HTTPS proxy values are required when proxying is enabled
  * Proxy settings are applied earlier in the installation flow so they take effect during provisioning

* **Documentation**
  * README updated with usage guidance and an example for exporting proxy variables before running installation


snc-library.sh
```diff
@@ -113,6 +113,22 @@ function replace_pull_secret() {
...
+        fi
+        if [[ -z "${SNC_HTTP_PROXY}" || -z "${SNC_HTTPS_PROXY}" ]]; then
+                echo "SNC_USE_PROXY=enabled requires SNC_HTTP_PROXY and SNC_HTTPS_PROXY"
+                exit 1
+        fi
+        local api_int="api-int.${SNC_PRODUCT_NAME}.${BASE_DOMAIN}"
+        local no_proxy="localhost,.cluster.local,.svc,127.0.0.1,${api_int}"
+        ${YQ} eval --inplace ".proxy.httpProxy = \"${SNC_HTTP_PROXY}\"" "${install_config}"
+        ${YQ} eval --inplace ".proxy.httpsProxy = \"${SNC_HTTPS_PROXY}\"" "${install_config}"
+        ${YQ} eval --inplace ".proxy.noProxy = \"${no_proxy}\"" "${install_config}"
```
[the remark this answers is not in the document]
praveenkumar: This is used for internal build and proxy doesn't content the credentials

snc-library.sh
```diff
@@ -113,6 +113,22 @@ function replace_pull_secret() {
         set -x
 }
 
+function configure_install_config_proxy() {
+        local install_config=$1
+        if [[ ${SNC_USE_PROXY} != "enabled" ]]; then
+                return 0
+        fi
```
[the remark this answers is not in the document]
praveenkumar: in the doc it is mentioned to use only enabled or disabled (which is by default)

createdisk.sh
```diff
@@ -49,6 +49,18 @@ VM_IP=$(sudo virsh domifaddr ${VM_NAME} | tail -2 | head -1 | awk '{print $4}' |
 
 wait_for_ssh ${VM_NAME} ${VM_IP}
 
+# Configure proxy if enabled
+if [ "${SNC_USE_PROXY}" == "enabled" ]; then
+    ${SSH} core@${VM_IP} 'sudo bash -x -s' <<EOF
+     tee /etc/environment <<TEE
+http_proxy=${SNC_HTTP_PROXY}
```
cfergeau: Is there actual http traffic during openshift installation? or everything goes over https?

createdisk.sh
```diff
@@ -49,6 +49,18 @@ VM_IP=$(sudo virsh domifaddr ${VM_NAME} | tail -2 | head -1 | awk '{print $4}' |
 
 wait_for_ssh ${VM_NAME} ${VM_IP}
 
+# Configure proxy if enabled
+if [ "${SNC_USE_PROXY}" == "enabled" ]; then
```
cfergeau: I’d mirror the behaviour of `http_proxy`/`https_proxy` in curl/… If `SNC_HTTP_PROXY` or `SNC_HTTPS_PROXY` is set, then this means we want to use a proxy. No need for a separate `SNC_USE_PROXY`.

createdisk.sh
praveenkumar: @cfergeau not sure, but as per docs of openshift it is mention to provide both.

createdisk.sh
praveenkumar: @cfergeau updated

createdisk.sh
praveenkumar: @cfergeau updated

PR #1240, opened 2026-06-08
Add fedora base repo to get qemu-user-static package

praveenkumar:
By default qemu-user-static package is part of base repo and doesn't get include to update repo until it have and update. For F44 it is observed that there is no update to this package so install fails since only update repo is checked. This PR make sure both base and update repo is enabled before downloading that package.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Fedora repository configuration for aarch64 disk creation.


praveenkumar:
@coderabbitai review

cfergeau (approved):
/lgtm
/approve

PR #1246, opened 2026-06-15
revert disk size to 31G and add workaround for download pods disk usuage.

praveenkumar:


snc.sh
```diff
@@ -272,6 +277,11 @@ retry ${OC} patch image.config.openshift.io cluster -p '{"spec": {"additionalTru
 # Remove the machine config for chronyd to make it active again
 retry ${OC} delete mc chronyd-mask
 
+# Remove the downloads pods because of https://redhat.atlassian.net/browse/OCPBUGS-88450
+# as workaround
+retry ${OC} delete pod -n openshift-console -l app=console,component=downloads
+retry ${OC} wait --for=condition=Ready pod -n openshift-console -l app=console,component=downloads --timeout=60s
```
cfergeau: Why do we do it twice? After the VM is installed/running with `crc`, will these pods be recreated and fill again the disk?

praveenkumar: here we are doing twice because it is observed that as part of MCO change when node reboot it sometime again have those download pod restarted hence we are doing it twice just to make sure it  doesn't fill the disk. 

> After the VM is installed/running with crc, will these pods be recreated 

Bundles are created after removing every pod/container so with crc those container come up cleanly without filling the disk space until user manually restart that download pod.

cfergeau: > until user manually restart that download pod.

but these download pods are not auto-created/auto-started, they would have to be explicitly started by the user? If that’s correct, then ack from me.

